Issue #19, opened 2018-06-11
Base UI components: TextView


sleepylee:
## Issue type ✅
- [x] Feature request

## Insights 🤔⁉️
Adding TextView with:
- [x] Proper styling, style inheritance
- [x] **Font weigth** support
- [x] **Custom Font** support
- [x] **Clickable** support (with ripple effect, borderless and border)
- [x] Update README so people know how to use it.

## Environment, Step to reproduce 💻
- API 19+

## Expected result:
A few text styles that are appplicable when initializing new project quickly:

<img src="https://user-images.githubusercontent.com/13435717/41225120-4beb5ba0-6d98-11e8-9629-8db28e7f7c68.png" width=200 /> 
(photo is for demonstration purpose only)

PR #29, opened 2018-06-18
[#19] Adding Custom TextView styles

sleepylee:
## What happened 🤔
Adding custom font TextView support. Including a few styles besides of the basic default TextView style:
- Bold 
- Clickable

## Insight 👀
- There are other ways that we can achieve the same thing, but these are the ways we usually use:
   - Adding default style to AppTheme at `<item name="android:textViewStyle">style_here</item>`
   - Using them by their explicit style from the UI component: `style="@style/Our_Style"`

- **TO USE THIS:**
   - First: adding your font files to the `/app/src/main/res/font` dir then adjust the clarification in font-family to point to the newly added files.
   - Adjust `Widget.Template.TextView` for your usage and start using them on the TextView you want to apply the style on.

## PoW (a.k.a Proof Of Work)💪

- Default                    |                                  Clickable                        |                            Bold


sleepylee:
Fix #19

PR #41, opened 2020-03-27
[#41] Update jacoco report

Wadeewee:
## What happened 🤔
The report has shown only one module when we open index.html in browser from both modules.
This is a step for checking.
- run ./gradlew jacocoTestReport testStagingDebugUnitTest
- open index.html in browser from the data module, You will see the report only one module that is "com.nimble3.data..."
- open index.html in browser from app module, The report has shown the same with data module that is "com.nimble3.data..." but it should be show "com.nimble3.ui..." or both modules.
 
Screen Shot 2563-03-27 at 10 13 36
Screen Shot 2563-03-27 at 10 13 51
Screen Shot 2563-03-27 at 10 14 10

## Insight 👀
Update the directory file in jacoco.gragle script.
We need to define a directory file for both app and data modules.

## PoW (a.k.a Proof Of Work)💪
Run script -> ./gradlew jacocoTestReport

Screen Shot 2563-03-30 at 10 09 34
Screen Shot 2563-03-30 at 10 20 39
Screen Shot 2563-03-30 at 10 20 56

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2,11 +2,16 @@
 
 buildscript {
     ext {
+        build_gradle_version = '3.6.1'
+
         android_version_code = 1
         android_version_name = "0.1.0"
         android_min_sdk_version = 19
         android_target_sdk_version = 28
         android_compile_sdk_version = 28
+
+        jacoco_core_version = "0.8.5"
+        jacoco_android_plugin_version = '0.1.4'
 
         // Dependencies (Alphabet sorted)
         android_legacy_support_version = '1.0.0'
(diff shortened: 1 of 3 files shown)


jean-t86:
When I run the ./gradlew jacocoTestReport testStagingDebugUnitTest command, I get a different report result than you:

image

I can see com.bumptech.glide in the coverage. This is not ideal because the coverage numbers will skew the real results.

I also see the DI classes. Could those be excluded? 🤔

Wadeewee:
> @Wadeewee When I run the ./gradlew jacocoTestReport testStagingDebugUnitTest command, I get a different report result than you:
> 
> image
> 
> I can see com.bumptech.glide in the coverage. This is not ideal because the coverage numbers will skew the real results.
> 
> I also see the DI classes. Could those be excluded? 🤔

@jean-t86  I got the report like this when I run ./gradlew jacocoTestReport testStagingDebugUnitTest
Screen Shot 2563-03-30 at 09 57 02 I didn't get  com.bumptech.glide in the coverage.
Can you try to run ./gradlew clean and then ./gradlew jacocoTestReport testStagingDebugUnitTest again.

PR #42, opened 2020-04-01
Revert "[#41] Update jacoco report"

jean-t86:
Reverts nimblehq/android-templates#41

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2,16 +2,11 @@
 
 buildscript {
     ext {
-        build_gradle_version = '3.6.1'
-
         android_version_code = 1
         android_version_name = "0.1.0"
         android_min_sdk_version = 19
         android_target_sdk_version = 28
         android_compile_sdk_version = 28
-
-        jacoco_core_version = "0.8.5"
-        jacoco_android_plugin_version = '0.1.4'
 
         // Dependencies (Alphabet sorted)
         android_legacy_support_version = '1.0.0'
@@ -27,7 +22,7 @@ buildscript {
         glide_version               = '4.8.0'
         gson_version                = '2.4'
 
-        kotlin_version              = '1.3.61'
+        kotlin_version              = '1.3.11'
 
         moshi_version               = '1.6.0'
         multidex_version            = '2.0.1'
@@ -54,9 +49,9 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath "com.android.tools.build:gradle:$build_gradle_version"
+        classpath 'com.android.tools.build:gradle:3.2.1'
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
-        classpath "com.dicedmelon.gradle:jacoco-android:$jacoco_android_plugin_version"
+        classpath 'com.dicedmelon.gradle:jacoco-android:0.1.2'
         // NOTE: Do not place your application dependencies here; they belong
         // in the individual module build.gradle files
     }
--- a/config/jacoco.gradle
+++ b/config/jacoco.gradle
@@ -1,20 +1,13 @@
 apply plugin: 'jacoco'
 
 jacoco {
-    toolVersion = "$jacoco_core_version"
+    toolVersion = "0.8.3"
     reportsDir = file("$buildDir/reports/jacoco")
 }
 
-tasks.withType(Test) {
-    jacoco.includeNoLocationClasses = true
-}
-
-task jacocoTestReport(type: JacocoReport) {
+task jacocoTestReport(type: JacocoReport, dependsOn: "testStagingDebugUnitTest") {
     group = "Reporting"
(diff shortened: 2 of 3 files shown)


build.gradle
```diff
@@ -27,7 +22,7 @@ buildscript {
         glide_version               = '4.8.0'
         gson_version                = '2.4'
 
-        kotlin_version              = '1.3.61'
+        kotlin_version              = '1.3.11'
```
luongvo, line 25: why do we need to downgrade kotlin version?

build.gradle
```diff
@@ -54,9 +49,9 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath "com.android.tools.build:gradle:$build_gradle_version"
+        classpath 'com.android.tools.build:gradle:3.2.1'
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
-        classpath "com.dicedmelon.gradle:jacoco-android:$jacoco_android_plugin_version"
+        classpath 'com.dicedmelon.gradle:jacoco-android:0.1.2'
```
luongvo, line 54: why do we remove version definition for those? ☝️

config/jacoco.gradle
```diff
@@ -1,21 +1,14 @@
 apply plugin: 'jacoco'
 
 jacoco {
-    toolVersion = "$jacoco_core_version"
+    toolVersion = "0.8.3"
```
luongvo, line 4: please explain the reason why we have to downgrade this jacoco ver too 🙏 I can't get it from PR description https://github.com/nimblehq/android-templates/pull/41
Also, why do we need to remove ver definition from build.gradle?

config/jacoco.gradle
```diff
@@ -29,39 +22,27 @@ task jacocoTestReport(type: JacocoReport) {
                          'android/**/*.*']
 
     def packagesExcluded = ['com/nimbl3/di/**',
-                            'com/nimbl3/ui/**/di/**',
                             'com/nimbl3/extension/**',
                             'com/nimbl3/**/extension/**']
     // def filesExcluded = [ ]
     // def fileFilter = fileGenerated + packagesExcluded + filesExcluded
```
luongvo, line 28: redundant comments, or we could enable them back to keep a common config
```
def filesExcluded = [ ]
def fileFilter = fileGenerated + packagesExcluded + filesExcluded
```

config/jacoco.gradle
```diff
@@ -29,39 +22,27 @@ task jacocoTestReport(type: JacocoReport) {
...
     def fileFilter = fileGenerated + packagesExcluded
 
-    classDirectories.from = fileTree(
-        dir: "$project.rootDir/app/build/intermediates/javac/stagingDebug/compileStagingDebugJavaWithJavac/classes",
-        excludes: fileFilter
-    ) + fileTree(
-        dir: "$project.rootDir/data/build/intermediates/javac/stagingDebug/compileStagingDebugJavaWithJavac/classes",
+    classDirectories = fileTree(
+        dir: "$buildDir/intermediates/javac/stagingDebug/compileStagingDebugJavaWithJavac/classes",
```
luongvo, line 32: FYI this part will be changed to `$buildDir/intermediates/javac/stagingDebug/classes` after upgraded to 0.8.5 https://github.com/redplanethotels/rphl-android-app/pull/2109, I guess @jean-t86 and @Tuubz  know this.

Wadeewee:
@jean-t86 Is this old version before you merge right?

Tuubz:
@luongvo 

This PR > https://github.com/nimblehq/android-templates/pull/41 was merged too early as it didn't receive enough approvals.

So what we're trying to do here is to revert the changes first, then once this has been approved and merged, we'll bring back this https://github.com/nimblehq/android-templates/pull/41

Is there a better approach for this?

config/jacoco.gradle
Tuubz: Let's not forget to check this when bringing back #41 @Wadeewee 🙏

config/jacoco.gradle
Tuubz: I have to agree, it doesn't feel like this comment has any value...

What do you mean with `could enable them back to keep a common config` ?

build.gradle
Tuubz: Just reverting the changes from #41

build.gradle
Tuubz: Just reverting the changes from #41

config/jacoco.gradle
Tuubz: Just reverting the changes from #41

config/jacoco.gradle
Tuubz: Anyhow, we can keep these changes when bringing back #41 

Please keep this in mind @Wadeewee 🙏

luongvo:
A test comment

PR #43, opened 2020-04-02
Update jacoco report

Wadeewee:
## What happened 🤔
The report has shown only one module when we open index.html in browser from both modules.
This is a step for checking.
- run ./gradlew jacocoTestReport testStagingDebugUnitTest
- open index.html in browser from the data module, You will see the report only one module that is "com.nimble3.data..."
- open index.html in browser from app module, The report has shown the same with data module that is "com.nimble3.data..." but it should be show "com.nimble3.ui..." or both modules.
 
Screen Shot 2563-03-27 at 10 13 36
Screen Shot 2563-03-27 at 10 13 51
Screen Shot 2563-03-27 at 10 14 10

## Insight 👀
Update the directory file task jacoco in jacoco.gragle script.
We need to define a directory file for both app and data modules.

## PoW (a.k.a Proof Of Work)💪
Run script -> ./gradlew jacocoTestReport, ./gradlew jacocoTestReport testStagingDebugUnitTest

Screen Shot 2563-04-08 at 13 01 54
Screen Shot 2563-04-08 at 14 24 18
Screen Shot 2563-04-08 at 14 24 27

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -10,9 +10,6 @@ buildscript {
         android_target_sdk_version = 28
         android_compile_sdk_version = 28
 
-        jacoco_core_version = "0.8.5"
-        jacoco_android_plugin_version = '0.1.4'
-
         // Dependencies (Alphabet sorted)
         android_legacy_support_version = '1.0.0'
         android_support_version     = '1.1.0-alpha01'
@@ -56,7 +53,6 @@ buildscript {
     dependencies {
         classpath "com.android.tools.build:gradle:$build_gradle_version"
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
-        classpath "com.dicedmelon.gradle:jacoco-android:$jacoco_android_plugin_version"
         // NOTE: Do not place your application dependencies here; they belong
         // in the individual module build.gradle files
     }
--- a/build.gradle
+++ b/build.gradle
@@ -2,11 +2,16 @@
 
 buildscript {
     ext {
+        build_gradle_version = '3.6.1'
+
         android_version_code = 1
         android_version_name = "0.1.0"
         android_min_sdk_version = 19
         android_target_sdk_version = 28
         android_compile_sdk_version = 28
+
+        jacoco_core_version = "0.8.5"
+        jacoco_android_plugin_version = '0.1.4'
 
         // Dependencies (Alphabet sorted)
         android_legacy_support_version = '1.0.0'
@@ -22,7 +27,7 @@ buildscript {
         glide_version               = '4.8.0'
         gson_version                = '2.4'
 
-        kotlin_version              = '1.3.11'
+        kotlin_version              = '1.3.61'
 
         moshi_version               = '1.6.0'
         multidex_version            = '2.0.1'
@@ -49,9 +54,9 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.2.1'
+        classpath "com.android.tools.build:gradle:$build_gradle_version"
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
-        classpath 'com.dicedmelon.gradle:jacoco-android:0.1.2'
+        classpath "com.dicedmelon.gradle:jacoco-android:$jacoco_android_plugin_version"
         // NOTE: Do not place your application dependencies here; they belong
         // in the individual module build.gradle files
     }
--- a/config/jacoco.gradle
+++ b/config/jacoco.gradle
@@ -1,13 +1,22 @@
 apply plugin: 'jacoco'
 
-jacoco {
-    toolVersion = jacoco_core_version
-    reportsDir = file("$buildDir/reports/jacoco")
-}
+def fileGenerated = ['**/R.class',
+                     '**/R$*.class',
+                     '**/*$ViewBinder*.*',
+                     '**/*$InjectAdapter*.*',
+                     '**/*Injector*.*',
+                     '**/BuildConfig.*',
+                     '**/Manifest*.*',
+                     '**/*_ViewBinding*.*',
+                     '**/*Adapter*.*',
+                     '**/*Test*.*',
+                     'android/**/*.*']
 
-tasks.withType(Test) {
-    jacoco.includeNoLocationClasses = true
-}
+def packagesExcluded = ['com/nimbl3/di/**',
+                        'com/nimbl3/ui/**/di/**',
+                        'com/bumptech/glide']
+
+def fileFilter = fileGenerated + packagesExcluded
 
 task jacocoTestReport(type: JacocoReport) {
     group = "Reporting"
@@ -16,30 +25,11 @@ task jacocoTestReport(type: JacocoReport) {
     dependsOn ":app:testStagingDebugUnitTest"
     dependsOn ":data:testStagingDebugUnitTest"
 
-    def fileGenerated = ['**/R.class',
-                         '**/R$*.class',
-                         '**/*$ViewBinder*.*',
-                         '**/*$InjectAdapter*.*',
-                         '**/*Injector*.*',
-                         '**/BuildConfig.*',
-                         '**/Manifest*.*',
-                         '**/*_ViewBinding*.*',
-                         '**/*Adapter*.*',
-                         '**/*Test*.*',
-                         'android/**/*.*']
-
-    def packagesExcluded = ['com/nimbl3/di/**',
-                            'com/nimbl3/ui/**/di/**',
-                            'com/nimbl3/extension/**',
-                            'com/nimbl3/**/extension/**']
-
-    def fileFilter = fileGenerated + packagesExcluded
-
     classDirectories.from = fileTree(
-        dir: "$buildDir/intermediates/javac/stagingDebug/classes",
+        dir: "$project.rootDir/app/build/intermediates/javac/stagingDebug/classes",
         excludes: fileFilter
     ) + fileTree(
-        dir: "$buildDir/intermediates/javac/stagingDebug/classes",
+        dir: "$project.rootDir/data/build/intermediates/javac/stagingDebug/classes",
         excludes: fileFilter
     ) + fileTree(
         dir: "$project.rootDir/app/build/tmp/kotlin-classes/stagingDebug",
@@ -51,16 +41,14 @@ task jacocoTestReport(type: JacocoReport) {
 
     sourceDirectories.from = files([
         "$project.rootDir/app/src/main/java",
-        "$project.rootDir/data/src/main/java"])
+        "$project.rootDir/data/src/main/java"
+    ])
 
     executionData.from = fileTree(dir: project.rootDir, includes: [
-        'app/build/jacoco/testStagingDebugUnitTest.exec',
-        'data/build/jacoco/testStagingDebugUnitTest.exec'])
+        "app/build/jacoco/testStagingDebugUnitTest.exec",
+        "data/build/jacoco/testStagingDebugUnitTest.exec"
+    ])
 
-    reports {
-        xml.enabled = true
-        html.enabled = true
-    }
 }
 
 tasks.withType(Test) {
--- a/config/jacoco.gradle
+++ b/config/jacoco.gradle
@@ -1,13 +1,20 @@
 apply plugin: 'jacoco'
 
 jacoco {
-    toolVersion = "0.8.3"
+    toolVersion = "$jacoco_core_version"
     reportsDir = file("$buildDir/reports/jacoco")
 }
 
-task jacocoTestReport(type: JacocoReport, dependsOn: "testStagingDebugUnitTest") {
+tasks.withType(Test) {
+    jacoco.includeNoLocationClasses = true
+}
+
+task jacocoTestReport(type: JacocoReport) {
     group = "Reporting"
     description = "Generate Jacoco coverage reports for Debug build"
+
+    dependsOn ":app:testStagingDebugUnitTest"
+    dependsOn ":data:testStagingDebugUnitTest"
 
     def fileGenerated = ['**/R.class',
                          '**/R$*.class',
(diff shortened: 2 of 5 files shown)


Tuubz:
Any reason we need `[#43]` in PR title?

cc @sleepylee

sleepylee:
@Tuubz it's the Issues number for me when I created the other PRs before, this one doesn't seem to have it. If we have a board with ticket number then let's attach it, if not I think we can just ditch it
image

Wadeewee:
@Tuubz @sleepylee I will remove it.

Wadeewee:
@luongvo Could you please review this PR again?

config/jacoco.gradle
```diff
@@ -1,14 +1,21 @@
 apply plugin: 'jacoco'
 
 jacoco {
-    toolVersion = "0.8.3"
+    toolVersion = "$jacoco_core_version"
```
luongvo, line 4: toolVersion = jacoco_core_version

config/jacoco.gradle
```diff
@@ -22,27 +29,38 @@ task jacocoTestReport(type: JacocoReport, dependsOn: "testStagingDebugUnitTest")
...
                             'com/nimbl3/**/extension/**']
-    // def filesExcluded = [ ]
-    // def fileFilter = fileGenerated + packagesExcluded + filesExcluded
+
     def fileFilter = fileGenerated + packagesExcluded
 
-    classDirectories = fileTree(
-        dir: "$buildDir/intermediates/javac/stagingDebug/compileStagingDebugJavaWithJavac/classes",
+    classDirectories.from = fileTree(
+        dir: "$project.rootDir/app/build/intermediates/javac/stagingDebug/compileStagingDebugJavaWithJavac/classes",
```
luongvo, line 39: Could you please check is this a correct path for java files? I think this path will be changed to `$buildDir/intermediates/javac/stagingDebug/classes` after upgraded to 0.8.5 https://github.com/redplanethotels/rphl-android-app/pull/2109

config/jacoco.gradle
```diff
@@ -22,27 +29,38 @@ task jacocoTestReport(type: JacocoReport, dependsOn: "testStagingDebugUnitTest")
...
+        excludes: fileFilter
+    ) + fileTree(
+        dir: "$project.rootDir/data/build/intermediates/javac/stagingDebug/compileStagingDebugJavaWithJavac/classes",
```
luongvo, line 42: same

Wadeewee:
@luongvo PR Updated.

config/jacoco.gradle
```diff
@@ -22,27 +29,38 @@ task jacocoTestReport(type: JacocoReport, dependsOn: "testStagingDebugUnitTest")
...
     def fileFilter = fileGenerated + packagesExcluded
 
-    classDirectories = fileTree(
-        dir: "$buildDir/intermediates/javac/stagingDebug/compileStagingDebugJavaWithJavac/classes",
+    classDirectories.from = fileTree(
+        dir: "$buildDir/intermediates/javac/stagingDebug/classes",
+        excludes: fileFilter
+    ) + fileTree(
+        dir: "$buildDir/intermediates/javac/stagingDebug/classes",
+        excludes: fileFilter
```
Tuubz, line 43: I see a duplication here? 🤔

Wadeewee: Yes @Tuubz we can use one, if we use $buildDir instead project.rootDir, I will update.

config/jacoco.gradle
```diff
@@ -32,15 +32,14 @@ task jacocoTestReport(type: JacocoReport) {
...
+
     def fileFilter = fileGenerated + packagesExcluded
 
     classDirectories.from = fileTree(
-        dir: "$project.rootDir/app/build/intermediates/javac/stagingDebug/compileStagingDebugJavaWithJavac/classes",
+        dir: "$buildDir/intermediates/javac/stagingDebug/classes",
         excludes: fileFilter
     ) + fileTree(
-        dir: "$project.rootDir/data/build/intermediates/javac/stagingDebug/compileStagingDebugJavaWithJavac/classes",
+        dir: "$buildDir/intermediates/javac/stagingDebug/classes",
```
Wadeewee, line 42: @luongvo Sorry I just noticed the report when we use $buildDir/intermediates/javac/stagingDebug/classes we got this package  com.bumptech.glide also like this
Screen Shot 2563-04-07 at 14 52 58

luongvo: Can we exclude them out?

Wadeewee: Yes, we can, I will update @luongvo

Wadeewee:
@Tuubz @luongvo Updated.

jean-t86:
@nimblehq/android-chapter guys we need two more approvals on this please 🙏

config/jacoco.gradle
```diff
@@ -49,13 +47,9 @@ task jacocoTestReport(type: JacocoReport) {
...
     )
 
-    sourceDirectories.from = files([
-        "$project.rootDir/app/src/main/java",
-        "$project.rootDir/data/src/main/java"])
+    sourceDirectories.from = files([android.sourceSets.main.java.srcDirs])
 
-    executionData.from = fileTree(dir: project.rootDir, includes: [
-        'app/build/jacoco/testStagingDebugUnitTest.exec',
-        'data/build/jacoco/testStagingDebugUnitTest.exec'])
```
luongvo: is this correct to remove `data` out of executionData?

Wadeewee: I'm not sure @luongvo but I think I will bring it back like above to make sure the file is exact.

Tuubz:
When running the gradle-script, I'm getting a small warning from jacoco:

`Kotlin plugin should be enabled before 'kotlin-android-extensions'`

Would you mind applying the change in `build.gradle` please 🙏

config/jacoco.gradle
```diff
@@ -22,27 +29,36 @@ task jacocoTestReport(type: JacocoReport, dependsOn: "testStagingDebugUnitTest")
...
+        dir: "$project.rootDir/app/build/tmp/kotlin-classes/stagingDebug",
+        excludes: fileFilter
+    ) + fileTree(
+        dir: "$project.rootDir/data/build/tmp/kotlin-classes/stagingDebug",
         excludes: fileFilter
     )
+
+    sourceDirectories.from = files([
+        "$project.rootDir/app/src/main/java",
+        "$project.rootDir/data/src/main/java"])
```
Tuubz, line 52: Is this considered to be the same?

```
    sourceDirectories.from = fileTree(dir: project.rootDir, includes: [
        "app/src/main/java",
        "data/src/main/java"
    ])
```

To keep the consistency with

```
  executionData.from = fileTree(dir: project.rootDir, includes: [
        'app/build/jacoco/testStagingDebugUnitTest.exec',
        'data/build/jacoco/testStagingDebugUnitTest.exec'
])
```

config/jacoco.gradle
```diff
@@ -22,27 +29,36 @@ task jacocoTestReport(type: JacocoReport, dependsOn: "testStagingDebugUnitTest")
...
+
+    executionData.from = fileTree(dir: project.rootDir, includes: [
```
Tuubz, line 54: This used to be `buildDir` instead of `rootDir`, just curious if this change made any difference? 🤔

config/jacoco.gradle
```diff
@@ -22,27 +29,36 @@ task jacocoTestReport(type: JacocoReport, dependsOn: "testStagingDebugUnitTest")
...
+        dir: "$buildDir/intermediates/javac/stagingDebug/classes",
         excludes: fileFilter
     ) + fileTree(
-        dir: "$buildDir/tmp/kotlin-classes/stagingDebug",
+        dir: "$project.rootDir/app/build/tmp/kotlin-classes/stagingDebug",
+        excludes: fileFilter
+    ) + fileTree(
+        dir: "$project.rootDir/data/build/tmp/kotlin-classes/stagingDebug",
         excludes: fileFilter
     )
```
Tuubz, line 48: `project.rootDir` can be replaced with `buildDir`

Then we can shorten `"$project.rootDir/app/build/tmp/kotlin-classes/stagingDebug"` to `"$buildDir/tmp/kotlin-classes/stagingDebug"`

What do you think?

build.gradle
```diff
@@ -49,9 +54,9 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.2.1'
+        classpath "com.android.tools.build:gradle:$build_gradle_version"
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
-        classpath 'com.dicedmelon.gradle:jacoco-android:0.1.2'
+        classpath "com.dicedmelon.gradle:jacoco-android:$jacoco_android_plugin_version"
```
Tuubz, line 59: Why do we need this? It seems like it's still working without 😊 

Any explanation, please?

cc @sleepylee

Wadeewee: Yes, we no need to use this plugin for this case so should we remove it @Tuubz @sleepylee

Tuubz: Still curious tho, perhaps I missed something? 🤔 

https://github.com/arturdm/jacoco-android-gradle-plugin

cc @sleepylee

sleepylee: it has this plugin init: https://github.com/arturdm/jacoco-android-gradle-plugin/blob/master/src/main/resources/META-INF/gradle-plugins/jacoco-android.properties that points to the gradle Task that help generating the Report. You can just add this 1 line to our jacoco properties file and that should be equivalent to this dependency I think 😄

config/jacoco.gradle
```diff
@@ -1,14 +1,21 @@
...
+    toolVersion = jacoco_core_version
     reportsDir = file("$buildDir/reports/jacoco")
 }
 
-task jacocoTestReport(type: JacocoReport, dependsOn: "testStagingDebugUnitTest") {
+tasks.withType(Test) {
+    jacoco.includeNoLocationClasses = true
+}
+
+task jacocoTestReport(type: JacocoReport) {
```
sleepylee, line 12: we already had this task, so I think that's why you won't need it, or, check again if u can use it and remove this task entirely

build.gradle
Tuubz: Aah, I see. So what this plugins does, is enabling to create reports for multiple variants/flavors automatically? Meaning without we would need to specify it manually?

sleepylee: @Tuubz I believe so. This is from the repository README

image

Also, since it's a plugin, it requires 
`apply plugin: 'jacoco-android'`
https://github.com/arturdm/jacoco-android-gradle-plugin

For us, we are creating our own task and only apply the origin jacoco plugin: https://github.com/nimblehq/android-templates/blob/kotlin/config/jacoco.gradle
I'm not sure where did @Wadeewee found this piece of code, but I think not many people used it because of different Kotlin config ever since

sleepylee: or maybe they have fixed it already. Then you will just need to use the plugin to reduce the amount of customized Task we are creating here. Our job is to try them out, test if they work, and propose a better solution if it is :)

sleepylee: https://github.com/nimblehq/android-templates/blob/82ff99dc3603aae145f3614785507f5248855ac8/app/build.gradle#L3

Tuubz: From my understanding is that by specifying the custom tasks ourselves it allows us to have more control over it (Ex: Choosing what files to exclude)

So I don't see the point of specifying both the tasks ourselves + using the plug-in at the same time :/

Tuubz: I understand what it does, but rather asking why we chose to integrate this plugin from the start? Even though we had customized tasks specified already. Just trying to get a better understanding on this 😇 🙏

config/jacoco.gradle
Tuubz: The reason why you're suggesting to remove this, is because it's supposed to be integrated in that plugin already, right?

build.gradle
sleepylee: no we didn't have this plugin before, @Tuubz 🤔 isn't that your question why @Wadeewee brought this dependency in without using yeah?

config/jacoco.gradle
sleepylee: yes if the plugin work per our expectation then we can make the template leaner by using just the plugin config instead of writing gradle Task ourselves

build.gradle
Tuubz: The plugin was already there before she started on it 🤔

sleepylee: ohhh yeah, sorry my bad reading, she only extracts the version out

sleepylee: https://github.com/nimblehq/android-templates/pull/16/files#diff-c197962302397baf3a4cc36463dce5eaR52 @Tuubz it was indeed added by me, so I think it was my bad at that time I didn't check it properly after following a few instructions on jacoco setup 🙏 we could potentially just ditch it.

Tuubz: Thank you @sleepylee 

@Wadeewee 

Please double check if everything is working as expected without the plugin, if so please remove it 😊

Wadeewee: I'm sorry @Tuubz @sleepylee I just understand and try that plugin work. it seems like it cannot support multi-module I think, see the log
Screen Shot 2563-04-08 at 11 57 54

Wadeewee: @jean-t86 @luongvo @Tuubz @sleepylee PR Update.Please review again :)

Wadeewee: I'm going update POW.

config/jacoco.gradle
Tuubz: 👋 Any reply to this?

config/jacoco.gradle
```diff
@@ -54,17 +43,7 @@ task jacocoTestReport(type: JacocoReport) {
...
-        'app/build/jacoco/testStagingDebugUnitTest.exec',
-        'data/build/jacoco/testStagingDebugUnitTest.exec'])
-
-    reports {
-        xml.enabled = true
-        html.enabled = true
-    }
-}
+        "app/build/jacoco/testStagingDebugUnitTest.exec",
+        "data/build/jacoco/testStagingDebugUnitTest.exec"])
```
Wadeewee, line 47: I updated it.@Tuubz

Tuubz:
Looking good! 😁 

I got one final question: Currently we only make reports for `stagingDebug`, would it be of any good use to have it for other variants as well? 

cc @jean-t86 @luongvo @sleepylee

jean-t86:
> Looking good! 😁
> 
> I got one final question: Currently we only make reports for `stagingDebug`, would it be of any good use to have it for other variants as well?
> 
> cc @jean-t86 @luongvo @sleepylee

Traditionally, unit tests run in a sandbox completely cut off from any sort of integration. Where needed, those integrations are either mocked or faked. Therefore, it does not really matter against which build variant they run.

Having said that, for industry standards purposes, I would suggest having a `develop` environment and have the unit tests run against the `developDebug` build variant.

Having unit tests run against `stagingDebug` makes little sense in terms of proper nomenclature.

To understand why we need to look into the definition of [staging](https://searchsoftwarequality.techtarget.com/definition/staging-environment).

In light of this, running unit tests which mock and fake their way into a sandbox make little sense to run against an environment named `staging`.

Amongst the risks is for developers to draw conclusions on the environment based on its naming, which would simply not stand true.

More on this [here](https://github.com/nimblehq/compass/issues/291).

Tuubz:
Alright, thanks for the clarification 🙏 

> Traditionally, unit tests run in a sandbox completely cut off from any sort of integration. Where needed, those integrations are either mocked or faked. Therefore, it does not really matter against which build variant they run.

I agree it would not make sense to have more than one report, as the data is mocked anyhow for all variants.

> Having said that, for industry standards purposes, I would suggest having a develop environment and have the unit tests run against the developDebug build variant.

As we currently don't have a develop variant, and still discussing about it in https://github.com/nimblehq/compass/issues/291, I think we keep it with `staging` for now and change later on then.

cc @Wadeewee

PR #44, opened 2020-04-10
Setup basic fast-lane components with Firebase app distribution

jean-t86:
## What happened 🤔

A basic setup of `fastlane` along with the documentation required to configure and set it up when initializing a new Android project.

This code was lifted straight from The 1 project.

## Insight 👀

Of interest is the fastlane folder, and in particular the following files:
- Appfile - the configuration global to the app, in particular, the package name.
- script/config.rb - contains the configuration that needs to be updated for fastlane to work.

## PoW (a.k.a Proof Of Work)💪

Documentation:
[Distribute Android apps using Fastlane](https://www.notion.so/nimblehq/Distribute-Android-apps-using-Fastlane-27787c03867948bd92ef974669d615b0)


fastlane/Fastfile
```diff
@@ -0,0 +1,38 @@
...
+
+default_platform(:android)
+
+platform :android do
+  before_all do
+    ENV['SLACK_URL'] = Config::SLACK_URL
+    Config.sanity_check
+  end
+
+  desc 'Execute staging deployment'
```
sleepylee, line 29: and Production deployment? 🤔

fastlane/Appfile
```diff
@@ -0,0 +1,4 @@
+# frozen_string_literal: true
+
+json_key_file('') # Path to the json secret file - Follow https://docs.fastlane.tools/actions/supply/#setup to get one
+package_name('com.nimbl3') # e.g. com.krausefx.app
```
sleepylee, line 4: I can't recall, but yeah just make sure this package name is mentioned to the README (HOWTO) as well 👍

fastlane/Fastfile
jean-t86: Supply will be done in a different PR.

Do we need to upload production to Firebase?

fastlane/Appfile
jean-t86: Should I maybe leave this blank instead?

I have blanked SLACK_URL and app_id already.

fastlane/Fastfile
sleepylee: yes we need it when releasing so the client can try it out too, currently, we are sending APK, which is equivalent but less automation 🤷‍♂️

fastlane/Appfile
sleepylee: you can use a cryptic name for it perhaps? e.g, `'com.your.package.name'`

fastlane/Fastfile
jean-t86: Should it be Production debug?

sleepylee: yes for this purpose 🤔 actually we should think of it in a more versatile way I think...because later you can add another lane that uploads the build to Play Store (productionRelease) right?

jean-t86: Yes, there will be a lane for production release that goes straight into Play Store with the help of the `supply` plugin. 

I feel that production builds should use the testers features exposed via the Play Store though. Alpha and Beta channel.

We will need an initiatives about the documentation for setting up a project in Play Store too.

sleepylee:
please update the PR Title

.gitignore
```diff
@@ -1,8 +1,65 @@
...
+captures/
+
+# Intellij
+*.iml
+.idea
+.idea/workspace.xml
+.idea/tasks.xml
+.idea/gradle.xml
+.idea/dictionaries
+.idea/libraries
```
Tuubz, line 45: Can we minimize it to `/.idea/*` as well? Or any reason why it's defined like this? 🙏

.gitignore
```diff
@@ -1,8 +1,65 @@
...
+# Freeline
+freeline.py
+freeline/
+freeline_project_description.json
+
+# Fastlane
+fastlane/report.xml
+fastlane/Preview.html
+fastlane/screenshots
+fastlane/test_output
```
Tuubz, line 65: I haven't run the script myself yet, but could it be that the last file is not relevant anymore, a.k.a can be removed? 😁

Gemfile
```diff
@@ -0,0 +1,7 @@
+source "https://rubygems.org"
+
+gem 'rubocop', require: false
```
Tuubz, line 3: Shall we go for double quotes here? 🙈

app/build.gradle
```diff
@@ -90,20 +90,12 @@ android {
         staging {
             applicationId "com.nimbl3.staging"
         }
-
-        staging_internal {
-            applicationId "com.nimbl3.staging"
-            versionNameSuffix ".${project.versionNameSuffix}"
```
Tuubz, line 96: Does this mean we can remove the following defined in the app gradle file? 

`project.ext.set("versionNameSuffix", "$gitCommitHash${ticketValue}")`

app/build.gradle
```diff
@@ -90,20 +90,12 @@ android {
...
-        }
     }
 
     sourceSets {
         staging {
             res.srcDirs = ['src/staging/res']
         }
-        staging_internal {
-            res.srcDirs = ['src/staging/res']
-        }
```
Tuubz, line 106: Should we do something about `staging_internal` inside the `data` submodule as well?

Tuubz:
I'm done here 😁🚀

.gitignore
jean-t86: Actually, I think the only relevant one still is `fastlane/report.xml.` 

I will remove the others.

Gemfile
jean-t86: Yes. Also, we use Rubocop for linting checks and using single quotes, in that case, is the recommended way.

app/build.gradle
```diff
@@ -23,8 +23,6 @@ android {
         logger.info("This could be a CI build on pre-merged branch so there is no ticket id found!")
     }
 
-    project.ext.set("versionNameSuffix", "$gitCommitHash${ticketValue}")
```
Tuubz, line 26: Oh! One last thing 🙏 

Seems like the part above is not being used anywhere but in this line, I believe we can remove it?

```
 // Our common part of fetching the git hash commit then appending to versionName for internal build.
    def gitCommitHash = "git rev-parse --short HEAD".execute().text.trim()
    def gitCommitMessage = "git show -s --format=%B HEAD".execute().text.trim()
    def ticketValue = ""
    try {
        ticketValue = "." + (gitCommitMessage =~ "\\[(.*)\\]")[0][1]
    } catch (Exception e) {
        logger.info("This could be a CI build on pre-merged branch so there is no ticket id found!")
    }
```

Gemfile
Tuubz: Oh, I was referring to `gem "fastlane"` actually, to have it like this instead: `gem 'fastlane'`

But no biggie, can leave it like it is

data/build.gradle
```diff
@@ -43,9 +39,6 @@ android {
         test {
             java.srcDirs += 'src/test/kotlin'
         }
-        staging_internal {
-            java.srcDirs += 'src/staging'
-        }
```
Tuubz, line 48: I'm actually not very familiar what's going on here, but instead of removing it completely, should it be like this instead?

```
staging {	
    java.srcDirs += 'src/staging'	
}
```

🙏

jean-t86: The Android Framework knows where to look for a default source set. In this case, we did not want to use the default source set location but instead add the staging folder to be shared with the staging_internal flavor.

So to answer your question, no it should not be like this instead.

You can read more about how the framework behaves for [default](https://developer.android.com/studio/build/build-variants#configure-sourcesets) source set folders, and we can discuss further if you feel the need.

Gemfile
jean-t86: Ah I see. I will update it. It doesn't look right at all.

data/build.gradle
Tuubz: Aha, got it! Thanks

Tuubz:
More approvals required here 👋

fastlane/Appfile
```diff
@@ -0,0 +1,4 @@
+# frozen_string_literal: true
+
+json_key_file('') # Path to the json secret file - Follow https://docs.fastlane.tools/actions/supply/#setup to get one
+package_name('com.your.package.name') # e.g. com.krausefx.app
```
luongvo, line 4: I would expect this is our `template app id`, that makes easy for us on finding and replacing later when cloning to the real project? 🤔

fastlane/Appfile
```diff
@@ -0,0 +1,4 @@
+# frozen_string_literal: true
+
+json_key_file('') # Path to the json secret file - Follow https://docs.fastlane.tools/actions/supply/#setup to get one
+package_name('com.nimbl3') # your package name here
```
sleepylee, line 4: @luongvo fixed

manh-t (approved):

luongvo:
@Tuubz 1 more review then we could merge this PR

luongvo:
@manh-t @Tuubz pls consider to review and merge this soon 🙇

lydiasama (approved):

fastlane/Appfile
Tuubz: Should we update `package_name` in the following PR as well ? https://github.com/nimblehq/android-templates/pull/44/files

Tuubz (approved):

luongvo:
@hoangnguyen92dn @Wadeewee needs your reviews at least 🙏

fastlane/Appfile
manh-t: @Tuubz I will merge this PR first, then will rebase the package name to be correct in https://github.com/nimblehq/android-templates/pull/48 . Does it make sense? 🙏

PR #46, opened 2020-04-20
WIP Android Bootstrap (Part 1)

Tuubz:
## What happened 🤔

☝️ 

To ensure all our developers are aligned on the same code styling, our IDEs need to be set accordingly. This will also make the onboarding process more convenient to set-up.

The following configurations will be set up automatically after executing the bootstrap script:

- Java, Kotlin, Groovy & XML code styling 
- Add unambiguous imports on the fly & Optimize imports on the fly (Java)
- Add unambiguous imports on the fly & Optimize imports on the fly (Kotlin)
- Ensure newline at end of file 

Please let me know if you there are any conventions you disagree with, or if there are others conventions we could add as our defaults. We could run a poll to figure things out 🙏 

## Insight 👀

To run the script, please checkout the README.md file

For a more detailed explanation, check https://www.notion.so/Android-Bootstrap-e76f41da49324d60852af01e91a0ea90

## PoW (a.k.a Proof Of Work)💪

You can try to delete your code scheme and uncheck all of the above, execute the script and see the differences

PR #47, opened 2020-04-23
WIP Android Bootstrap (Part 2)

Tuubz:
## What happened 🤔

This is a small extension to #46, which will set-up chosen plugins automatically in our IDE environment:

- https://plugins.jetbrains.com/plugin/7380-adb-idea
- https://plugins.jetbrains.com/plugin/7658-android-drawable-importer

## Insight 👀

External plugins are stored in `"$HOME"/Library/'Application Support'`

Currently the plugin version is fixed, as we have the zip downloaded straight in to our project. As an improvement we could download the plugin dynamically from IntelliJ.

## PoW (a.k.a Proof Of Work)💪

You can confirm this by checking out on the plugins if they have been added/removed


bootstrap/script
```diff
@@ -69,5 +90,9 @@ say "Setting up project configurations... 🚀🔥\n"
 copyCodeStyleConfig
 copyKotlinCodeInsights
 
+say "Setting up plugins... 🚀🔥\n"
+copyAndUnzipPlugin plugin_adb_idea_1.5.4.zip
+copyAndUnzipPlugin plugin_android_drawable_importer_0.6.zip
```
jean-t86, line 95: How often do you think we should update those plugins versions? 

In theory, the IDE would update those for us. However, if too much time goes by, we will start having major updates.

Do you think it could be an issue down the line? 🤔

Tuubz: I don't think there's any need for updating these plugins. Unless it gets to a point where it's unusable with a new version of Android Studio ☠️ 

Will confirm later if it's still working with Android Studio version 4.0.0

PR #48, opened 2020-06-18
[MVVM Structure] Part 2 - Update gradle build config + repackage all modules

luongvo:
https://trello.com/c/AOepchkv

## What happened 🤔
- Update gradle build config
- Update env config, separate them into 3 env: `development`, `staging`, `production`
- Repackage all current modules


manh-t:
Rebased 👨‍🎤  🎸

sleepylee:
can we separate this PR into 4 pieces? 😭  they are:
- Update dependencies lib version
- Add UseCase structure
- Add common-rx module
- Add domain module

manh-t:
@sleepylee we already separate them into commits. Let me separate them into small PRs for tracking easily 🙏

manh-t:
Rebased 🚀

build.gradle
```diff
@@ -4,51 +4,59 @@ buildscript {
...
-        glide_version               = '4.8.0'
-        gson_version                = '2.4'
+        firebase_version = "17.2.2"
+        firebase_perf_version = "19.0.5"
+
+        glide_version = '4.8.0'
+        gson_version = '2.4'
 
         hilt_version                = '2.28-alpha'
         hilt_androidx_version       = '1.0.0-alpha01'
```
manh-t, line 31: lol 🤦

manh-t:
Note to self: Rebase and update package name in fastlane

manh-t:
Rebased and updated Appfile package name 🚀

app/src/main/java/co/nimblehq/ui/main/MainActivity.kt
```diff
@@ -52,11 +52,11 @@ class MainActivity : BaseActivity() {
             .subscribe(this::gotoNextScreen)
             .bindForDisposable()
 
-        RxView.clicks(buttonRefresh)
+        buttonRefresh.clicks()
             .subscribe { viewModel.inputs.refresh() }
             .bindForDisposable()
 
-        RxView.clicks(buttonNext)
+        buttonNext.clicks()
```
lydiasama, line 59: just note: We should rename the view using team convention.

build.gradle
```diff
@@ -4,51 +4,57 @@ buildscript {
...
+        rxkotlin_version = '2.2.0'
+
+        android_test_version = '1.2.0'
+        android_test_junit_version = '1.1.1'
+        test_espresso_version = "3.1.0"
+        test_junit_version = '4.13'
+        test_kluent_version = '1.60'
+        test_mockito_version = '2.1.0'
+        test_robolectric_version = '4.3.1'
+        timber_log_version = '4.7.1'
```
lydiasama, line 57: Do we want to use this style, not align the `=` ?
If we want the old format I think we might need to rearrange it for align `=` in every line.

manh-t: @lydiasama yeah, we should use the old style, I will update it 💪

app/build.gradle
```diff
@@ -76,18 +76,25 @@ android {
...
-            applicationId "com.nimbl3.staging"
+            applicationId "co.nimblehq.staging"
+        }
+
+        development {
+            applicationId "co.nimblehq.dev"
         }
     }
 
     sourceSets {
```
Tuubz, line 91: Should there also be a sourceSet for `production`?

app/build.gradle
```diff
@@ -119,20 +126,19 @@ dependencies {
...
         "androidx.recyclerview:recyclerview:$android_support_version",
-        "com.google.android.material:material:$android_material",
+        "com.google.android.material:material:$android_material_version",
         "androidx.constraintlayout:constraintlayout:$constraint_layout_version",
 
-        "io.reactivex.rxjava2:rxjava:$rxjava_version",
-        "io.reactivex.rxjava2:rxandroid:$rxandroid_version",
-        "com.jakewharton.rxbinding2:rxbinding:$rxbinding_version",
+        "com.jakewharton.rxbinding3:rxbinding:$rxbinding_version",
+        "com.jakewharton.rxbinding3:rxbinding-core:$rxbinding_version",
```
Tuubz, line 133: Thinking if this should be inside `common-rx` module or not

app/src/main/java/co/nimblehq/ui/main/MainActivity.kt
Tuubz: @manh-t What's the reason why we're not using `RxView` anymore? 🙏  Just curious

Tuubz (changes requested):
I tried running the app, but it crashes 😨  

Does it happen to you as well? 🙏

app/src/main/java/co/nimblehq/ui/main/MainActivity.kt
manh-t: @Tuubz we upgrade from `rxbinding2` to `rxbinding3` so the implementation has changed based on the version 😓

manh-t:
@Tuubz I'm able to run app on all 3 variants 😅  Can you please show the error that you are getting?

app/build.gradle
```diff
@@ -76,18 +76,25 @@ android {
...
         }
     }
 
     sourceSets {
         staging {
             res.srcDirs = ['src/staging/res']
         }
+        development {
+            res.srcDirs = ['src/staging/res']
+        }
```
luongvo, line 97: same, source set for custom build flavors are unnecessary

data/build.gradle
```diff
@@ -39,6 +43,9 @@ android {
         test {
             java.srcDirs += 'src/test/kotlin'
         }
+        development {
+            java.srcDirs += 'src/staging'
+        }
```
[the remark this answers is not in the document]
manh-t: This is from old implementation, will remove :pray:

app/build.gradle
manh-t: @Tuubz I think we can keep it in `app` module since this only use in `app` and it is not mandatory 🙏

app/build.gradle
luongvo: @Tuubz nope, we don't need source set for custom build flavor https://github.com/nimblehq/android-templates/pull/48#discussion_r534038256

build.gradle
luongvo: @lydiasama @sleepylee @manh-t @hoangnguyen92dn @Tuubz  I would not agree to this special formatting: `align all dependency versions on the same vertical column on the right`, because:
- We don't touch on this dependency versions config too much, even in case we keep the format without aligning `=` we still able to determine versions easily for upgrading or downgrading... not so hard from my view tbh.
- Applying this special formatting would make hard for us to maintenance, especially for new members to understand what should they do when modifying this file. I know we could create a special style for Groovy file type to keep this format but does that make sense for us to add 1 more step in our process?

I personally only agree on which action helps us to speed up and make easy our development process, not just to keep our code beautiful but brings unnecessary extra steps to achieve. Please bring your ideas up here 🙏

luongvo (commented):
lgtm, approved

luongvo:
@manh-t @Tuubz @hoangnguyen92dn @lydiasama please finalize your review here in your spare time 🍡

hoangnguyen92dn (approved):

manh-t (approved):
On behalf of @luongvo 🚀

build.gradle
```diff
@@ -4,51 +4,59 @@ buildscript {
...
+
+        // testing libraries
+        android_test_version = '1.2.0'
+        android_test_junit_version = '1.1.1'
+        test_espresso_version = "3.1.0"
+        test_junit_version = '4.13'
+        test_kluent_version = '1.60'
+        test_mockito_version = '2.1.0'
+        test_robolectric_version = '4.3.1'
     }
```
lydiasama, line 60: We will use the old format, won't we? cc. @manh-t

lydiasama (commented):
Just a not important thing.

build.gradle
luongvo: @lydiasama I would suggest keeping the current formatting in this PR as below
```
        est_junit_version = '4.13'
        test_kluent_version = '1.60'
        test_mockito_version = '2.1.0'
        test_robolectric_version = '4.3.1'
```

Tuubz (approved):

build.gradle
lydiasama: @luongvo  I'm sorry, I just saw your reply. I agree with this tbh. I used to read it from somewhere about the clean code. 👍🏻

PR #49, opened 2020-06-25
Update mvvm structure with usecase

hoangnguyen92dn:
## What happened 🤔
https://trello.com/c/AOepchkv

## Insight 👀
Update new UseCase structure with Hilt


app/src/main/java/co/nimblehq/ui/base/BaseActivity.kt
```diff
@@ -1,19 +1,51 @@
...
+
+    protected fun Disposable.addToDisposables() = addTo(disposables)
+
+    protected inline infix fun <T> Observable<T>.bindTo(crossinline action: (T) -> Unit) {
+        observeOn(AndroidSchedulers.mainThread())
+            .subscribe { action(it) }
+            .addToDisposables()
+    }
+
+    fun displayError(error: Throwable) {
```
Tuubz, line 47: Should this one be `protected` as well?

app/src/main/java/co/nimblehq/ui/second/SecondActivity.kt
```diff
@@ -61,11 +55,19 @@ class SecondActivity : BaseActivity() {
...
+        requestCode: Int,
+        permissions: Array<out String>,
+        grantResults: IntArray
+    ) {
         super.onRequestPermissionsResult(requestCode, permissions, grantResults)
         onRequestPermissionsResult(requestCode, grantResults)
     }
 
+    private fun bindPersistedData(data: Data) {
+        persistTextView.text = data.content
```
Tuubz, line 68: Should this be `tvPersist` ? Maybe out of scope

app/src/main/java/co/nimblehq/ui/second/SecondViewModel.kt
```diff
@@ -1,37 +1,28 @@
...
-
-        dataFromIntent
-            .subscribe(persistData::onNext)
-            .bindForDisposable()
+    override fun dataFromIntent(data: Data) {
+        persistData.onNext(data)
     }
 
-    override fun setPersistedData() = this.persistData!!
+    override fun setPersistedData() = this.persistData
```
Tuubz, line 20: Hmmm, a bit of confusing name, no? 🙈  We don't `set` here, but we do `get`

Tuubz: Perhaps I'm wrong, I might have been away from Rx too long 😅

domain/src/main/java/co/nimblehq/domain/repository/ApiRepository.kt
```diff
@@ -0,0 +1,24 @@
+package co.nimblehq.domain.repository
+
+import co.nimblehq.data.service.ApiService
+import co.nimblehq.data.service.response.ExampleResponse
+import co.nimblehq.domain.schedulers.BaseSchedulerProvider
+import io.reactivex.Flowable
+import javax.inject.Inject
+
+interface ApiRepository {
+
```
lydiasama, line 10: I think we can remove this redundant line. 🙏🏻

domain/src/main/java/co/nimblehq/domain/data/error/ValidateError.kt
```diff
@@ -0,0 +1,10 @@
+package co.nimblehq.domain.data.error
+
+sealed class ValidateError(
+    cause: Throwable?
+) : AppError(cause) {
+
+    class InvalidEmailError(cause: Throwable?) : ValidateError(cause)
+
+    class InvalidPasswordError(cause: Throwable?) : ValidateError(cause)
```
lydiasama, line 9: We can change this to `data class` to be easier to check its type. 

Ref: https://kotlinlang.org/docs/reference/sealed-classes.html

app/src/main/java/co/nimblehq/ui/second/SecondViewModel.kt
lydiasama: But weird, why `set` function is in `Outputs`?  Should it be in `Inputs` instead if we really want to set it?

app/src/main/java/co/nimblehq/extension/ObservableExt.kt
```diff
@@ -0,0 +1,39 @@
...
+ * An utility method to help preventing the Observable source to emit multiple times (debounce)
+ * within the [DEFAULT_DEBOUNCE_TIME] amount of time.
+ *
+ * Example: This is helpful to <b>prevent multiple click events</b> triggering the emission.
+ *
+ * @param: [scheduler] where to manage the timers that handle timeout for each event
+ * @return: an [Observable] that performs the throttle operation.
+ */
+fun <T> Observable<T>.debounce(scheduler: Scheduler): Observable<T> {
+    return this.throttleFirst(DEFAULT_DEBOUNCE_TIME, TimeUnit.MILLISECONDS, scheduler)
```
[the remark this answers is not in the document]
manh-t: @Tuubz as I understand, basically `throttle` is a specific case for handling which item should be emitted after amount of time, and it is similar to `debounce` somehow. In RxJava, we have `throttleFirst`, `throttleLast`, `throttleLatest` and `throttleLast` is equivalent to `debounce`. Reference: https://proandroiddev.com/throttling-in-rxjava-2-d640ea5f7bf1
So if we rename, could you suggest me a meaning name for this case? 😸

app/src/main/java/co/nimblehq/ui/second/SecondActivity.kt
manh-t: @Tuubz I'll put TODO comment here to refactor naming later 🙆‍♂️

app/src/main/java/co/nimblehq/extension/ObservableExt.kt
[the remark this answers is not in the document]
Tuubz: Hmmm, please correct me if I'm wrong 🙇 , but the reference only says that `throttleWithTimeout` is equivalent to `debounce`. 🙏 

I would rename it to `throttleFirst` 😁

app/build.gradle
```diff
@@ -48,6 +48,27 @@ android {
...
+    productFlavors {
+        production {
+            applicationId "co.nimblehq"
+        }
+
+        staging {
+            applicationId "co.nimblehq.staging"
+        }
+
+        development {
```
Wadeewee, line 61: Should be rename development -> develop 😃

app/build.gradle
```diff
@@ -48,6 +48,27 @@ android {
         }
     }
 
+    flavorDimensions "stage"
+    productFlavors {
+        production {
+            applicationId "co.nimblehq"
+        }
+
+        staging {
```
Wadeewee, line 57: Why do we need a staging variants I think production and develop are cover the product flavor that we have  🤔 What do you think?

app/src/main/java/co/nimblehq/di/modules/RepositoryModule.kt
```diff
@@ -0,0 +1,19 @@
...
+import dagger.Provides
+import dagger.hilt.InstallIn
+import dagger.hilt.android.components.ActivityRetainedComponent
+
+@InstallIn(ActivityRetainedComponent::class)
+@Module
+class RepositoryModule {
+
+    @Provides
+    fun provideApiRepository(apiService: ApiService, scheduler: SchedulerProvider): ApiRepository =
```
Wadeewee, line 17: Format Please 🙏 

    @Provides
    fun provideApiRepository(
        apiService: ApiService, 
        scheduler: SchedulerProvider
    ): ApiRepository = ApiRepositoryImpl(apiService, scheduler)

app/src/main/java/co/nimblehq/extension/ErrorExt.kt
```diff
@@ -0,0 +1,20 @@
+package co.nimblehq.extension
+
+import android.content.Context
+import co.nimblehq.R
+import co.nimblehq.domain.data.error.AppError
+import co.nimblehq.domain.data.error.ValidateError
+
+fun Throwable.userReadableMessage(context: Context, vararg formatArgs: Any?): String {
```
Wadeewee, line 8: How differ between "readableMessage" in AppError & "userReadableMessage" in here ,Do these serve the same purpose? if yes, why do we need to create a new extension 🤔

app/src/main/java/co/nimblehq/extension/ObservableExt.kt
```diff
@@ -0,0 +1,39 @@
...
+ * Subscribe to the Observable<T>, with debounce timeout of 200ms.
+ * When an error happens, the stream isn't stopped, it just prints out the error with [Log.e]
+ */
+fun <T> Observable<T>.debounceAndSubscribe(onNext: (T) -> Unit): Disposable {
+    return this
+        .debounce(mainThread())
+        .flatMapSingle {
+            Single.fromCallable { onNext(it) }
+                .onErrorReturn {
+                    Log.e("debounceAndSubscribe", it.message ?: "Unknown error")
```
Wadeewee, line 32: Use Timber instead Log 🙏

app/src/main/java/co/nimblehq/extension/ObservableExt.kt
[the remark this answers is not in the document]
manh-t: yeah @Tuubz 
image
Will rename 🙏

app/build.gradle
manh-t: @Wadeewee we use noun here ;) `production` -> `development`

app/build.gradle
manh-t: @Wadeewee I think we should keep them in template because they are typical variants for development process, later we can update or remove variants based on each project's requirement 🙏

app/src/main/java/co/nimblehq/extension/ErrorExt.kt
manh-t: `readableMessage` is the message that we get from API.
`userReadableMessage` is the message that we display in UI (if `readableMessage` is `null` or empty we set a general message for it)

manh-t:
Updated 🚀

app/build.gradle
Tuubz: Just curious about Omise & Deemoney project, what variants are you guys using in the project?

manh-t: @Tuubz on DeeMoney, we have `staging` and `production`. cc @luongvo ☝️

app/src/main/java/co/nimblehq/extension/ErrorExt.kt
Wadeewee: Hmm 🤔 readableMessage is **inside** AppError and ValidateError **extend** from **AppError** also that why I think we can have only one(**userReadableMessage/readableMessage**) in AppError because we can handler error message for user readable  in VM and then we just send back a message string to view.

val errorGeneralMessage = context.getString(R.string.error_generic)
    return when (throwable) {
        is ValidateError.InvalidEmailError -> context.getString(R.string.error_email_invalid)
        is ValidateError.InvalidPasswordError -> throwable.message ?: errorGeneralMessage
        else -> errorGeneralMessage
    }

lydiasama (approved):

app/src/main/java/co/nimblehq/extension/ErrorExt.kt
manh-t: @Wadeewee ah I see what you mean, we should check the condition for **AppError** instead of **ValidateError**. I'll put a TODO comment here to fix in later PR because it's out of scope. Thanks for pointing the issue out 😄

app/src/test/java/co/nimblehq/testutil/MockSchedulersProvider.kt
```diff
@@ -1,9 +1,9 @@
 package co.nimblehq.testutil
 
-import co.nimblehq.data.lib.schedulers.SchedulersProvider
+import co.nimblehq.domain.schedulers.BaseSchedulerProvider
 import io.reactivex.schedulers.Schedulers
 
-object MockSchedulersProvider : SchedulersProvider {
+object MockSchedulersProvider : BaseSchedulerProvider {
```
Wadeewee, line 6: What about `TrampolineSchedulerProvider` Can we use it instead `MockSchedulersProvider` 🤔 Correct me if I miss somthing 🙏

manh-t: You are right, they are the same 🙏  Will update

manh-t:
updated 🚀

app/src/main/java/co/nimblehq/extension/RxViewExt.kt
```diff
@@ -0,0 +1,24 @@
...
+import android.util.Log
+import android.view.View
+import androidx.recyclerview.widget.RecyclerView
+import com.jakewharton.rxbinding3.view.clicks
+import io.reactivex.Observable
+import io.reactivex.disposables.Disposable
+
+/**
+ * Subscribe to the [View]'s OnClick event, with debounce timeout of 200ms.
+ * When an error happens, the stream isn't stopped, it just prints out the error with [Log.e]
```
Wadeewee, line 12: [Log.e] - > [Timber.e]

app/src/main/java/co/nimblehq/ui/base/BaseActivity.kt
```diff
@@ -1,19 +1,51 @@
...
-        disposables.clear()
+        disposables.dispose()
+    }
+
+    abstract fun bindViewModel()
+
+    protected fun Disposable.addToDisposables() = addTo(disposables)
+
+    protected inline infix fun <T> Observable<T>.bindTo(crossinline action: (T) -> Unit) {
+        observeOn(AndroidSchedulers.mainThread())
```
Wadeewee, line 42: Should it be `SchedulerProvider().main()` 🙏  same with BaseFragment

manh-t: I think this is unnecessary to initialize `SchedulerProvider` whenever we use `bindTo`. So use static method from `AndroidSchedulers` is still good imo 🙏

app/build.gradle
Tuubz: Yeah, in 7-Eleven we only have two variants as well. Which makes me think it makes more sense to have only 2 instead of 3 🙏  Let's see with Omise team says

app/src/main/java/co/nimblehq/ui/base/BaseActivity.kt
Wadeewee: What about if we inject 🤔 If not work so let's use the same one 🙏 

```
 @Inject
 lateinit var schedulerProvider: SchedulerProvider

 observeOn(schedulerProvider.main())
            .subscribe { action(it) }
            .addToDisposables()
```

manh-t: uhm yeah, it works, updated 🙏

app/build.gradle
luongvo: actually, development and staging are quite similar, we have staging, uat and production in Omise project
I think we could combine dev and staging @manh-t

manh-t:
Removed `development` variant, @Tuubz @Wadeewee please finalize your review 🙇

luongvo (approved):
SGTM

luongvo:
@manh-t @hoangnguyen92dn @lydiasama @Tuubz guys please finalize your review 👨‍🎤

Wadeewee (approved):

PR #54, opened 2020-09-30
Update detekt Config, README, gradle version etc...

sleepylee:
## What happened 🤔
Some minor update on the static code analysis we are using: `detekt`

## Insight 👀
- Update detekt config, use newer version
- Update README

## POW
image

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -61,10 +61,6 @@ buildscript {
     }
 }
 
-plugins {
-    id "io.gitlab.arturbosch.detekt" version "1.0.0.RC6-3"
-}
-
 allprojects {
     repositories {
         google()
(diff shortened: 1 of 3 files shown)


detekt-config.yml
```diff
@@ -1,24 +1,3 @@
...
-    - 'comments'
-  exclude-rules:
-    - 'NamingRules'
-    - 'WildcardImport'
-    - 'MagicNumber'
-    - 'MaxLineLength'
-    - 'LateinitUsage'
-    - 'StringLiteralDuplication'
-    - 'SpreadOperator'
-    - 'TooManyFunctions'
```
luongvo, line 20: 👍

build.gradle
```diff
@@ -78,13 +74,21 @@ task clean(type: Delete) {
...
-        output = "${project.buildDir}/reports/detekt/"
-        parallel = true
-        disableDefaultRuleSets = false
-    }
+configurations {
+    detekt
+ }
+
+dependencies {
+    detekt "io.gitlab.arturbosch.detekt:detekt-cli:1.12.0"
```
luongvo, line 82: Why do we switch to use detekt-cli? 🤔 
Also switch from using `detekt` block to a custom task?

sleepylee: creating a task is easier so you can still run `./gradlew detekt` but also you can run it directly from the gradle file, it looks like this (just like task :clean we had before):
image

sleepylee: I'll recheck whether `detekt-cli` is necessary 🤔  from a setup tutorial I think they used it and I didn't realize that 🤔

sleepylee: updated 🙏

ViktorArtemiev (dismissed):

build.gradle
```diff
@@ -62,7 +62,11 @@ buildscript {
 }
 
 plugins {
-    id "io.gitlab.arturbosch.detekt" version "1.0.0.RC6-3"
+    id "io.gitlab.arturbosch.detekt" version "1.14.0"
+}
+
+repositories {
+    jcenter()
```
luongvo, line 69: @sleepylee do we still need this config as we had it in project gradle already?

sleepylee: I followed their instruction to have a top-level `repositories`🤔  maybe we can try to remove it 
image

manh-t (approved):

build.gradle
luongvo: @sleepylee so, will you remote it? 🤔

sleepylee: yes we can remove it 💯 confirmed and will commit soon

sleepylee:
updated, please review again @nimblehq/android-chapter

manh-t (approved):

luongvo (approved):

PR #55, opened 2020-09-30
Add GitHub Action

sleepylee:
## What happened 🤔
Adding GitHub action for the template, replace bitrise

## Insight 👀
- Configure GitHub Action to do validation on each `push` - note that each project can configure further follow this base config
- Caching support for gradle dependencies artifact/dir

## PoW (a.k.a Proof Of Work)💪
Check GitHub Action:

image

### DIFF
--- a/.github/workflows/lint_and_unit_test.yml
+++ b/.github/workflows/lint_and_unit_test.yml
@@ -5,7 +5,7 @@ on: push
 jobs:
   lint_and_test:
     name: Static code analysis and unit test
-    runs-on: macOS-latest
+    runs-on: ubuntu-latest
     steps:
       - name: Set up JDK 1.8
         uses: actions/setup-java@v1
--- a/.github/workflows/lint_and_unit_test.yml
+++ b/.github/workflows/lint_and_unit_test.yml
@@ -1,4 +1,4 @@
-name: Android Lint & Unit Test flow
+name: Lint and Test
 
 # Choose the branch or branches you want to run this workflow
 on: push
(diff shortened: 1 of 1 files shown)


.github/workflows/lint_and_unit_test.yml
```diff
@@ -0,0 +1,47 @@
+name: Android Lint & Unit Test flow
```
olivierobert, line 1: While we still don't have conventions for workflow naming, I would like to recommend to name workflow using verbs e.g. `test.yml`, `lint.yml`, `distribute.yml`... And second rule, adding the environment if needed e.g. `distribute-staging.yml`, `distribute-production.yml`...

The name attribute can be anything but keeping it short would be great as it can be referenced in other worflows as in:

```
on:
  workflow_run:
    workflows:
      - Test <- this is actually the name attribute 😅 So imagine now if it has to be Android Lint & Unit Test flow
    branches:
      - master
      - development
```

So maybe the names can be just `name: Test` and let the steps describe what it does 💭

sleepylee: @olivierobert for the top `name` attribute I agree we can make it short. 
At first, I separated into multiple `steps` but apparently, we had to run a few duplication setups in each of the steps, so I rolled back to use 1 single step could reduce the overhead of setting up the env, in this case, the `name` of each action should be explicit and readable, so naming it short wouldn't be a good idea. Fyi, I took the reference from google's android project https://github.com/android/compose-samples/blob/master/.github/workflows/Crane.yaml 🙇

sleepylee: workflow name is already verbs imo 🤔  👉  lint & unit test 🤔

manh-t (approved):

luongvo (approved):
lgtm

lydiasama (approved):

sleepylee:
@manh-t could you please merge this PR if there is no further concern? 🙏

Issue #58, opened 2021-01-12
JSON Parser: Update template to use Moshi


minhnimble:
## Issue type ✅
- [x] Improvement request

## The problem 🤔⁉️

Apply to use `Moshi` instead of the old `Gson` as discussed here on Compass 🙏
https://github.com/nimblehq/compass/issues/437 

## Environment, Step to reproduce 💻
 * API 21+

## Further comment, evidence... 🙇🙇
`Moshi` should be integrated successfully as it is the most-voted option 💪
https://github.com/square/moshi

PR #59, opened 2021-01-12
Update the Android Template to use Moshi

minhnimble:
## What happened 🤔
Implement #58 : Remove the old usage of `Gson` for JSON Parser and replace it with `Moshi` - the most-voted option by the team 💪:
 https://github.com/square/moshi

Also, update the Android Template to use the latest versions for some Moshi-related libraries:
- updated `build_gradle_version` from '4.0.1' to '4.1.1'
- updated `retrofit_version` from '2.4.0' to '2.9.0'

## Insight 👀
We can find all the Moshi library's setup instructions for Kotlin from the following section:
https://github.com/square/moshi#kotlin

and sample usages of the library:
https://github.com/square/moshi#moshi

## PoW (a.k.a Proof Of Work)💪
The template is built successfully and all loading data behaviors are working as expected 👌
Screen Shot 2021-01-12 at 11 39 35


app/src/main/java/co/nimblehq/di/modules/RetrofitModule.kt
```diff
@@ -31,9 +31,7 @@ class RetrofitModule {
     }
 
     @Provides
-    fun provideConverterFactory(gson: Gson): Converter.Factory {
-        return ConverterFactoryProvider.getConverterFactoryProvider(gson)
-    }
+    fun provideMoshiConverterFactory(moshi: Moshi): Converter.Factory = ConverterFactoryProvider.getMoshiConverterFactory(moshi)
```
hoangnguyen92dn, line 34: Could we add a break-line after `=` to reduce the length of the single line?

data/src/main/java/co/nimblehq/data/service/providers/MoshiBuilderProvider.kt
```diff
@@ -0,0 +1,16 @@
...
+import com.squareup.moshi.adapters.Rfc3339DateJsonAdapter
+import com.squareup.moshi.kotlin.reflect.KotlinJsonAdapterFactory
+import java.util.*
+
+object MoshiBuilderProvider {
+
+    val moshiBuilder: Moshi.Builder
+        get() {
+            return Moshi.Builder()
+                .add(Date::class.java, Rfc3339DateJsonAdapter())
```
hoangnguyen92dn, line 13: As I can see, we don't have `Date` in Data class so it's unnecessary. But still thinking about keep it there for usage usage if needed 🤔

data/src/main/java/co/nimblehq/data/service/response/ErrorResponse.kt
```diff
@@ -1,8 +1,7 @@
 package co.nimblehq.data.service.response
 
-import com.google.gson.annotations.SerializedName
+import com.squareup.moshi.Json
 
 data class ErrorResponse(
-    @SerializedName("message")
-    val message: String
+    @Json(name = "message") val message: String
```
hoangnguyen92dn, line 6: Should we keep the old format with annotation and val separate line? 🤔

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,7 +1,6 @@
-#Fri Sep 18 13:17:45 ICT 2020
+#Tue Jan 12 10:52:22 ICT 2021
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.1.1-all.zip
-org.gradle.parallel=true
```
hoangnguyen92dn, line 7: Hmm. Why do we disable parallel on gradle build? 🤔

data/src/main/java/co/nimblehq/data/service/providers/MoshiBuilderProvider.kt
minhnimble: yup, I keep it here for later usage purpose if needed 👌

gradle/wrapper/gradle-wrapper.properties
minhnimble: hmm my AS automatically updates it when I the new gradle version. Maybe it is not needed anymore, let me check again 🙏

data/src/main/java/co/nimblehq/data/service/response/ErrorResponse.kt
minhnimble: hmm, I think in new projects we implement them in one-line like DeeMoney so I guess it is fine to keep. What do you think? 🤔

data/src/main/java/co/nimblehq/data/service/providers/MoshiBuilderProvider.kt
```diff
@@ -0,0 +1,16 @@
...
+import com.squareup.moshi.Moshi
+import com.squareup.moshi.adapters.Rfc3339DateJsonAdapter
+import com.squareup.moshi.kotlin.reflect.KotlinJsonAdapterFactory
+import java.util.*
+
+object MoshiBuilderProvider {
+
+    val moshiBuilder: Moshi.Builder
+        get() {
+            return Moshi.Builder()
```
luongvo, line 12: we could get rid of return by `get() = Moshi.Builder()...`

data/src/main/java/co/nimblehq/data/service/response/ErrorResponse.kt
luongvo: I'm not a fan of keeping all things in just 1 line 🤪 
for example
```
@Inject
lateinit var navigator: MainNavigator
```
or
```
@Json(name = "message") 
val message: String
@Json(name = "error_code") 
val errorCode: Int
```
looks cleaner and easy to look for all fields vertically

luongvo (commented):
the rest looks good to me, thanks @minhnimble

data/src/main/java/co/nimblehq/data/service/providers/MoshiBuilderProvider.kt
minhnimble: nice refactor, let me update 👍

luongvo (approved):
sgtm 🏁

hoangnguyen92dn (approved):
LGTM

manh-t (approved):
LGTM

PR #62, opened 2021-01-13
[Feature] Optimize loading handling by adding `doShowLoading` operator

luongvo:
## What happened 🤔
- Currently, whenever a task is being subscribed, we always try to show and hide the loading by `doOnSubscribe` and `doFinally` (or sometimes `doOnError`). 

For example:
```
getExampleDataUseCase
            .execute(Unit)
            .doOnSubscribe { _showLoading.onNext(true) }
            .toObservable()
```
```
_refresh
            .flatMap { fetchApi() }
            .subscribeBy(
                onNext = {
                    _data.onNext(it)
                    _showLoading.onNext(false)
                },
                onError = _error::onNext
            )
            .addToDisposables()
```

This way will inject multiple items into `_showLoading` and made our tests hard to check and messy.
```
loadingObserver
            .assertNoErrors()
            .assertValues(false, true, true, true, false, false, false)
```
- It also makes us hard to manage to show or hide loading when there are multiple parallel running tasks in VM

## Insight 👀
- Keep a counted object of `_showLoading` state, only show the loading if there is no loading triggered yet and hide loading if there is only 1 loading.

In `BaseViewModel`
```
private var loadingCount: Int = 0

    protected fun showLoading() {
        if (loadingCount == 0) {
            _showLoading.onNext(true)
        }
        loadingCount++
    }

    protected fun hideLoading() {
        loadingCount--
        if (loadingCount == 0) {
            _showLoading.onNext(false)
        }
    }
```

### DIFF
--- a/app/src/main/java/co/nimblehq/ui/base/BaseViewModel.kt
+++ b/app/src/main/java/co/nimblehq/ui/base/BaseViewModel.kt
@@ -37,7 +37,7 @@ abstract class BaseViewModel : ViewModel() {
         get() = _navigator
 
     /**
-     * To show loading manually
+     * To show loading manually, should call `hideLoading` after
      */
     protected fun showLoading() {
         if (loadingCount == 0) {
@@ -47,7 +47,7 @@ abstract class BaseViewModel : ViewModel() {
     }
 
     /**
-     * To hide loading manually
+     * To hide loading manually, should be called after `showLoading`
      */
     protected fun hideLoading() {
         loadingCount--
--- a/app/src/main/java/co/nimblehq/ui/base/BaseViewModel.kt
+++ b/app/src/main/java/co/nimblehq/ui/base/BaseViewModel.kt
@@ -2,21 +2,25 @@ package co.nimblehq.ui.base
 
 import androidx.lifecycle.ViewModel
 import co.nimblehq.lib.IsLoading
+import io.reactivex.Completable
+import io.reactivex.Flowable
 import io.reactivex.Observable
+import io.reactivex.Single
 import io.reactivex.disposables.CompositeDisposable
 import io.reactivex.disposables.Disposable
 import io.reactivex.rxkotlin.addTo
 import io.reactivex.subjects.BehaviorSubject
 import io.reactivex.subjects.PublishSubject
 
-@Suppress("VariableNaming")
+@Suppress("PropertyName")
 abstract class BaseViewModel : ViewModel() {
 
-    protected val _showLoading = BehaviorSubject.create<IsLoading>()
+    private val _showLoading = BehaviorSubject.createDefault(false)
     protected val _error = BehaviorSubject.create<Throwable>()
     protected val _navigator = PublishSubject.create<NavigationEvent>()
 
     private val disposables by lazy { CompositeDisposable() }
+    private var loadingCount: Int = 0
 
     override fun onCleared() {
         super.onCleared()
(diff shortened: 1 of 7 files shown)


detekt-config.yml
```diff
@@ -109,7 +109,7 @@ empty-blocks:
   EmptyForBlock:
     active: true
   EmptyFunctionBlock:
-    active: true
+    active: false
```
hoangnguyen92dn, line 112: Wait... Why do we need to disable this validation?

detekt-config.yml
```diff
@@ -213,7 +213,7 @@ naming:
     minimumVariableNameLength: 1
   VariableNaming:
     active: true
-    variablePattern: '[a-z][A-Za-z0-9]*'
+    variablePattern: '(_)?[a-z][A-Za-z0-9]*'
```
hoangnguyen92dn, line 216: 🤔  We don't use `_` as the prefix for a non-private variable.

app/src/main/java/co/nimblehq/ui/base/BaseViewModel.kt
```diff
@@ -32,5 +36,46 @@ abstract class BaseViewModel : ViewModel() {
...
+        if (loadingCount == 0) {
+            _showLoading.onNext(true)
+        }
+        loadingCount++
+    }
+
+    /**
+     * To hide loading manually, should be called after `showLoading`
+     */
+    protected fun hideLoading() {
```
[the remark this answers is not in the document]
luongvo: hmm, basically yes, but I tend to keep them as `protected` for manual usages in case we don't like to use `doShowLoading()` operator 💭  One use case for manual usages is: webview loading between page navigations. 🤓

detekt-config.yml
luongvo: correct, fixed

luongvo: @hoangnguyen92dn we need this for protected fields https://github.com/nimblehq/android-templates/pull/62/files#diff-0cbbe6c8fd6798d5dc14330d3a7012d4a92a30727ddfb72849c96da3b291a982R19-R20 🤔

detekt-config.yml
luongvo: @hoangnguyen92dn To allow empty function blocks (such as overriding) https://github.com/nimblehq/android-templates/blob/da9680602cf4bb08989c5ef97ac6f94c7ec6a4e5/app/src/main/java/co/nimblehq/ui/screens/MainActivity.kt#L15 🤓

hoangnguyen92dn (approved):

detekt-config.yml
hoangnguyen92dn: 🆗

Tuubz (approved):

minhnimble (approved):
The rest LGTM 🚀

app/src/main/java/co/nimblehq/ui/base/BaseViewModel.kt
```diff
@@ -32,5 +36,46 @@ abstract class BaseViewModel : ViewModel() {
...
+        loadingCount--
+        if (loadingCount == 0) {
+            _showLoading.onNext(false)
+        }
+    }
```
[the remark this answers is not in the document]
luongvo: @minhnimble possible, but the main purpose of this PR is trying to build a common mechanism to show/hide loading progress, I'm wondering about the `forceHideLoading` method will break the rule, instead of trying to figure out whether we missed any show/hide actions, we try to use that method to speed up the dev progress.

hoangnguyen92dn (approved):

minhnimble (approved):

PR #74, opened 2021-04-26
Update the Android Template with new test cases design convention

### LINKED ISSUE #73: Testing Convention: Update template test case design
## Issue type ✅
- [ ] Bug
- [ ] Feature request
- [x] Improvement request

## The problem 🤔⁉️

Improve test case design convention by applying the `Method B: Create multiple test functions that each of them will have a specific description message and one assertion for each observer used per test function` as we discussed and agreed on Compass 🙏: 
https://github.com/nimblehq/compass/issues/585

## Environment, Step to reproduce 💻
 * API 21+

## Further comment, evidence... 🙇🙇

As expressed from the Compass link above ☝️

minhnimble:
## What happened 🤔

Resolve #73 : Improve test case design convention by applying the `Method B: Create multiple test functions that each of them will have a specific description message and one assertion for each observer used per test function` - the most-voted option by the team 💪

## Insight 👀
- Add more detailed descriptions for all existing test functions for correctly reflecting what is being tested in each function.
- Split some test functions that have multiple assertions into each function per assertion.
- Add some small refactors to clean up some warnings.
- Update `build_gradle_version` to latest existing version: `4.2.0-rc01`

## PoW (a.k.a Proof Of Work)💪
The template and all the updated tests are built and running successfully 👌
Screen Shot 2021-04-26 at 16 14 59

### DIFF
--- a/RxJavaTemplate/app/src/test/java/co/nimblehq/rxjava/ui/screens/MainNavigatorTest.kt
+++ b/RxJavaTemplate/app/src/test/java/co/nimblehq/rxjava/ui/screens/MainNavigatorTest.kt
@@ -32,7 +32,7 @@ class MainNavigatorTest {
     }
 
     @Test
-    fun `When calling navigate responds successfully with destination as HomeFragment, it should navigate to Second Screen from Home Screen correctly`() {
+    fun `When navigating to Second screen from Home screen, it navigates with corresponding bundle`() {
         When calling mockDestination.id itReturns R.id.homeFragment
 
         val bundle = SecondBundle(MockUtil.dataList[0])
@@ -46,7 +46,7 @@ class MainNavigatorTest {
     }
 
     @Test
-    fun `Should navigate to WebView screen from Second correctly`() {
+    fun `When navigating to WebView screen from Second screen, it navigates with corresponding bundle`() {
         When calling mockDestination.id itReturns R.id.secondFragment
 
         val bundle = WebViewBundle("url")
--- a/RxJavaTemplate/app/src/test/java/co/nimblehq/rxjava/ui/screens/MainNavigatorTest.kt
+++ b/RxJavaTemplate/app/src/test/java/co/nimblehq/rxjava/ui/screens/MainNavigatorTest.kt
@@ -32,7 +32,7 @@ class MainNavigatorTest {
     }
 
     @Test
-    fun `Should navigate to Second screen from Home correctly`() {
+    fun `When calling navigate responds successfully with destination as HomeFragment, it should navigate to Second Screen from Home Screen correctly`() {
         When calling mockDestination.id itReturns R.id.homeFragment
 
         val bundle = SecondBundle(MockUtil.dataList[0])
--- a/RxJavaTemplate/app/src/test/java/co/nimblehq/rxjava/ui/screens/home/HomeViewModelTest.kt
+++ b/RxJavaTemplate/app/src/test/java/co/nimblehq/rxjava/ui/screens/home/HomeViewModelTest.kt
@@ -32,7 +32,7 @@ class HomeViewModelTest {
     }
 
     @Test
-    fun `When initializing HomeViewModel, it doesn't emit to show any loading indicator as the value is false`() {
+    fun `When initializing HomeViewModel, it doesn't emit to show any loading indicator, defaulting to false`() {
         val loadingObserver = viewModel.showLoading.test()
 
         loadingObserver
--- a/RxJavaTemplate/app/src/test/java/co/nimblehq/rxjava/ui/screens/home/HomeViewModelTest.kt
+++ b/RxJavaTemplate/app/src/test/java/co/nimblehq/rxjava/ui/screens/home/HomeViewModelTest.kt
@@ -23,7 +23,7 @@ class HomeViewModelTest {
     }
 
     @Test
-    fun `When initializing HomeViewModel, it emits the first data correspondingly`() {
+    fun `When initializing the view model, it emits the first data correspondingly`() {
         val dataObserver = viewModel.data.test()
 
         dataObserver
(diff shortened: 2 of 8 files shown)


luongvo:
@minhnimble conflict 😢

minhnimble:
@luongvo no worries, I will update this PR accordingly since you modified the tests a bit. Let me update your new tests to match our new convention as well 🙏

RxJavaTemplate/build.gradle
```diff
@@ -2,7 +2,7 @@
 
 buildscript {
     ext {
-        build_gradle_version = '4.1.1'
+        build_gradle_version = '4.2.0-rc01'
```
luongvo, line 5: I don't think we should consider using beta version, especially for the Template project. The reason for this it's just because we're using beta AS in some project ❌

RxJavaTemplate/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -3,5 +3,5 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.5-all.zip
+distributionUrl=https://services.gradle.org/distributions/gradle-6.7.1-all.zip
```
luongvo, line 6: same

RxJavaTemplate/app/src/test/java/co/nimblehq/rxjava/ui/screens/MainNavigatorTest.kt
```diff
@@ -33,7 +33,7 @@ class MainNavigatorTest {
     }
 
     @Test
-    fun `Should navigate to Second screen from Home correctly`() {
+    fun `When calling navigate responds successfully with destination as HomeFragment, it should navigate to Second Screen from Home Screen correctly`() {
```
luongvo, line 36:
```suggestion
    fun `When navigating to Second screen from Home screen, it should navigate with corresponding bundle`() {
```
the condition and result are quite similar, I think we could omit it but still keep following the correct format

RxJavaTemplate/app/src/test/java/co/nimblehq/rxjava/ui/screens/home/HomeViewModelTest.kt
```diff
@@ -20,45 +20,53 @@ class HomeViewModelTest {
...
         When calling mockGetExampleDataUseCase.execute(any()) itReturns Single.just(MockUtil.data)
-        viewModel = HomeViewModel(
-            mockGetExampleDataUseCase
-        )
+        viewModel = HomeViewModel(mockGetExampleDataUseCase)
     }
 
     @Test
-    fun `When initializing, it should emit first data loading`() {
+    fun `When initializing HomeViewModel, it emits the first data correspondingly`() {
```
luongvo, line 27: the viewmodel name could be in a common name for all test classes
```suggestion
    fun `When initializing the view model, it emits the first data correspondingly`() {
```

RxJavaTemplate/app/src/test/java/co/nimblehq/rxjava/ui/screens/home/HomeViewModelTest.kt
```diff
@@ -20,45 +20,53 @@ class HomeViewModelTest {
...
         val dataObserver = viewModel.data.test()
-        val loadingObserver = viewModel.showLoading.test()
 
         dataObserver
             .assertValueCount(1)
             .assertValue(MockUtil.data)
+    }
+
+    @Test
+    fun `When initializing HomeViewModel, it doesn't emit to show any loading indicator, defaulting to false`() {
```
luongvo, line 36: same

RxJavaTemplate/app/src/test/java/co/nimblehq/rxjava/ui/screens/home/HomeViewModelTest.kt
```diff
@@ -20,45 +20,53 @@ class HomeViewModelTest {
...
+        val loadingObserver = viewModel.showLoading.test()
 
         loadingObserver
             .assertNoErrors()
-            .assertValues(false)
+            .assertValue(false)
```
luongvo, line 41: from my understanding here, we can't test the loading states correctly here because of `init` block executing. Should we apply the same way to use `TestScheduler` in DM project here to have a better test, the expected result should be `false, true, false`?

RxJavaTemplate/app/src/test/java/co/nimblehq/rxjava/ui/screens/home/HomeViewModelTest.kt
```diff
@@ -20,45 +20,53 @@ class HomeViewModelTest {
...
+
+    @Test
+    fun `When calling refresh regardless of success or failure, it emits to showLoading as true then false`() {
+        val loadingObserver = viewModel.showLoading.test()
+
+        viewModel.input.refresh()
 
         loadingObserver
             .assertNoErrors()
             .assertValues(false, true, false)
```
luongvo, line 63: then this one should be `false, true, false, true, false` 😛

RxJavaTemplate/app/src/test/java/co/nimblehq/rxjava/ui/screens/home/HomeViewModelTest.kt
luongvo: and we need to update the test case name 😛

RxJavaTemplate/build.gradle
minhnimble: ah okay

RxJavaTemplate/app/src/test/java/co/nimblehq/rxjava/ui/screens/home/HomeViewModelTest.kt
minhnimble: hmm, I think this requirement is out of this ticket's scope man, maybe we can do that as a separate convention or improvement for sure 👍

luongvo: Okay, let's do it in a separate PR 🆗

luongvo (approved):

chornerman (approved):
lgtm

hoangnguyen92dn (approved):
LGTM

PR #80, opened 2021-05-28
Add UI tests to RxJavaTemplate

chornerman:
https://github.com/nimblehq/android-templates/issues/95

## What happened 🤔
Add UI tests to RxJavaTemplate (using `Espresso`) ☕ 

## Insight 👀
- <del>We might have to consider adding the naming convention for UI test functions since currently, our Compass doesn't seem to have one yet. 🙏🏼 </del>
**Updated:** Follow the convention of [this](https://github.com/android/architecture-samples) project

- <del>`IdlingResource` is being used in order to simulate behavior of loading the real data, which will have some delay time and we have to test that the app handles it correctly. When we register `IdlingResource`, the test will be waiting until there is no `IdlingResource` left, then proceeding with the below codes. So that we need to add `IdlingResource` in our logic for getting the data by incrementing the `IdlingResource` when start retrieving and decrementing it when all data have been retrieved. By the way, I'm admitted that it seems a little ugly when putting this kind of code into production. 😅 </del>
**Updated:** Using `Thread.sleep()` instead since we know exactly how much time the API would take to fetch the data and to get rid of `IdlingResource` codes in production. 🚀 

- In order to make the tests compatible with `Hilt`, we have to use custom `launchFragment` function instead of normal `launchFragmentInContainer`, which will require `EmptyHiltActivity` (in `debug` directory). You can find the reference [here](https://github.com/android/architecture-samples/blob/dev-hilt/app/src/androidTest/java/com/example/android/architecture/blueprints/todoapp/HiltExt.kt). We also need to create a custom `TestRunner` to archive the same goal.

## PoW (a.k.a Proof Of Work)💪
✅ All tests are passed.


sleepylee:
@chornerman please write up a bit on the description for everyone to catch up quickly on what is being covered here 🙏  , for example, modifying ViewModel logic for the `IdlingResource` to serve the UI Testing purpose is something we need to understand why you introduced it.

chornerman:
@sleepylee Updated 🙏🏼

RxJavaTemplate/app/src/androidTest/java/co/nimblehq/rxjava/repository/TestApiRepositoryImpl.kt
```diff
@@ -0,0 +1,15 @@
...
+import co.nimblehq.rxjava.domain.repository.ApiRepository
+import co.nimblehq.rxjava.domain.test.MockUtil
+import io.reactivex.Single
+import java.util.concurrent.TimeUnit
+import javax.inject.Inject
+
+class TestApiRepositoryImpl @Inject constructor() : ApiRepository {
+
+    override fun exampleData(): Single<List<Data>> {
+        return Single.just(MockUtil.dataList).delay(3, TimeUnit.SECONDS)
```
manh-t, line 13: delay 3 secs is too long, maybe we can decrease to 300 millisecond 😅

RxJavaTemplate/app/src/androidTest/java/co/nimblehq/rxjava/ui/screens/home/HomeFragmentTest.kt
```diff
@@ -0,0 +1,159 @@
...
+        onView(allOf(withId(R.id.tvDataAuthor), withText("author3"))).check(matches(isDisplayed()))
+
+        unregisterIdlingResource()
+    }
+
+    @Test
+    fun when_clicking_on_refresh_button_should_refresh_data() {
+        registerIdlingResource()
+
+        unregisterIdlingResource()
```
manh-t, line 73: redundant call?

RxJavaTemplate/app/src/androidTest/java/co/nimblehq/rxjava/ui/screens/home/HomeFragmentTest.kt
```diff
@@ -0,0 +1,159 @@
...
+            .withNoTimeout()
+            .check(webMatches(getCurrentUrl(), containsString("google.com")))
+
+        pressBack()
+
+        onView(withId(R.id.ivSecondThumbnail)).check(matches(isDisplayed()))
+
+        pressBack()
+
+        onView(withId(R.id.rvHomeData)).check(matches(isDisplayed()))
```
manh-t, line 141: maybe we forget `unregisterIdlingResource()` 🤔

RxJavaTemplate/app/src/androidTest/java/co/nimblehq/rxjava/ui/screens/home/HomeFragmentTest.kt
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    private fun registerIdlingResource() {
+        IdlingRegistry.getInstance().register(IdlingResource.countingIdlingResource)
+    }
+
+    private fun unregisterIdlingResource() {
+        IdlingRegistry.getInstance().unregister(IdlingResource.countingIdlingResource)
+    }
```
manh-t, line 150: maybe we can create helper function then putting it in helper class. How do you think? 🙏 
```suggestion
    private fun setIdlingResource(onIdling: () -> Unit) {
        IdlingRegistry.getInstance().register(IdlingResource.countingIdlingResource)
        onIdling.invoke()
        IdlingRegistry.getInstance().unregister(IdlingResource.countingIdlingResource)
    }
```
then calling:
```dart
setIdlingResource {
  onView(..)...
}
```

RxJavaTemplate/app/src/main/java/co/nimblehq/rxjava/ui/screens/home/HomeViewModel.kt
```diff
@@ -42,11 +43,15 @@ class HomeViewModel @ViewModelInject constructor(
...
+        IdlingResource.increment()
         getExampleDataUseCase
             .execute(Unit)
             .doShowLoading()
             .subscribeBy(
-                onSuccess = _data::onNext,
+                onSuccess = {
+                    _data.onNext(it)
+                    IdlingResource.decrement()
+                },
```
manh-t, line 54: why do we need to add `IdlingResource` functions here? Shouldn't it be called in test?

RxJavaTemplate/app/src/androidTest/java/co/nimblehq/rxjava/ui/screens/home/HomeFragmentTest.kt
chornerman: Exactly 😆 Thank you 🙏🏼

RxJavaTemplate/app/src/androidTest/java/co/nimblehq/rxjava/ui/screens/home/HomeFragmentTest.kt
chornerman: No, The point of this is to test how the app handles the loading state. When we register IdlingResource, the test will stop running until there's no IdlingResource left, which this case is when we finished getting data. After that, we unregister IdlingResource which means we don't care about delay time when getting data anymore. If we don't unregister it at that point when we start clicking on the refresh button the test will hang until the data is finished loading. At that time the progress bar should be disappeared already which means we unable to test the progress bar while it's showing. Anyway, if you can think of a less obscure way please tell me. 😄

RxJavaTemplate/app/src/androidTest/java/co/nimblehq/rxjava/ui/screens/home/HomeFragmentTest.kt
chornerman: Agree krub, now it a lot cleaner 🙏🏼

RxJavaTemplate/app/src/androidTest/java/co/nimblehq/rxjava/ui/screens/second/SecondFragmentTest.kt
```diff
@@ -0,0 +1,90 @@
...
+        launchFragment()
+    }
+
+    @After
+    fun teardown() {
+        Intents.release()
+    }
+
+    @Test
+    fun when_initialize_should_shows_ui_correctly() {
```
sleepylee, line 59: just like unit test, I think you can add `@Suppress("IllegalIdentifier")` from the beginning to write the function name with `space` instead of `_` no? 🤔

RxJavaTemplate/app/src/main/java/co/nimblehq/rxjava/ui/screens/home/HomeViewModel.kt
sleepylee: also a ViewModel is usually containing only the real resources, when you want to run UI test or integration test with devices, we shouldn't duplicate what we already had in Unit Test, so you might actually want to bring these part to the View (Fragment) instead of this ViewModel logic code.

RxJavaTemplate/app/src/androidTest/java/co/nimblehq/rxjava/ui/screens/second/SecondFragmentTest.kt
chornerman: It's not working. Naming the test with blank space will result in build failure with this error
`com.android.tools.r8.errors.b: Space characters in SimpleName 'when initialize_should_shows_ui_correctly' are not allowed prior to DEX version 040` 😅

RxJavaTemplate/app/src/main/java/co/nimblehq/rxjava/ui/screens/home/HomeViewModel.kt
chornerman: Updated 🙏🏼

RxJavaTemplate/app/src/androidTest/java/co/nimblehq/rxjava/repository/TestApiRepositoryImpl.kt
chornerman: 300 millisecond is too short 😅 One of the tests (which using `IdlingResource`) failed with this value. So I changed it to 500 for now. 🙏🏼

RxJavaTemplate/app/src/androidTest/java/co/nimblehq/rxjava/ui/screens/webview/WebViewFragmentTest.kt
```diff
@@ -0,0 +1,45 @@
...
+    fun setup() {
+        hiltRule.inject()
+        launchFragment()
+    }
+
+    @Test
+    fun when_initialize_should_shows_ui_correctly() {
+        onWebView()
+            .withNoTimeout()
+            .check(webMatches(getCurrentUrl(), containsString("google.com")))
```
hoangnguyen92dn, line 35: Should we validate if it matches the URL rathen than `containsString`? 🤔

RxJavaTemplate/app/src/debug/AndroidManifest.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    package="co.nimblehq.rxjava">
+
+    <application>
+        <activity
+            android:name=".EmptyHiltActivity"
+            android:exported="false" />
+    </application>
```
hoangnguyen92dn, line 9: Hmm. I am not sure if you tried to config the `sourceSets` in `gradle` rather than create debug folder for testing purpose 🤔  FYI: https://stackoverflow.com/a/56951415

RxJavaTemplate/app/src/debug/java/co/nimblehq/rxjava/EmptyHiltActivity.kt
```diff
@@ -0,0 +1,7 @@
+package co.nimblehq.rxjava
+
+import androidx.appcompat.app.AppCompatActivity
+import dagger.hilt.android.AndroidEntryPoint
+
+@AndroidEntryPoint
+class EmptyHiltActivity : AppCompatActivity()
```
hoangnguyen92dn, line 7: Same here. Let's try to use `sourceSets` instead 🙏

RxJavaTemplate/app/src/staging/java/co/nimblehq/rxjava/IdlingResource.kt
```diff
@@ -0,0 +1,21 @@
+package co.nimblehq.rxjava
+
+import androidx.test.espresso.idling.CountingIdlingResource
+
+object IdlingResource {
```
hoangnguyen92dn, line 5: I think if we could get rid of duplicated on `staging` and `production` by using `sourceSets`

RxJavaTemplate/app/src/main/java/co/nimblehq/rxjava/ui/screens/home/HomeFragment.kt
```diff
@@ -89,5 +90,7 @@ class HomeFragment : BaseFragment<FragmentHomeBinding>() {
     private fun showLoading(isLoading: IsLoading) {
         binding.btHomeRefresh.isEnabled = !isLoading
         viewLoadingBinding.pbLoading.visibleOrGone(isLoading)
+
+        IdlingResource.increaseOrDecrease(isLoading)
```
hoangnguyen92dn, line 94: Should we mark this `VisibleForTesting`?

RxJavaTemplate/app/src/debug/AndroidManifest.xml
chornerman: I'm unable to use `sourceSets` to move `AndroidManifest.xml` and `EmptyHiltActivity` to `androidTest` folder instead of `debug` 😅 I try every possible solution I found on the internet but none of which work as I always end up with unable to find Activity error, please let me know if I misunderstand something or if you have other solutions for this. 🙏🏼

RxJavaTemplate/app/src/main/java/co/nimblehq/rxjava/ui/screens/home/HomeFragment.kt
chornerman: Please correct me if I'm wrong but I'm not quite sure this is necessary 🙏🏼  Could you please tell me where exactly you think we should add this annotation?

RxJavaTemplate/app/src/androidTest/java/co/nimblehq/rxjava/ui/screens/webview/WebViewFragmentTest.kt
chornerman: Resolved, Thank you 🙏🏼

RxJavaTemplate/app/src/debug/AndroidManifest.xml
hoangnguyen92dn: @chornerman I think you missed the step add `androidTest.java.srcDirrs` in `build.gradle(:app)` 🤔  I've just tried and it works. Please try on your side and let me know if it works 🙏 

- Remove debug folder and move the files to `androidTest` folder

image

- Add `sourceSets` in `build.gradle(:app)`

image

RxJavaTemplate/app/src/main/java/co/nimblehq/rxjava/ui/screens/home/HomeFragment.kt
hoangnguyen92dn: @chornerman About `@VisibleForTesting`, please refer [this](https://developer.android.com/reference/kotlin/androidx/annotation/VisibleForTesting)

One more thing, I see you created `IdlingResource` on `staging` and `production`, it's quite redundant IMO. 
Here is the other approach from Android official [testing-sample](https://github.com/android/testing-samples/blob/456eadca14a5cd64e8787deaa4e4cec25ca27587/ui/espresso/IdlingResourceSample/app/src/main/java/com/example/android/testing/espresso/IdlingResourceSample/MainActivity.java#L44)
We can define `IdlingResource` as nullable and skip it to leave it as null in production code.

RxJavaTemplate/app/src/debug/AndroidManifest.xml
chornerman: It's still not working 😢 
Screenshot 2021-06-17 at 12 26 06

hoangnguyen92dn: @chornerman Ah. I executed test on `HomeFragmentTest`, didn't notice that `HomeFragmentTest` you are using `launchActivity` instead of `launchFragment`. As they commented from their sample, we have to put these files in `debug` folder 😞 

I'm curious why `HomeFragmentTest` but we launch activity instead of launch fragment as `SecondFragmentTest` and `WebViewFragmentTest`

chornerman: @hoangnguyen92dn Actually I follow [this](https://github.com/android/architecture-samples/blob/dev-hilt/app/src/androidTest/java/com/example/android/architecture/blueprints/todoapp/tasks/TasksFragmentTest.kt) approach. As I can recall when we call `launchFragment()`, it will launch the activity first and then open that fragment, but with `HomeFragment` it will launch the activity which will also open `HomeFragment` already since it's the main fragment and then open `HomeFragment` again as we call `launchFragment()` which will cause duplicate id problem when trying to match the view id 🙏🏼

chornerman:
@hoangnguyen92dn @manh-t After discussion with @sleepylee I removed the `IdlingResource` completely and use `Thread.sleep()` instead since we already know exactly how much time the API would take for fetching the data. Please let me know how you guys think of this approach. 🙏🏼

RxJavaTemplate/app/src/androidTest/java/co/nimblehq/rxjava/repository/TestApiRepositoryImpl.kt
chornerman: Resolved

manh-t:
@chornerman I'm okay with that, at least we don't add test code to production 🙆‍♂️

lydiasama (commented):
I have an idea to declare a repository for storing each screen assertion and action. It's useful when we will do the UI-regression test also. 

Example: 
```kotlin
object ScreenA {

    fun verifyHasSubmitButton() { onView(.....).check(....) }

    fun clickSubmitButton() { onView(....).perform(click()) }

    ...
}
```

and inside of the UI Test it will be

```kotlin
@Test
fun when_click_on_submit_button..() {
    with(ScreenA) {
        verifyHasSubmitButton()
        clickSubmitButton()
    }
}
```

Imagine in the regression test:
```kotlin
@Test
fun test_integration_submision_flow() {
    with(ScreenA) {
        verifyHasSubmitButton()
        clickSubmitButton()
    }

    with(ScreenB) {
        verifyOfScreenB()
        actionOfScreenB()
    }

    with(ScreenC) {
        verifyOfScreenC()
        actionOfScreenC()
    }
}
```

What do you think?

RxJavaTemplate/app/src/androidTest/java/co/nimblehq/rxjava/ui/screens/home/HomeFragmentTest.kt
```diff
@@ -0,0 +1,145 @@
...
+
+            onView(withId(R.id.ivSecondThumbnail)).check(matches(isDisplayed()))
+
+            pressBack()
+
+            onView(withId(R.id.rvHomeData)).check(matches(isDisplayed()))
+        }
+    }
+
+    private fun waitForApi(action: () -> Unit) {
```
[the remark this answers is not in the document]
chornerman: Resolved 🙏🏼

RxJavaTemplate/app/src/debug/AndroidManifest.xml
chornerman: Refactored by removing tests that needed to navigate to other screens to `AppFlowTest`. Now `HomeFragmentTest` use `launchFragment` like the others 🙏🏼

RxJavaTemplate/app/src/main/java/co/nimblehq/rxjava/ui/screens/home/HomeFragment.kt
chornerman: Use Thread.sleep() instead 🙏🏼

RxJavaTemplate/app/src/androidTest/java/co/nimblehq/rxjava/ui/screens/home/HomeFragmentTest.kt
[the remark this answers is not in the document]
chornerman: Updated structure accordingly. Thanks for your advice 😸

chornerman:
@hoangnguyen92dn @manh-t @sleepylee @lydiasama The repo has been updated...
- No more `IdlingResource` code in production as we use `Thread.sleep()` instead
- Follow the test function naming convention of [Android architecture-samples](https://github.com/android/architecture-samples) project
- Extract tests that required opening several fragments to `AppFlowTest`. All tests contain in the fragment test class are executed only in that fragment. So call `launchFragment` in each fragment test class and call `lauchActivity` only in flow test class
- Use @lydiasama suggested structure, which is the same we did in `DeeMoney`
Please take a look again when you guys have time 🙏🏼

luongvo (approved):
Great work @chornerman, lgtm

lydiasama (approved):
Nice work. 🎉  LGTM.

RxJavaTemplate/app/src/androidTest/java/co/nimblehq/rxjava/ui/screens/home/HomeFragmentTest.kt
```diff
@@ -0,0 +1,50 @@
...
+    lateinit var testApiRepository: ApiRepository
+
+    @Before
+    fun setup() {
+        hiltRule.inject()
+        launchFragment()
+    }
+
+    @Test
+    fun showUiCorrectly() {
```
manh-t, line 31: need to define a context. How about? 🙏 
```suggestion
    fun initData_showUiCorrectly() {
```

RxJavaTemplate/app/src/androidTest/java/co/nimblehq/rxjava/ui/screens/second/SecondFragmentTest.kt
```diff
@@ -0,0 +1,69 @@
...
+
+    @Before
+    fun setup() {
+        hiltRule.inject()
+        Intents.init()
+        launchFragment()
+    }
+
+    @Test
+    fun showUiCorrectly() {
```
manh-t, line 42: same here 
```suggestion
    fun prefillData_showUiCorrectly() {
```

RxJavaTemplate/app/src/androidTest/java/co/nimblehq/rxjava/ui/screens/webview/WebViewFragmentTest.kt
```diff
@@ -0,0 +1,38 @@
...
+    var hiltRule = HiltAndroidRule(this)
+
+    @Before
+    fun setup() {
+        hiltRule.inject()
+        launchFragment()
+    }
+
+    @Test
+    fun showUiCorrectly() {
```
manh-t, line 27:
```suggestion
    fun openWebView_showUiCorrectly() {
```

RxJavaTemplate/app/src/main/java/co/nimblehq/rxjava/di/modules/RepositoryModule.kt
```diff
@@ -6,9 +6,9 @@ import co.nimblehq.rxjava.domain.repository.ApiRepositoryImpl
 import dagger.Module
 import dagger.Provides
 import dagger.hilt.InstallIn
-import dagger.hilt.android.components.ActivityRetainedComponent
+import dagger.hilt.android.components.ViewModelComponent
 
-@InstallIn(ActivityRetainedComponent::class)
+@InstallIn(ViewModelComponent::class)
```
manh-t, line 11: 👍

manh-t (commented):
some improvements, the rest looks good

chornerman:
@manh-t Updated 🙏🏼

manh-t (approved):
LGTM 🚀

Issue #98, opened 2021-06-04
Add BaseSharedPreferences


hoangnguyen92dn:
## Why

We are using SharedPreferences on most current projects for caching data. Creating BaseSharedPreferences for common usage:
- BaseSharedPreferences
- BaseEncryptedSharedPreferences

Source:
- https://developer.android.com/topic/security/data

## Who Benefits?

@nimblehq/android-chapter

PR #101, opened 2021-06-18
[98] Add BaseSharedPreferences for RxJavaTemplate

minhnimble:
#98 

## What happened 👀

We are using SharedPreferences on most current projects for caching data. Creating BaseSharedPreferences for common usage.

## Insight 📝

- Add one abstract class and two new classes for Storage management and caching data: `BaseSharedPreferences`, `EncryptedSharedPreferences` and `NormalSharedPreferences`.
- Setup Hilt with the new `StorageModule`.
- Add some utility functions for `BaseSharedPreferences`.
- Update minSdk to `23` since it is required for the crypto lib to work:
Screen Shot 2021-06-17 at 18 14 53

 
## Proof Of Work 📹

- Just under the hood optimization, no visual changes. The app is built and running successfully:

Screen Shot 2021-06-18 at 12 14 04


RxJavaTemplate/app/src/main/java/co/nimblehq/rxjava/storage/BaseSharedPreferencesImpl.kt
```diff
@@ -0,0 +1,39 @@
...
+import android.content.SharedPreferences
+import android.preference.PreferenceManager
+import co.nimblehq.rxjava.extension.execute
+import co.nimblehq.rxjava.extension.put
+import javax.inject.Inject
+
+// TODO: For demo purpose only, replace with actual value that needs to be stored securely
+private const val PREF_TEST_ID = "PREF_TEST_ID"
+
+class BaseSharedPreferencesImpl @Inject constructor(applicationContext: Context) :
```
hoangnguyen92dn, line 13: I think we don't need to create `interface` and `Impl` class. Why don't we create abstract class? 🤔 

One more thing, we should pass the `prefsName` from the constructor.

RxJavaTemplate/app/src/main/java/co/nimblehq/rxjava/storage/BaseSharedPreferencesImpl.kt
```diff
@@ -0,0 +1,39 @@
...
+    init {
+        sharedPreferences = defaultPreference(applicationContext)
+    }
+
+    // TODO: For demo purpose only, replace with actual value that needs to be stored securely
+    override var testId: String?
+        get() = sharedPreferences.getString(PREF_TEST_ID, null)
+        set(value) {
+            sharedPreferences.execute { it.put(PREF_TEST_ID to value) }
+        }
```
hoangnguyen92dn, line 27: If we have a dozen of values so we have to duplicate:

```
        get() = sharedPreferences.getXXX()
        set(value) {
            sharedPreferences.execute { it.put(XXX to value) }
        }
```

Could we have the function with simple `set(KEY_XXX, value)` and `get(KEY_XXX)` for common usage?

RxJavaTemplate/app/src/main/java/co/nimblehq/rxjava/extension/SharePreferencesExt.kt
```diff
@@ -0,0 +1,22 @@
...
+    val key = pair.first
+    when (val value = pair.second) {
+        is String -> putString(key, value)
+        is Int -> putInt(key, value)
+        is Boolean -> putBoolean(key, value)
+        is Long -> putLong(key, value)
+        is Float -> putFloat(key, value)
+        else -> error("Only primitive types can be stored in SharedPreferences")
+    }
+}
```
hoangnguyen92dn, line 22: I'd suggest having the inline function with generic type inside the BaseSharedPreferences like this

```
    protected inline fun <reified T> get(key: String): T? =
        if (prefs.contains(key)) {
            when (T::class) {
                Boolean::class -> prefs.getBoolean(key, false) as T?
                String::class -> prefs.getString(key, null) as T?
                Float::class -> prefs.getFloat(key, 0f) as T?
                Int::class -> prefs.getInt(key, 0) as T?
                Long::class -> prefs.getLong(key, 0L) as T?
                else -> null
            }
        } else {
            null
        }
```

```
    protected fun <T> set(key: String, value: T?) {
        prefs.edit().apply {
            when (value) {
                null -> remove(key)
                is Boolean -> putBoolean(key, value)
                is String -> putString(key, value)
                is Float -> putFloat(key, value)
                is Long -> putLong(key, value)
                is Int -> putInt(key, value)
            }
        }.apply()
    }
```

Then from inherited classes:

```
    private var userId: Int?
        get() = get(PREFS_KEY_USER_ID)
        set(value) = set(PREFS_KEY_USER_ID, value)
```

It would look more simple than having getter/setter methods from SharedPreferences classes

RxJavaTemplate/app/src/main/java/co/nimblehq/rxjava/extension/SharePreferencesExt.kt
```diff
@@ -0,0 +1,10 @@
+package co.nimblehq.rxjava.extension
+
+import android.content.SharedPreferences
+
+fun SharedPreferences.execute(operation: (SharedPreferences.Editor) -> Unit) {
```
hoangnguyen92dn, line 5: Do we still need this? 🤔  Why don't we call `sharedPreferences.clear()` from the Base class? 🤔

hoangnguyen92dn:
@minhnimble You are adding the BaseSharedPreferences in the RxJavaTemplate so I think it could be shared for common usage for both templates 🤔

RxJavaTemplate/app/src/main/java/co/nimblehq/rxjava/extension/SharePreferencesExt.kt
minhnimble: yeah we need it @hoangnguyen92dn, this is an extension function that helps us don't have to rewrite all the repeated calls like apply(), edit(), etc. and just need to write the main operation in the operation block

minhnimble:
> @minhnimble You are adding the BaseSharedPreferences in the RxJavaTemplate, I think it could be shared for common usage for both templates 🤔 
Yeah that would make sense, now I prefer we do it in another PR instead since this PR is for RxJavaTemplate only. I will create another PR for the other template once the final implementation of the SharedPreferences is good. What do you think?

RxJavaTemplate/app/src/main/java/co/nimblehq/rxjava/extension/SharePreferencesExt.kt
hoangnguyen92dn: Ah. I missed it. So it's oke 😄

hoangnguyen92dn (approved):
LGTM

manh-t:
@minhnimble currently, our `app` module does not contain `repository` or `usecase` (where we should use `SharePreference`), so that there is no usage for `SharePreference` here. Therefore, I would recommend moving `SharePreference` to `domain` package, where `repository` and `usecase` are located 😄  How do you think? 🙏

chornerman (approved):

RxJavaTemplate/domain/src/main/java/co/nimblehq/rxjava/domain/storage/NormalSharedPreferences.kt
```diff
@@ -0,0 +1,24 @@
...
+    }
+
+    fun useDefaultSharedPreferences() {
+        sharedPreferences = defaultSharedPreferences
+    }
+
+    // Use this function for creating a custom sharedPreferences if needed
+    fun useCustomSharedPreferences(name: String) {
+        sharedPreferences = applicationContext.getSharedPreferences(name, Context.MODE_PRIVATE)
+    }
```
manh-t, line 23: Could we make these functions private? 🙏

manh-t (commented):
rest LGTM

RxJavaTemplate/domain/src/main/java/co/nimblehq/rxjava/domain/storage/NormalSharedPreferences.kt
minhnimble: @manh-t this is intentional since in the future, we might need to use it for specific usecases. Please ignore the warnings here.  🙏

manh-t (approved):

Issue #111, opened 2021-08-09
Gradle scripts with Kotlin DSL


hoangnguyen92dn:
## Why

Kotlin DSL is now supported to handle Gradle scripts with Kotlin syntax. It's more familiar than Groovy-based syntax for Android developers.

-> Convert some common Groovy-based Gradle scripts into Kotlin DSL scripts.

## Who Benefits?

@nimblehq/android-chapter

PR #116, opened 2021-08-23
[#111] Gradle scripts with Kotlin DSL - Part 1

hoangnguyen92dn:
https://github.com/nimblehq/android-templates/issues/111

## What happened 👀

Kotlin DSL is now supported to handle Gradle scripts with Kotlin syntax. It's more familiar than Groovy-based syntax for Android developers.

-> Convert some common Groovy-based Gradle scripts into Kotlin DSL scripts.
 
## Insight 📝

- Create new `buildSrc` folder to support Kotlin-DSL
- Migrate Gradle scripts to `gradle.kts` supported Kotlin on `RxTemplate` project
- There are still having 2 Gradle scripts (`jacoco.gradle` & `codequality.gradle`) need to be migrated

👉  Part 2 🏃 :
- Migrate `jacoco.gradle` & `codequality.gradle` to `.kts`
- Migrate scripts on `CoroutineTemplate`
 
## Proof Of Work 📹

🚀 CI build should be passed ✅


Tuubz:
@hoangnguyen92dn  Should target branch be develop ? 🙏

Tuubz:
@hoangnguyen92dn Will we also apply the same changes for `CoroutineTemplate`? 🙏

hoangnguyen92dn:
@Tuubz For sure 😄 I mentioned the changes on `CoroutineTemplate` on Part 2 😄

RxJavaTemplate/buildSrc/src/main/java/DependencyHandlerExtensions.kt
```diff
@@ -0,0 +1,50 @@
+import org.gradle.api.artifacts.dsl.DependencyHandler
+
+// Util functions for adding the different type dependencies from build.gradle file
+fun DependencyHandler.api(list: List<String>) {
+    list.forEach { dependency ->
+        add("api", dependency)
```
Tuubz, line 6: Optional: Add `private const val`

RxJavaTemplate/buildSrc/src/main/java/DependencyHandlerExtensions.kt
```diff
@@ -0,0 +1,50 @@
...
+    list.forEach { dependency ->
+        add("androidTestImplementation", dependency)
+    }
+}
+
+fun DependencyHandler.testImplementation(list: List<String>) {
+    list.forEach { dependency ->
+        add("testImplementation", dependency)
+    }
+}
```
Tuubz, line 50: New line 🙏

RxJavaTemplate/buildSrc/src/main/java/Dependencies.kt
```diff
@@ -0,0 +1,139 @@
+class Dependencies {
+
+    object AppModule {
+        val implementations = listOf(
+            // Kotlin stdlib
+            "org.jetbrains.kotlin:kotlin-stdlib-jdk7:${Versions.KOTLIN_VERSION}",
+            "org.jetbrains.kotlin:kotlin-stdlib:${Versions.KOTLIN_VERSION}",
+            // Android components
+            "androidx.core:core-ktx:${Versions.ANDROID_CORE_VERSION}",
```
Tuubz, line 9: Ex: `androidx.core:core-ktx:${Versions.ANDROID_CORE_VERSION}` is added multiple times in this file.

I wonder if it would be better if we define all dependencies in a separate object class first, then we can choose which dependencies to include based on the module 

With this approach we can reduce duplication, but also keep all the modules synced 

What do you think? 🙏

RxJavaTemplate/buildSrc/src/main/java/Versions.kt
```diff
@@ -0,0 +1,56 @@
+object Versions {
+     const val BUILD_GRADLE_VERSION = "4.2.2"
+
+     const val ANDROID_VERSION_CODE = 9
+     const val ANDROID_VERSION_NAME = "3.1.0"
+     const val ANDROID_MIN_SDK_VERSION = 23
+     const val ANDROID_TARGET_SDK_VERSION = 28
```
doannimble, line 7: According to https://developer.android.com/distribute/best-practices/develop/target-sdk, could we change the ANDROID_TARGET_SDK_VERSION version to 30? And the ANDROID_COMPILE_SDK_VERSION as well

RxJavaTemplate/buildSrc/src/main/java/Versions.kt
```diff
@@ -0,0 +1,56 @@
...
+     const val ANDROID_COMPILE_SDK_VERSION = 28
+
+     // Dependencies (Alphabet sorted)
+     const val ANDROID_CORE_VERSION = "1.2.0"
+     const val ANDROID_CRYPTO_VERSION = "1.0.0"
+     const val ANDROID_FRAGMENT_VERSION = "1.3.3"
+     const val ANDROID_LEGACY_SUPPORT_VERSION = "1.0.0"
+     const val ANDROID_LIFECYCLE_VERSION = "2.2.0"
+     const val ANDROID_MATERIAL_VERSION = "1.2.0-alpha06"
+     const val ANDROID_SUPPORT_VERSION = "1.2.0-beta01"
```
doannimble, line 17: Do we really need to use ANDROID_SUPPORT_VERSION ? As we used AndroidX already. Please consider about this, thank you.

RxJavaTemplate/app/build.gradle.kts
```diff
@@ -0,0 +1,141 @@
...
+
+fun loadGradleProperties(projectRootDir: File, file: String): Properties {
+    val properties = Properties()
+    val signingProperties = File(projectRootDir, file)
+
+    if (signingProperties.isFile) {
+        properties.load(signingProperties.inputStream())
+    }
+    return properties
+}
```
Tuubz, line 25: Optional: Move to `FileExt.kt`

RxJavaTemplate/buildSrc/src/main/java/Configurations.kt
```diff
@@ -0,0 +1,18 @@
+class Configurations {
+    object Flavor {
```
Tuubz, line 2: Would it be better if `Flavour`, `BuildType` & `Module` were not a sub class of `Configurations?

Then the gradle files will be easier to read:

Ex: `Congfigurations.Flavor.Production` will be `Flavors.Production` 🙏

RxJavaTemplate/buildSrc/src/main/java/Dependencies.kt
hoangnguyen92dn: @Tuubz Initially, I did create separate dependencies ex: `const val ANDROIDX_CORE = "androidx.core:core-ktx:${Versions.ANDROID_CORE_VERSION}"` for common usage, but I see that quite hard to maintenance the version of dependencies between modules, once we update the version, it might cause the unnecessary update to other modules 🤔 Agree with you about the point that approach can reduce the duplication.

RxJavaTemplate/buildSrc/src/main/java/Configurations.kt
hoangnguyen92dn: @Tuubz Make sense 👍 Three deep level is quite hard for readabilty

RxJavaTemplate/buildSrc/src/main/java/Versions.kt
hoangnguyen92dn: @doannimble 🤔 Sorry what do you mean by `As we used AndroidX already`? The current `ANDROID_SUPPORT_VERSION` is using for dependencies as below

```
            "androidx.appcompat:appcompat:${Versions.ANDROID_SUPPORT_VERSION}",
            "androidx.recyclerview:recyclerview:${Versions.ANDROID_SUPPORT_VERSION}",
```

I think that the `SUPPORT` makes you confused a bit. I would do rename for better meaning here 😄

RxJavaTemplate/buildSrc/src/main/java/Versions.kt
hoangnguyen92dn: @doannimble 👍 About the scope of the update, I'd like to keep all the current versions rather than an update on this PR, it might cause unexpected issues 😄 We would open a new PR to update the version.

RxJavaTemplate/buildSrc/src/main/java/Dependencies.kt
Tuubz: > once we update the version, it might cause the unnecessary update to other modules

Hmm, but that will be the case in the current implementation too, right? 🙏 

If we change `Versions.ANDROID_CORE_VERSION` it will be applied to all modules :still_thinking:

hoangnguyen92dn: @Tuubz I will update to define the dependencies for common usage 😄

RxJavaTemplate/app/build.gradle.kts
```diff
@@ -0,0 +1,129 @@
...
+
+    implementation(Dependencies.AppModule.implementations)
+    debugImplementation(Dependencies.AppModule.debugImplementations)
+    releaseImplementation(Dependencies.AppModule.releaseImplementations)
+
+    // Testing
+    testImplementation(Dependencies.AppModule.testImplementations)
+    androidTestImplementation(Dependencies.AppModule.androidTestImplementations)
+    kaptAndroidTest(Dependencies.AppModule.androidTestKapts)
+}
```
Tuubz, line 129: Should there be a new line? 🙏 

Same goes for other files

RxJavaTemplate/app/build.gradle.kts
```diff
@@ -0,0 +1,129 @@
...
+
+dependencies {
+    implementation(project(Module.COMMON_RX))
+    implementation(project(Module.DOMAIN))
+
+    implementation(fileTree(mapOf("dir" to "libs", "include" to listOf("*.jar"))))
+
+    kapt(Dependencies.AppModule.kapts)
+
+    implementation(Dependencies.AppModule.implementations)
```
Tuubz, line 121: Would it be better if we could reduce it to `AppModule.implementations` ? 🙏 

But keep the class named as `Dependencies.kt`

RxJavaTemplate/buildSrc/src/main/java/FileExt.kt
```diff
@@ -0,0 +1,12 @@
+import java.io.File
+import java.util.*
+
+fun loadGradleProperties(projectRootDir: File, fileName: String): Properties {
```
Tuubz, line 4: Optional: `fun File.loadGradleProperties(fileName: String): Properties`

Tuubz (approved):
Cool! 😍

I was going to ask if Kotlin was faster than Groovy, but I already found the answer 🙏 

> In the future, KTS will be preferred over Groovy for writing Gradle scripts because Kotlin is more readable and offers better compile-time checking and IDE support. Although KTS currently offers better integration in Android Studio’s code editor when compared to Groovy, builds using KTS tend to be slower than builds using Groovy, so migrating to KTS should be considered with build performance in mind.

RxJavaTemplate/buildSrc/src/main/java/Dependencies.kt
```diff
@@ -0,0 +1,136 @@
...
+    )
+
+    val testImplementations = listOf(
+        JUNIT,
+        // kluent
+        KLUENT_ANDROID,
+        // Mockito
+        MOCKITO_KOTLIN,
+        // Robolectric
+        ROBOLECTRIC
```
lydiasama, line 76: I wonder do we really need these comments? I feel like the constant name is explicit itself better. 🤔  (The same idea as at other variables as well.)

RxJavaTemplate/settings.gradle.kts
```diff
@@ -0,0 +1,2 @@
+// FIXME Project build error when using Configurations.Module constants
+include(":app", ":common-rx", ":data", ":domain")
```
lydiasama, line 2: EOF issue. 🙋🏻‍♀️

RxJavaTemplate/buildSrc/src/main/java/Dependencies.kt
hoangnguyen92dn: @lydiasama I was supposed to add the comment to group the dependencies with the same package. 🤔  I will remove the comment if there is only one dependency from the package. What do you think?

hoangnguyen92dn: @lydiasama I updated to remove all the unnecessary comments 😄

RxJavaTemplate/buildSrc/src/main/java/Versions.kt
doannimble: Okay, if it is our current development process, then we will need the new PR later.

RxJavaTemplate/buildSrc/src/main/java/Versions.kt
doannimble: Yes please change to ANDROIDX_VERSION or something like that to avoid confusing.

RxJavaTemplate/build.gradle.kts
```diff
@@ -0,0 +1,59 @@
...
+        xml {
+            enabled = true
+            destination = file("build/reports/detekt.xml")
+        }
+        html {
+            enabled = true
+            destination = file("build/reports/detekt.html")
+        }
+    }
+}
```
lydiasama, line 59: Ah I just see, EOF issue again. 🙏🏻 
Or you can let AS manages it for you. 😉 
Screen Shot 2564-08-31 at 11 29 20

hoangnguyen92dn: @lydiasama I disabled the settings on AS because of the new project doesn’t use that style 😄  Updated btw 🚀

lydiasama: @hoangnguyen92dn I see.

lydiasama (approved):

doannimble (approved):

luongvo:
@hoangnguyen92dn putting some words such as Solves, Resolves will automatically close the issue after merging the PR I guess. Try it 🤓 
image

RxJavaTemplate/buildSrc/src/main/java/DependencyHandlerExt.kt
```diff
@@ -0,0 +1,59 @@
...
+private const val DEPENDENCY_API = "api"
+private const val DEPENDENCY_KAPT = "kapt"
+private const val DEPENDENCY_KAPT_ANDROID_TEST = "kaptAndroidTest"
+private const val DEPENDENCY_IMPLEMENTATION = "implementation"
+private const val DEPENDENCY_IMPLEMENTATION_DEBUG = "debugImplementation"
+private const val DEPENDENCY_IMPLEMENTATION_RELEASE = "releaseImplementation"
+private const val DEPENDENCY_IMPLEMENTATION_TEST = "testImplementation"
+private const val DEPENDENCY_IMPLEMENTATION_ANDROID_TEST = "androidTestImplementation"
+
+// Util functions for adding the different type dependencies from build.gradle file
```
minhnimble, line 12: small wording improvement 
```suggestion
// Util functions for adding the different type of dependencies from build.gradle file
```

minhnimble (commented):
The rest LGTM 🚀

hoangnguyen92dn:
@luongvo I think it depends on the Project board settings. If we link issue with PR then issue will be closed automatically :thinking:

lydiasama:
Screen Shot 2564-08-31 at 12 01 37
@luongvo The issue will close automatically once we merge the PR that linked to it.

RxJavaTemplate/buildSrc/src/main/java/Dependencies.kt
```diff
@@ -0,0 +1,127 @@
...
+        ANDROIDX_RECYCLERVIEW,
+        ANDROIDX_CONSTRAINT_LAYOUT,
+        ANDROIDX_MULTIDEX,
+
+        ANDROIDX_NAVIGATION_RUNTIME_KTX,
+        ANDROIDX_NAVIGATION_FRAGMENT_KTX,
+        ANDROIDX_NAVIGATION_UI_KTX,
+
+        RX_BINDING,
+        RX_BINDING_CORE,
```
AVI5HEK, line 20: How about putting these two below `ANDROIDX_LIFECYCLE_RUNTIME,`? Keeping the `ANDROIDX` dependencies close to each other would make it prettier. 😃

RxJavaTemplate/app/build.gradle.kts
```diff
@@ -0,0 +1,129 @@
...
+    id("com.android.application")
+    id("kotlin-android")
+    id("kotlin-kapt")
+    id("kotlin-parcelize")
+
+    id("dagger.hilt.android.plugin")
+    id("androidx.navigation.safeargs.kotlin")
+}
+
+// TODO Migrate to gradle.kts
```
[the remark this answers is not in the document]
hoangnguyen92dn: @AVI5HEK The files below are still in `.gradle` extension. I added the TODO here to mention the migrations later on.

AVI5HEK (approved):
LGTM. 🚀

RxJavaTemplate/buildSrc/src/main/java/Dependencies.kt
```diff
@@ -0,0 +1,127 @@
...
+object CommonRxModule {
+    val implementations = listOf(
+        KOTLIN_STDLIB_JDK7
+    )
+
+    val apis = listOf(
+        RX_JAVA,
+        RX_KOTLIN
+    )
+}
```
luongvo, line 127: There are still some concerns from me to manage project dependencies by this approach. Because this change affects too much on the base structure of our template and the way to use it in our development process so I recommend having an open discussion before continuing 🙏

sleepylee: I agree

RxJavaTemplate/build.gradle.kts
```diff
@@ -0,0 +1,59 @@
...
+    repositories {
+        google()
+        jcenter()
+    }
+
+    dependencies {
+        classpath("com.android.tools.build:gradle:${Versions.BUILD_GRADLE_VERSION}")
+        classpath("com.google.dagger:hilt-android-gradle-plugin:${Versions.HILT_VERSION}")
+        classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:${Versions.KOTLIN_VERSION}")
+        classpath("androidx.navigation:navigation-safe-args-gradle-plugin:${Versions.NAVIGATION_VERSION}")
```
sleepylee, line 13: can't these paths be managed in the same manner as you did for the Dependencies declaration? 
`"androidx.navigation:navigation-safe-args-gradle-plugin:` -> `ANDROIDX_SAFE_ARGS_PLUGIN`? 🤔

RxJavaTemplate/app/build.gradle.kts
```diff
@@ -0,0 +1,130 @@
...
+    buildFeatures {
+        viewBinding = true
+    }
+}
+
+dependencies {
+    implementation(project(Module.COMMON_RX))
+    implementation(project(Module.DOMAIN))
+
+    implementation(fileTree(mapOf("dir" to "libs", "include" to listOf("*.jar"))))
```
sleepylee, line 118: I think this can be removed now since we are not having any jar dependency 🤔

luongvo: @sleepylee I think we can keep it here for `sample`? Kotlin is still new for someone, it's easier in case we need to add a lib in client project? 🤔

minhnimble:
Need to fix conflicts 🛠🦀

RxJavaTemplate/build.gradle.kts
hoangnguyen92dn: @sleepylee The configuration for classpath will be updated in the next upcoming PR. On this PR's scope, I am doing migration rather than refactoring the way we organize dependencies 🙏

RxJavaTemplate/build.gradle.kts
```diff
@@ -0,0 +1,59 @@
+// Top-level build file where you can add configuration options common to all sub-projects/modules.
+
+buildscript {
+    repositories {
+        google()
+        jcenter()
```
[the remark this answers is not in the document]
hoangnguyen92dn: @luongvo Yup. We should replace by `mavenCentral()` instead. But for now, it requires more steps to config `detekt` and `jacoco` without jcenter(). So I will leave it as it is for now.

hoangnguyen92dn:
@luongvo Please finalize your review 🚀

sleepylee (approved):

luongvo (approved):
tested, good to go now 🏁

Issue #117, opened 2021-08-24
Add Proxyman configurations to intercept requests/responses


Tuubz:
## Why

[Proxyman](https://docs.proxyman.io/) is a tool that acts as proxy, to intercept requests and responses between the client and server. By doing this we can gain control over the traffic that's being sent and received in our app. 

We need to add a network security config file to whitelist the domains that we will be intercepting (only for staging env)

## Who Benefits?

@nimblehq/android-chapter can view, inspect and manipulate the content of requests and response for testing 🚀

lydiasama:
@Tuubz We need to disable the pinning certification as well, Does it include inside of this content?

Tuubz:
@lydiasama Could you please refresh my memory why we need to disable the pinning certificate? 🙇 

Anyhow, I don't think it relates directly to this ticket 🙏

lydiasama:
@Tuubz As I use the proxyman with DeeMoney project. only adding the network security config is not enough. We need to disable the pinning certificate as well. Have you tried the proxyman with the pinning certificated projects? If it works, I maybe need your help. 🙏🏻

Tuubz:
@lydiasama Aah, I see

No, it wasn't required for the 7-Eleven project. 

Also, for this project it's not required too. So I'm not sure if we should add it here

What did you do to solve it on DeeMoney project? 🙏

lydiasama:
@Tuubz I just create my local branch to add those configurations. I'm thinking to bring it into the new buildType `mock` as well for convenience.

Tuubz:
Continue discussion in [this thread](https://nimble-co.slack.com/archives/G41UCTGHL/p1629781993001700)

PR #118, opened 2021-08-25
[#117] Add network security config to support Proxyman

Tuubz:
https://github.com/nimblehq/android-templates/issues/117

## What happened 👀

To use Proxyman, we need to add network security config files to whitelist the domain of our API

I added this for both `RxJavaTemplate` & `CoroutineTemplate`

Read more: [slides](https://docs.google.com/presentation/d/17QIg1Ru5w2m2MkoSb-NTrkl7RVkhdC_DoMIZsGommk8/edit?usp=sharing)
 
## Insight 📝

- Add `network_security_config.xml` for main
- Add `network_security_config.xml` for staging

This is because we should not allow any proxy configurations in production environment
  
## Proof Of Work 📹

https://user-images.githubusercontent.com/18277915/130769062-5352efda-bedc-4297-b739-280e87d899da.mov


CoroutineTemplate/app/src/staging/res/xml/network_security_config.xml
```diff
@@ -0,0 +1,22 @@
...
+    </debug-overrides>
+
+    <base-config cleartextTrafficPermitted="true">
+        <trust-anchors>
+            <certificates src="system" />
+        </trust-anchors>
+    </base-config>
+
+    <domain-config>
+        <domain includeSubdomains="true">www.jsonplaceholder.typicode.com</domain>
```
hoangnguyen92dn, line 16: 🤔  So when we update the base URL then we have to update the domain as well. For now, I think we could add the comment to notice that replacing `includeSubdomains` to match the base URL. What do you think?

CoroutineTemplate/app/src/main/res/xml/network_security_config.xml
```diff
@@ -0,0 +1 @@
+<network-security-config />
```
sleepylee, line 1: please add eof line break

CoroutineTemplate/app/src/staging/res/xml/network_security_config.xml
```diff
@@ -0,0 +1,22 @@
+<network-security-config>
+    <debug-overrides>
+        <trust-anchors>
+            <certificates src="user" />
+            <certificates src="system" />
+        </trust-anchors>
+    </debug-overrides>
+
+    <base-config cleartextTrafficPermitted="true">
```
sleepylee, line 9: I think we don't need this tho? 🤔  or do we need to grant the permission to transmit via `http` somewhere?

Tuubz: Aah, I see!

We can remove this indeed, if we don't want to support `http` 🙏

Tuubz:
Updated 🚀

sleepylee (approved):

lydiasama (approved):

AVI5HEK (approved):

luongvo:
LGTM ✅

Issue #127, opened 2021-09-06
Add Danger to the Bitise workflow


minhnimble:
## Why

While each platform has specific requirements for their Continuous Integration pipelines, automated linters and code checks (security, code smells) must be made standard across all platforms. Android is lagging behind in this regard. Simultaneously, the mobile chapters should adopt the same tooling to limit the number of different tools we need to manage at the team level.

The Web Chapter already uses [Danger](https://danger.systems/) to automate code review comments from linters and code checks. Since it is a proven tool even written in [Swift](https://danger.systems/swift/) and [Kotlin](https://github.com/danger/kotlin) (recently released) projects, the Android chapter should adopt it as well.

## Who Benefits?

All engineering chapters such as @nimblehq/android-chapter, @nimblehq/ios-chapter and @nimblehq/web-chapter since the goal is to use a unified tool for all platforms.

PR #130, opened 2021-09-13
[#127] [WIP] Add newworkflow for automatic code review using Danger with Bitrise

minhnimble:
https://github.com/nimblehq/android-templates/issues/127

## What happened 👀

While each platform has specific requirements for their Continuous Integration pipelines, automated linters and code checks (security, code smells) must be made standard across all platforms. Android is lagging behind in this regard. Simultaneously, the Android chapter should adopt the same tooling to limit the number of different tools we need to manage at the team level.

The Web Chapter already uses [Danger](https://danger.systems/) to automate code review comments from linters and code checks. Since it is a proven tool even written in [Kotlin](https://github.com/danger/kotlin) (recently released) project, the Android chapter should adopt it as well.

## Insight 📝

- To setup Bitrise with Github Action, we need to use danger-ruby because that lib is the only support for Bitrise.
- Here are the steps I did for integrating Danger to our Android Template:
   - Create a new Bitrise workflow - `automatic_code_review` to `bitrise.yml` with Danger step for performing automatically PR review.
   - Add the `Dangerfile` file to our root project that can help Danger look through the PR metadata and check with some predefined rules provided there: `Changelog update` check, `Big PR` check and `Work in progress` check.
   - The instructions for setting up the Danger step can be found [here](https://github.com/danger/kotlin#github-actions) and the latest `Dangerfile.df.kts` file can be found [here](https://github.com/danger/kotlin#what-it-looks-like-today).
- I also added support for the following danger-ruby plugins:  
   - TBD
   - TBD

## Proof Of Work 📹

TBD


RxJavaTemplate/domain/build.gradle
```diff
@@ -44,6 +44,12 @@ android {
         sourceCompatibility JavaVersion.VERSION_1_8
         targetCompatibility JavaVersion.VERSION_1_8
     }
+
+    android {
+        lintOptions {
+            xmlReport true
```
sleepylee, line 50: hmm... why the previous config for github action did not have this lint result attached as well? 🤔

Dangerfile
```diff
@@ -0,0 +1,30 @@
...
+
+# Check commits lint and warn on all checks (instead of failing)
+commit_lint.check warn: :all
+
+# Detekt output check
+kotlin_detekt.detekt
+kotlin_detekt.gradle_task = "detekt"
+
+# Android Lint output check
+android_lint.lint
```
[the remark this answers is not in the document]
minhnimble: actually we can, back then I wanted to try the danger-kotlin one 😅

RxJavaTemplate/domain/build.gradle
minhnimble: Ah because I remove the usage of the [danger-kotlin-android-lint-plugin](https://github.com/gianluz/danger-kotlin-android-lint-plugin) as it was not working properly and thus this lint result attachment is not needed. 

Now since we all agreed to use danger-ruby for both CIs, let me update to attach this lint result for the Github Action 🙏

.cicdtemplate/.bitrise/bitrise.yml
```diff
@@ -1,224 +1,283 @@
...
+- push_branch: master
+  workflow: master
+  
+- push_branch: "*"
+  workflow: automatic_code_review
 
-  - pull_request_source_branch: '*'
-    workflow: primary
+- pull_request_source_branch: "*"
+  workflow: automatic_code_review
```
luongvo, line 20: `automatic_pull_request_review` 💭

.cicdtemplate/.bitrise/bitrise.yml
```diff
@@ -1,224 +1,283 @@
...
+- push_branch: release/*
+  workflow: release
 
-  - push_branch: '*'
-    workflow: primary
+- push_branch: master
+  workflow: master
+  
+- push_branch: "*"
+  workflow: automatic_code_review
```
luongvo, line 17: which kind of editor config are you using to edit and format this file? I'm using the default VSCode and it seems the old format looks good. I tried to re-format the whole file and everything back to the previous format. Could you please re-format again then we could focus on the main content of Danger config?

image

.cicdtemplate/.bitrise/bitrise.yml
```diff
@@ -1,224 +1,283 @@
...
+    # Upload jacocoTestReport to Apps & Artifacts for later downloading
+    - deploy-to-bitrise-io@1:
+        inputs:
+        - deploy_path: "$PROJECT_LOCATION/$MODULE/build/reports/jacoco/jacocoTestReport/html/index.html"
+        - is_enable_public_page: 'false'
+        - is_compress: 'true'
+        - notify_user_groups: none
+  
+  # Workflow on release/* branch
+  release:
```
luongvo, line 220: any reason why do we need to re-arrange the order of workflows here? I think the previous order is good to keep: primary, develop, release, master.

.cicdtemplate/.bitrise/bitrise.yml
```diff
@@ -1,224 +1,283 @@
...
-    workflow: primary
+- pull_request_source_branch: "*"
+  workflow: automatic_code_review
 
 # Work-flows definition
 workflows:
+
   # Workflow on * branch
-  primary:
+  automatic_code_review:
```
luongvo, line 26: I think we don't need to create a new workflow here since there is 1 more step added to run Danger. Could we apply Danger on primary workflow instead?

luongvo (commented):
almost issues about formatting, workflows ordering... 😄

.cicdtemplate/.bitrise/bitrise.yml
minhnimble: @luongvo nice question, I used the Bitrise's website editor view and this is how they are indented now, I suggest we keep it to Bitrise style since whenever we try to make a change to the Bitrise's workflow using the website GUI, the indentation will be automatically updated to my current indentation style 🙏

.cicdtemplate/.bitrise/bitrise.yml
minhnimble: oh it is generated by Bitrise when applying and modifying on the web editor view, please just ignore 🙏

.cicdtemplate/.bitrise/bitrise.yml
minhnimble: it is not a new workflow, we can not create a new workflow with same triggers using Bitrise, I just rename it from `primary` to `automatic_code_review` for correctly reflecting the current workflow's functionalities 😅

.cicdtemplate/.bitrise/bitrise.yml
luongvo: @minhnimble nope, I don't think so, we need to manage it as repository source code and everything should be formatted in the same way.

.cicdtemplate/.bitrise/bitrise.yml
luongvo: Same to what I explained above, we should manage it here, try to not follow all the results from the Editor please 🙏 
Reformatting or reordering should be done in a separate task or PR I think.

.cicdtemplate/.bitrise/bitrise.yml
luongvo: @minhnimble but the primary workflow is still [here](https://github.com/nimblehq/android-templates/pull/130/files#diff-9a67f94064187042e4635e3345dee64427399a151989b21b0eeb151616bacbc1R174). Imo, better to replace the primary with the new workflow or keep both of them (as you said it's impossible).

minhnimble: @luongvo yeah I that is why I added the comment - `# Workflow not used for now` there as well to keep the `primary` workflow code in case we still need it later on 😅

luongvo: @minhnimble imo yes, let's ask the author too @hoangnguyen92dn ✋

Issue #131, opened 2021-09-24
Fix: Clean Architecture rules in Coroutine template 


doannimble:
## Issue

Our current implementation of Clean Architecture rules in Coroutine template is not correct, details : 
- Domain module : depends on Data module and Android framework
- Data module : no repository implementation, no mapper from Data to Domain model

## Expected

- Domain should contain only Kotlin code which has no dependency
- Data module should have Repository implementation, mapper from Data to Domain model

## Steps to reproduce

N/A

lydiasama:
@doannimble I think it's not a bug. 🤔  Should it be a chore ticket for improvement?

doannimble:
@lydiasama I don't think so. Which does not correct should be a bug. This is a correction, not an improvement.
Just a little bit improvement of Detekt included by using the latest version.

sleepylee:
@doannimble @lydiasama: just answer a simple question: did the app work as expected before? If yes - it's a bug, otherwise it's not.
Do these changes alter the app behavior or just a refactor work at the codebase level? - if not, it's a refactor; if it changes the app behavior i.e performing better: enhancement

luongvo:
@doannimble I'm quite sure this is just an improvement task, not a bug.

doannimble:
@sleepylee @lydiasama 
- Did the app work as expected before ? : Yes
- Do these changes alter the app behavior or just a refactor work at the codebase level? : Refactor work at codebase level.
So we only catching bug on app level? If that is true, then I agree it a refactor or improvement tasks.

@luongvo I agree, it should be refactor or improvement.

sleepylee:
@doannimble yes, and it could happens too; sometimes if there is an issue with our subscription model (missing context) it could crash the app when performing an action on the UI thread; or duplicated emission when we forgot to unsubscribe sth; so in that cases, we need a space to log the bug

doannimble:
@sleepylee I got it. Thank you 🙏

lydiasama:
@sleepylee Should we create content about how to categorize these as IBM did? 😅

> @doannimble @lydiasama: just answer a simple question: did the app work as expected before? If no - it's a bug, otherwise it's not.
> Do these changes alter the app behavior or just a refactor work at the codebase level? - if not, it's a refactor; if it changes the app behavior i.e performing better: enhancement.
> 
> Here I'm taking a reference definition from IBM:
> 
> https://www.ibm.com/topics/bug-tracking
[...]

PR #132, opened 2021-09-26
[#131] Fix: Clean Architecture rules in Coroutine template

doannimble:
Close https://github.com/nimblehq/android-templates/issues/131

## What happened 👀

- Update : Domain should contain only Kotlin code which has no dependency
- Update : Data module should have Repository implementation, mapper from Data to Domain model
- Add : Dispatcher to ViewModel
- Update Detekt to the latest version (1.18.1) which is supported by MavenCentral

## Insight 📝

- The should be build and run without error as previous
- Take the rule of Clean Architecture
- Detekt run successfully 

## Proof Of Work 📹

Screen Shot 2021-09-26 at 11 07 34 PM

Screen Shot 2021-09-26 at 11 07 34 PM


CoroutineTemplate/app/src/main/java/co/nimblehq/coroutine/ui/screens/home/HomeViewModel.kt
```diff
@@ -49,10 +50,10 @@ class HomeViewModel @Inject constructor(
 
     private fun fetchUsers() {
         showLoading()
-        viewModelScope.launch {
-            when (val result = getUsersUseCase.execute()) {
-                is UseCaseResult.Success -> _users.value = result.data
-                is UseCaseResult.Error -> _error.emit(result.exception.message.orEmpty())
+        execute {
+            when (val result = getUsersUseCase.execute(page = 1, size = 10)) {
+                is co.nimblehq.domain.usecase.UseCaseResult.Success -> _users.value = result.data
```
Wadeewee, line 55: Let's import `import co.nimblehq.domain.usecase.UseCaseResult` so we can use just `UseCaseResult.Success` 🙏

CoroutineTemplate/app/src/main/java/co/nimblehq/coroutine/ui/base/BaseViewModel.kt
```diff
@@ -43,4 +47,8 @@ abstract class BaseViewModel : ViewModel() {
             _showLoading.value = false
         }
     }
+
+    fun execute(job: suspend () -> Unit) = viewModelScope.launch {
+        withContext(dispatchers.getIO()) { job.invoke() }
```
Wadeewee, line 52: I have a question please🙋‍♀️  ,Is it possible on some case we don't want to use `io` thread 🙏

CoroutineTemplate/domain/src/main/java/co/nimblehq/domain/usecase/GetUsersUseCase.kt
```diff
@@ -0,0 +1,13 @@
+package co.nimblehq.domain.usecase
+
+import co.nimblehq.domain.entity.UserEntity
+import co.nimblehq.domain.repository.UserRepository
+import javax.inject.Inject
+
+
```
Wadeewee, line 7: Redundant line 🙏

CoroutineTemplate/domain/src/main/java/co/nimblehq/domain/usecase/GetUsersUseCase.kt
```diff
@@ -0,0 +1,13 @@
...
+class GetUsersUseCase @Inject constructor(private val userRepository: UserRepository) {
+    suspend fun execute(page: Int, size: Int): UseCaseResult<List<UserEntity>> {
+        return userRepository.getUsers(page, size)
+    }
+
```
Wadeewee, line 12: Same here

CoroutineTemplate/build.gradle
```diff
@@ -19,19 +19,20 @@ buildscript {
...
-        kotlin_version = '1.5.10'
+        kotlin_version = '1.5.21'
         kotlinx_coroutines_version = '1.5.0'
 
         moshi_version = '1.12.0'
 
         okhttp_version = '4.9.1'
         retrofit_version = '2.9.0'
 
+        mockito_version = '3.12.4'
```
Wadeewee, line 35: I think we can remove `Mockito` for now because we didn't implement  the `Test` yet, Also  should we consider to use `Mockk` instead `Mockito` or What do you think? 🙏

Wadeewee:
I think for this PR should it be type `Chore` instead `bug` 🤔 because It's a part of improvement 🙇‍♀️

CoroutineTemplate/app/src/main/java/co/nimblehq/coroutine/ui/base/BaseViewModel.kt
doannimble: @Wadeewee In that case, you can create another `function`, like `executeOnMain` and get the corresponding `Dispatcher` from `dispatchers ` or provide the type of `Dispatcher` you want to do as the parameter of the `execute` function.
What do you think?

CoroutineTemplate/app/src/main/java/co/nimblehq/coroutine/ui/screens/home/HomeViewModel.kt
doannimble: @Wadeewee Sure, I've just copied and pasted from previous version without changing.

CoroutineTemplate/build.gradle
doannimble: @Wadeewee It makes sense to remove the `Test` dependencies. May I know why we should consider to use `Mockk` instead of `Mockto` ?

sleepylee: Mockito was evolved from Java; while Mockk is pure Kotlin; there are a few things you can do with Mockk i.e mocking the singleton 😄  we had a presentation somewhere I think @Wadeewee can point you to her/Toby's presentation back then.

Tuubz: @doannimble You find the presentation [here](https://docs.google.com/presentation/d/1pSrDeMFRsSXq9gS58o7y3G5dnAKCUJlQn7EPWlaqQgo/edit) 🙏

doannimble: @Tuubz Tu
Thank you so much, I have a task to create basic unit test. I will use this for my reference 🥇

CoroutineTemplate/app/src/main/java/co/nimblehq/coroutine/ui/base/BaseViewModel.kt
Wadeewee: @doannimble Providing a type via parameter is good to me 😃   but why we didn't inject the `dispatchers` in `UseCase` instead inject in `VM` so you can define the thread whatever you want in `UseCase` Please correct me If I missed something 🙇‍♀️

CoroutineTemplate/app/src/main/java/co/nimblehq/coroutine/ui/screens/compose/ComposeViewModel.kt
```diff
@@ -3,10 +3,10 @@ package co.nimblehq.coroutine.ui.screens.compose
 import androidx.compose.runtime.State
 import androidx.compose.runtime.mutableStateOf
 import androidx.lifecycle.viewModelScope
```
Wadeewee, line 5: Please remove `androidx.lifecycle.viewModelScope` also for `kotlinx.coroutines.launch` 🙏

CoroutineTemplate/app/src/main/java/co/nimblehq/coroutine/ui/base/BaseViewModel.kt
doannimble: @Wadeewee It such a good question. As I remember, it is related to the cancelation of suspend thread. But not exactly,
let me figure out the correct answer then explain to you soon 💯

CoroutineTemplate/domain/src/main/java/co/nimblehq/domain/usecase/UseCaseResult.kt
```diff
@@ -1,4 +1,4 @@
-package co.nimblehq.coroutine.domain.usecase
```
hoangnguyen92dn, line 1: FYI: If we update the package name here, the script to init project will not work 😄 

https://github.com/nimblehq/android-templates/blob/c301ab05aca6249cbe8e62351a5ef35e0e1b7aad/newproject.sh#L71-L82

doannimble: @hoangnguyen92dn Good catch, I've updated the package name.

CoroutineTemplate/build.gradle
```diff
@@ -19,28 +19,26 @@ buildscript {
...
-        test_espresso_version = '3.1.0'
-        test_junit_version = '4.13.2'
         timber_log_version = '4.7.1'
     }
 
     repositories {
         google()
-        jcenter()
+        mavenCentral()
+
```
lydiasama, line 41: A redundant line here.

CoroutineTemplate/data/src/main/java/co/nimblehq/coroutine/data/service/response/UserResponse.kt
```diff
@@ -24,3 +25,36 @@ data class UserResponse(
...
+private fun UserResponse.toEntity(): UserEntity {
+    return UserEntity(
+        id = this.id,
+        name = this.name.orEmpty(),
+        username = this.username.orEmpty(),
+        email = this.email.orEmpty(),
+        address = this.address?.toEntity(),
+        phone = this.phone.orEmpty(),
+        website = this.website.orEmpty()
+    )
```
lydiasama, line 42:
```suggestion
    return with(this) {
        UserEntity(
            id = id,
            name = name.orEmpty(),
            username = username.orEmpty(),
            email = email.orEmpty(),
            address = address?.toEntity(),
            phone = phone.orEmpty(),
            website = website.orEmpty()
        )
    }
```
Can we use the scope function? What do you think?

CoroutineTemplate/data/src/main/java/co/nimblehq/coroutine/data/service/response/UserResponse.kt
```diff
@@ -24,3 +25,36 @@ data class UserResponse(
...
+}
+
+private fun UserResponse.Address.toEntity(): UserEntity.Address {
+    return UserEntity.Address(
+        street = this.street.orEmpty(),
+        suite = this.suite.orEmpty(),
+        city = this.city.orEmpty(),
+        zipCode = this.zipCode.orEmpty(),
+        geo = this.geo?.toEntity()
+    )
```
lydiasama, line 52: Same here.

CoroutineTemplate/domain/src/main/java/co/nimblehq/coroutine/entity/UserEntity.kt
```diff
@@ -0,0 +1,26 @@
...
+data class UserEntity(
+    val id: Int?,
+    val name: String,
+    val username: String,
+    val email: String?,
+    val address: Address?,
+    val phone: String,
+    val website: String
+) {
+
```
lydiasama, line 12: Can we get rid of this line break in case we don't have the `@Parcelize` annotation to be consistent with the `UserResponse`?

CoroutineTemplate/domain/src/main/java/co/nimblehq/coroutine/entity/UserEntity.kt
```diff
@@ -0,0 +1,26 @@
...
+    data class Address(
+        val street: String,
+        val suite: String,
+        val city: String,
+        val zipCode: String,
+        val geo: Geo?
+    ) {
+
```
lydiasama, line 20: Same here: Can we get rid of this line break in case we don't have the `@Parcelize` annotation to be consistent with the `UserResponse`?

CoroutineTemplate/domain/src/main/java/co/nimblehq/coroutine/usecase/GetUsersUseCase.kt
```diff
@@ -0,0 +1,11 @@
+package co.nimblehq.domain.usecase
+
+import co.nimblehq.domain.entity.UserEntity
+import co.nimblehq.domain.repository.UserRepository
+import javax.inject.Inject
+
+class GetUsersUseCase @Inject constructor(private val userRepository: UserRepository) {
+    suspend fun execute(page: Int, size: Int): UseCaseResult<List<UserEntity>> {
```
lydiasama, line 8: Can we have a line-break before this line?

CoroutineTemplate/app/src/main/java/co/nimblehq/coroutine/dispatcher/DispatchersProvider.kt
```diff
@@ -0,0 +1,9 @@
+package co.nimblehq.coroutine.dispatcher
+
+import kotlinx.coroutines.CoroutineDispatcher
+
+interface DispatchersProvider {
+    fun getIO(): CoroutineDispatcher
+    fun getMain(): CoroutineDispatcher
+    fun getDefault(): CoroutineDispatcher
+}
```
lydiasama, line 9:
```suggestion
interface DispatchersProvider {

    fun getIO(): CoroutineDispatcher
    
    fun getMain(): CoroutineDispatcher
    
    fun getDefault(): CoroutineDispatcher
}
```

For consistency. 😁  (You can have a look at other interfaces of the project.)

CoroutineTemplate/data/src/main/java/co/nimblehq/coroutine/data/service/response/UserResponse.kt
doannimble: @lydiasama I think it is a redundant step, just make our code more complicated without having more benefit. 🤔

CoroutineTemplate/data/src/main/java/co/nimblehq/coroutine/data/service/response/UserResponse.kt
doannimble: 🙏

CoroutineTemplate/domain/src/main/java/co/nimblehq/coroutine/entity/UserEntity.kt
doannimble: @lydiasama Make sense. I've updated them for consistensy.

CoroutineTemplate/app/src/main/java/co/nimblehq/coroutine/dispatcher/DispatchersProvider.kt
doannimble: @lydiasama I agree with this.

CoroutineTemplate/domain/src/main/java/co/nimblehq/coroutine/usecase/GetUsersUseCase.kt
doannimble: @lydiasama 👌

CoroutineTemplate/data/src/main/java/co/nimblehq/coroutine/data/service/response/UserResponse.kt
lydiasama: Hmm. true 🤔  Actually we can get rid of `this` because it can directly call the properties inside of the parent class also. But for the extension function, I think to keep the `this` is better to ensure it will use the correct `this` always. 

I just feel it has multiple `this` and it's quite long on the right side. It's just an optional. I'm ok. 👌🏻

CoroutineTemplate/app/src/main/java/co/nimblehq/coroutine/dispatcher/DispatchersProviderImpl.kt
```diff
@@ -0,0 +1,14 @@
...
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.MainCoroutineDispatcher
+
+class DispatchersProviderImpl : DispatchersProvider {
+
+    override fun getMain(): MainCoroutineDispatcher = Dispatchers.Main
+
+    override fun getIO(): CoroutineDispatcher = Dispatchers.IO
+
+    override fun getDefault(): CoroutineDispatcher = Dispatchers.Default
```
sleepylee, line 13: the order is a bit different from the `DispatchersProvider` interface: IO > Main > Default vs Main > IO > Default;

CoroutineTemplate/data/src/main/java/co/nimblehq/coroutine/repositoryimpl/UserRepository.kt
```diff
@@ -0,0 +1,21 @@
...
+import co.nimblehq.coroutine.data.service.response.toUserEntities
+import co.nimblehq.coroutine.entity.UserEntity
+import co.nimblehq.coroutine.repository.UserRepository
+import co.nimblehq.coroutine.usecase.UseCaseResult
+
+class UserRepositoryImpl constructor(
+    private val apiService: ApiService,
+) : UserRepository {
+
+    override suspend fun getUsers(page: Int, size: Int): UseCaseResult<List<UserEntity>> {
```
sleepylee, line 13: why would we need to pass the pagination to get a User profile? 🤔  especially that it's not being used in the later call:

`val response = apiService.getUsers()`

CoroutineTemplate/domain/build.gradle
```diff
@@ -1,41 +1,19 @@
...
-        "androidx.core:core-ktx:$androidx_core_ktx_version",
-        "androidx.security:security-crypto:$android_crypto_version",
-        "com.google.dagger:hilt-android:$hilt_version",
-        "org.jetbrains.kotlinx:kotlinx-coroutines-android:$kotlinx_coroutines_version"
+    api(
+        "com.squareup.moshi:moshi-kotlin:$moshi_version",
+        "com.squareup.retrofit2:converter-moshi:$retrofit_version",
+        "com.squareup.moshi:moshi:$moshi_version",
+        "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version",
+        "javax.inject:javax.inject:1"
```
sleepylee, line 17: `inject:1`?

CoroutineTemplate/domain/src/main/java/co/nimblehq/coroutine/domain/usecase/GetUsersUseCase.kt
```diff
@@ -1,24 +0,0 @@
...
-import co.nimblehq.coroutine.domain.repository.UserRepository
-import kotlinx.coroutines.CoroutineDispatcher
-import kotlinx.coroutines.withContext
-import javax.inject.Inject
-
-class GetUsersUseCase @Inject constructor(
-    private val userRepository: UserRepository,
-    @IoDispatcher private val ioDispatcher: CoroutineDispatcher
-) {
-    suspend fun execute(): UseCaseResult<List<UserEntity>> {
```
sleepylee, line 14: originally, it doesn't have any pagination argument 🤔

CoroutineTemplate/domain/src/main/java/co/nimblehq/coroutine/domain/usecase/GetUsersUseCase.kt
```diff
@@ -1,24 +0,0 @@
...
-        return withContext(ioDispatcher) {
-            try {
-                val result = userRepository.getUsers()
```
sleepylee, line 17: since you are here, I'd like to suggest changing the name of the related method to `getUser` instead of the plural forms of `Users` 🤔

CoroutineTemplate/app/src/main/java/co/nimblehq/coroutine/dispatcher/DispatchersProviderImpl.kt
doannimble: @sleepylee Oh, I actually didn't care about the order, but anyway let me update to make them more consistent.

CoroutineTemplate/data/src/main/java/co/nimblehq/coroutine/repositoryimpl/UserRepository.kt
doannimble: @sleepylee Yes, they must be removed as redundant.

CoroutineTemplate/domain/src/main/java/co/nimblehq/coroutine/domain/usecase/GetUsersUseCase.kt
doannimble: @sleepylee Yep, I thought I had, but not. I will update by removing the pagination argument.

CoroutineTemplate/domain/build.gradle
doannimble: @sleepylee `1` is correct but should be put with version name 😄

CoroutineTemplate/app/src/main/java/co/nimblehq/coroutine/dispatcher/DispatchersProviderImpl.kt
```diff
@@ -0,0 +1,14 @@
...
+
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.MainCoroutineDispatcher
+
+class DispatchersProviderImpl : DispatchersProvider {
+
+    override fun getIO(): CoroutineDispatcher = Dispatchers.IO
+
+    override fun getMain(): MainCoroutineDispatcher = Dispatchers.Main
```
sleepylee, line 11: I think we can confront with the Interface return type to return explicitly `CoroutineDispatcher` as defined from the interface; upcasting in this case doesn't hurt imo, instead of returning a specific type `MainCoroutineDispatcher`?

CoroutineTemplate/app/src/main/java/co/nimblehq/coroutine/ui/base/BaseViewModel.kt
sleepylee: this seem to force every `job` to be executed always with `dispatchers.getIO()` 🤔  so in case we want to switch to default, the API doesn't seem to be very open for it; perhaps you can pass the `CoroutineDispatcher` in the params and default it as `IO()` instead?

CoroutineTemplate/build.gradle
```diff
@@ -51,13 +50,12 @@ buildscript {
 }
 
 plugins {
-    id "io.gitlab.arturbosch.detekt" version "1.14.0"
+    id "io.gitlab.arturbosch.detekt" version "1.18.1"
```
sleepylee, line 53: there seem to be quite many unrelated changes to the actual intention of this PR: `Refactor architecture` - next time please scope the PR appropriately so it doesn't cost too much time of the other to review 🙇

CoroutineTemplate/data/src/main/java/co/nimblehq/coroutine/data/service/providers/MoshiBuilderProvider.kt
```diff
@@ -1,14 +1,11 @@
...
 import com.squareup.moshi.Moshi
-import com.squareup.moshi.adapters.Rfc3339DateJsonAdapter
 import com.squareup.moshi.kotlin.reflect.KotlinJsonAdapterFactory
-import java.util.*
 
 object MoshiBuilderProvider {
 
     val moshiBuilder: Moshi.Builder
         get() = Moshi.Builder()
-            .add(Date::class.java, Rfc3339DateJsonAdapter())
```
sleepylee, line 12: it's a utility that most of the time we face with the API Date response type - does this relate to the `Clean Code refactor`?

CoroutineTemplate/domain/build.gradle
```diff
@@ -1,41 +1,19 @@
...
-        project(':data'),
-
-        "androidx.core:core-ktx:$androidx_core_ktx_version",
-        "androidx.security:security-crypto:$android_crypto_version",
-        "com.google.dagger:hilt-android:$hilt_version",
-        "org.jetbrains.kotlinx:kotlinx-coroutines-android:$kotlinx_coroutines_version"
+    api(
+        "com.squareup.moshi:moshi-kotlin:$moshi_version",
+        "com.squareup.retrofit2:converter-moshi:$retrofit_version",
+        "com.squareup.moshi:moshi:$moshi_version",
```
sleepylee, line 15: I have a feeling that we wouldn't want to expose `moshi` or `retrofit` to the upper level 🤔  I mean - domain will handle all the related jobs to the API and returning response to the `:data` to continue on processing right? So no module actually needs to know about these libs - hence `implementation` would work imo. Please correct me if I missed anything

CoroutineTemplate/domain/src/main/java/co/nimblehq/coroutine/usecase/GetUsersUseCase.kt
```diff
@@ -0,0 +1,12 @@
+package co.nimblehq.coroutine.usecase
+
+import co.nimblehq.coroutine.entity.UserEntity
+import co.nimblehq.coroutine.repository.UserRepository
+import javax.inject.Inject
+
+class GetUsersUseCase @Inject constructor(private val userRepository: UserRepository) {
+
+    suspend fun execute(): UseCaseResult<List<UserEntity>> {
+        return userRepository.getUsers()
```
sleepylee, line 10: shouldn't you catch the exception here?

sleepylee (changes requested):

CoroutineTemplate/build.gradle
doannimble: @sleepylee I see, I will change this kind of habit 🙏

CoroutineTemplate/data/src/main/java/co/nimblehq/coroutine/data/service/providers/MoshiBuilderProvider.kt
doannimble: @sleepylee It should have a comment to explain why we need to add it?
I thought it was unnecessary as I couldn't see its usage in the template 🤔

CoroutineTemplate/domain/build.gradle
doannimble: @sleepylee You're correct, each module has its responsibility, then we don't need to expose its adapter. 
Let me correct it.

CoroutineTemplate/data/src/main/java/co/nimblehq/coroutine/data/service/providers/MoshiBuilderProvider.kt
sleepylee: I agree this might seem to be odd without the comment and the usage 🤔  but I think it's only hard to the new users to the template 🤔  `Rfc3339` is a standard way of returning Date in this format: `2019-10-12T07:20:50.52Z` from the API - we faced this most of the time so parsing this become a norm to us 🤔 .

https://medium.com/easyread/understanding-about-rfc-3339-for-datetime-formatting-in-software-engineering-940aa5d5f68a

Perhaps you can add a comment or question in a separate PR later instead of unilaterally removing in this unrelated PR?

doannimble: @sleepylee Yep, I think I can revert it and add the comment inside like `// Utility for the API Date response type`. 
What do you think?

sleepylee: or you can make it even more specific: 
`// Parse the DateTime in this format: [2019-10-12T07:20:50.52Z]`

CoroutineTemplate/domain/src/main/java/co/nimblehq/coroutine/usecase/GetUsersUseCase.kt
doannimble: @sleepylee No need, we have `try/catch` in `RepositoryImpl` already to make sure the `Repository` always returns UseCaseResult.

sleepylee: erh....why would Repository know about the UseCase?

CoroutineTemplate/app/src/main/java/co/nimblehq/coroutine/ui/base/BaseViewModel.kt
doannimble: @sleepylee Okay, I think that is what @Wadeewee wants to. So let update to make it more open.

CoroutineTemplate/domain/src/main/java/co/nimblehq/coroutine/usecase/GetUsersUseCase.kt
doannimble: @sleepylee Ah the `UseCaseResult` is just a `Model Class` that contains Result or Error.
Certainly, the Repository would not know about the UseCase.

CoroutineTemplate/data/src/main/java/co/nimblehq/coroutine/repositoryimpl/UserRepository.kt
```diff
@@ -0,0 +1,21 @@
...
+class UserRepositoryImpl constructor(
+    private val apiService: ApiService,
+) : UserRepository {
+
+    override suspend fun getUsers(): UseCaseResult<List<UserEntity>> {
+        return try {
+            val response = apiService.getUsers()
+            UseCaseResult.Success(response.toUserEntities())
+        } catch (e: Exception) {
+            UseCaseResult.Error(e)
```
sleepylee, line 18: now I see the systematic issue here, @doannimble. Why would we let the `Repository` returns `UseCase` result? 🤔  it's kinda crossing the line. Imagine if you have a different Service that doesn't run through UseCase class, the `Repository` become very odd to returns `UseCaseResult` wrapper in such case.

CoroutineTemplate/domain/src/main/java/co/nimblehq/coroutine/usecase/GetUsersUseCase.kt
sleepylee: @doannimble there is a flaw in this design, combining with the comment I put above:
- The UseCaseResult.Error is very useless atm, it doesn't contain a useful Exception type for further handling.
- The Repository acts as the source of connecting various data type, while it can leverage multiple sources, it doesn't have to tie the result to the downstream classes (UseCase) that are actually using it.

I'm not sure why wouldn't we just through the Exception to the UseCase, capturing it and returns as UseCaseResult.Error or .Success? In the end, the UseCase expects to returns `UseCaseResult` from the `call()` method right?

CoroutineTemplate/data/src/main/java/co/nimblehq/coroutine/repositoryimpl/UserRepository.kt
doannimble: @sleepylee Perhaps the name of `UseCaseResult` makes you confuse.
It's just the `Data Model` of `Repository` and the `UseCase` must depend on it.
I usually use with name `Result`. Do I need to refactor the name of `UseCaseResult` to avoid confusion?

sleepylee: I think it's better that we return the type as we did before for now. I see that there are some resources on Medium that refer to returns in this way, but it doesn't go along with our intention.

- Your approach is:
`Repository returns Result [ Success, Error] - UseCase is dumb, it returns the Result that belongs to Repository`

https://proandroiddev.com/why-you-need-use-cases-interactors-142e8a6fe576

- While our approach currently is:

`Repository returns raw <Type> and throw exception or onError with RxJava; UseCase has UseCaseResult.Success or Error that can be manipulated by the ViewModel`

https://github.com/android10/Android-CleanArchitecture/blob/master/domain/src/main/java/com/fernandocejas/android10/sample/domain/repository/UserRepository.java

CoroutineTemplate/domain/src/main/java/co/nimblehq/coroutine/domain/usecase/GetUsersUseCase.kt
```diff
@@ -1,24 +0,0 @@
...
-                UseCaseResult.Success(result)
-            } catch (exception: Exception) {
-                UseCaseResult.Error(exception)
```
sleepylee, line 20: check this intention again, @doannimble - you are changing the return type and the rule we intended to set on the responsibility of the [Repository - UseCase]. I think it is out of the scope of the PR again

doannimble: @sleepylee Agree, it should be in another PR with further discussion.
I will keep the previous approach to avoid making more complicated.

CoroutineTemplate/data/src/main/java/co/nimblehq/coroutine/repositoryimpl/UserRepository.kt
doannimble: @sleepylee Okay, we may have another discussion about this in other PR.
I will keep our current approach and I think it is better if we have a deep understanding of each to answer the question: why we use this one, not that one.

sleepylee: here is another ref I found that they followed your style (in this case, they returned `Resource` == your `Result`) at the Repository layer: https://github.com/android/architecture-components-samples/blob/88747993139224a4bb6dbe985adf652d557de621/GithubBrowserSample/app/src/main/java/com/android/example/github/repository/UserRepository.kt#L40

It's from Google's sample, notice the difference is that they are not using `Interactor` or `UseCase` like us, so their return type is being wrapped properly before sending back to the ViewModel directly.

For us, we segregated the responsibility more via an interim layer, so, as the thumb rule: we must return the wrapped type back to the ViewModel to avoid further processing; doing so at the UseCase can reduce the heavy logic at the Repository layer. Reducing the size of each class is always better than creating more dumb classes to invoke the other layer blindly imo. 🙏 

Let me know your thoughts in the end.

AVI5HEK: In my humble opinion, wrapping the data inside `UseCaseResult` doesn't violate anything. But I think it would be better to leave this job to the UseCase. Sometimes the UseCase does manipulations on the data or has internal logic based on the data. In those cases, unwrapping it inside the UseCase seems redundant. So I would recommend making the method return `List<UserEntity>` only.

CoroutineTemplate/domain/src/main/java/co/nimblehq/coroutine/entity/UserEntity.kt
```diff
@@ -0,0 +1,24 @@
...
+data class UserEntity(
+    val id: Int?,
+    val name: String,
+    val username: String,
+    val email: String?,
+    val address: Address?,
+    val phone: String,
+    val website: String
+) {
+    data class Address(
```
AVI5HEK, line 12: No biggie, but our [convention](https://nimblehq.co/compass/development/code-conventions/kotlin/#formatting) requires a blank line after the class declaration and before the first line of code inside the class. 😄

CoroutineTemplate/domain/src/main/java/co/nimblehq/coroutine/entity/UserEntity.kt
```diff
@@ -0,0 +1,24 @@
...
+        val street: String,
+        val suite: String,
+        val city: String,
+        val zipCode: String,
+        val geo: Geo?
+    ) {
+        data class Geo(
```
AVI5HEK, line 19: The same blank line convention goes here as well.

CoroutineTemplate/domain/src/main/java/co/nimblehq/coroutine/usecase/GetUsersUseCase.kt
AVI5HEK: @doannimble `RepositoryImpl` should catch exceptions related to I/O such as network, database, etc. and map them to some common errors (e.g., NoConnectionError, ApiHttpError, etc.) recognized by the UseCase. It's up to the UseCase to catch those errors and map to more specific errors that would be consumed by the `presentation` layer, e.g., `DefaultUserError`, `CreateUserError`, etc.

doannimble: @AVI5HEK Yes, I'm doing like that, `RepositoryImpl` just catch exceptions related to Network so far.
Hence I think, the Error mapper should happen on `UseCase`

CoroutineTemplate/data/src/main/java/co/nimblehq/coroutine/repositoryimpl/UserRepository.kt
doannimble: @sleepylee @AVI5HEK Okay, so I agree with you guys, we should let `UseCase` handle all of the business logic, included wrapping data.
For example, let say if we want to save a token of `UserEntity` after logged in into` SharedPreference`, we can do it as a flow on try/catch inside `UseCase` without unwrapping data.

CoroutineTemplate/app/src/main/java/co/nimblehq/coroutine/ui/screens/compose/composables/UserItem.kt
```diff
@@ -40,11 +40,11 @@ fun UserItem(
             ) {
                 with(user) {
                     Text(
-                        text = name,
+                        text = name.orEmpty(),
```
sleepylee, line 43: perhaps for later, we should create a UserUiModel that handles this logic from the UserEntity. As `name` can be null, then `UiModel` can prepare the `empty string` for this field instead of asking the View to manage the null case itself

CoroutineTemplate/app/src/main/java/co/nimblehq/coroutine/ui/screens/home/HomeViewModel.kt
```diff
@@ -49,12 +49,14 @@ class HomeViewModel @Inject constructor(
 
     private fun fetchUsers() {
         showLoading()
-        viewModelScope.launch {
-            when (val result = getUsersUseCase.execute()) {
-                is UseCaseResult.Success -> _users.value = result.data
-                is UseCaseResult.Error -> _error.emit(result.exception.message.orEmpty())
+        execute(kotlin.run {
```
sleepylee, line 52: 🤔  `kotlin.run`? 🤔  it seems a bit odd for me unless we are having a namespace conflict

CoroutineTemplate/data/build.gradle
```diff
@@ -29,6 +29,13 @@ android {
 }
 
 dependencies {
+    implementation(
+        project(':domain'),
+        "androidx.core:core-ktx:$androidx_core_ktx_version",
+        "androidx.security:security-crypto:$android_crypto_version",
+        "com.google.dagger:hilt-android:$hilt_version"
+    )
```
sleepylee, line 37: 👍

CoroutineTemplate/domain/src/main/java/co/nimblehq/coroutine/entity/UserEntity.kt
```diff
@@ -1,8 +1,8 @@
-package co.nimblehq.coroutine.data.service.response
+package co.nimblehq.coroutine.entity
 
 import com.squareup.moshi.Json
 
-data class UserResponse(
+data class UserEntity(
     @Json(name = "id") val id: Int?,
     @Json(name = "name") val name: String?,
     @Json(name = "username") val username: String?,
```
sleepylee, line 8: hmmm...this is a bit strange for me; right now this `UserEntity` with json serializer is being used at the `ApiService` (networking) - passing through the `Respository` - then to `UseCase` and then serving to `ViewModel` and `View.`. there is no segregation in between any of these, so when there is change, let's say the API responds a change in the structure, we will have to make changes at every layer.

sleepylee: the best is to keep `UserResponse` from the network layer. Then having another `UserEntity` from `UserResponse.toUserEntity()` and passing it through would be the best. It's cumbersome a little bit, but it's better for us to isolate the concern

sleepylee (commented):
some last comments I hope

CoroutineTemplate/domain/src/main/java/co/nimblehq/coroutine/entity/UserEntity.kt
doannimble: @sleepylee If we do that, we need to put try/catch and mapping from `UserRepositoryImpl` As :
- `UserRepository` comes from `domain` and it does not have information on `UserResponse`
With the current implementation, I think we can add `UserModel` on `domain` return `List<UserModel>` in `UseCase`.

CoroutineTemplate/app/src/main/java/co/nimblehq/coroutine/ui/screens/home/HomeViewModel.kt
doannimble: @sleepylee Yes, no need for the prefix `kotlin.` . I'll remove it.

CoroutineTemplate/app/src/main/java/co/nimblehq/coroutine/ui/screens/compose/composables/UserItem.kt
doannimble: @sleepylee I think I need to return the `UserModel` instead of `UserEntity`

CoroutineTemplate/domain/src/main/java/co/nimblehq/coroutine/entity/UserEntity.kt
sleepylee: @doannimble that would work: separating the `Response` we often define serialize annotation (for parsing logic) with the actual `Model` we make public is something we usually do (as can be seen from the previous implementation). For the `Model`, I'd recommend to just make it `User` - it should be clear enough imo. So we shall have:

Repository -> returning Response type - which is directed from the API response
to UseCase -> returning entity/model that is mapped from the Response - ready to be used by the downstream.

doannimble: @sleepylee AFAIK, the Repository Interface must be defined in `domain`, and in `data`, we create an implementation of Repository Interface.
I wonder how can we set : Repository -> returning Response type - which is directed from the API response 🤔

sleepylee: @doannimble I mean: just revert the name `UserEntity` back to `UserResponse`. 

As you said, in the end, the implementation `UserRepoImpl` at the `:data` will comply with that return type from `:domain`, in fact, it's returning the actual response we parsed from the network request, right? 🤔  as I can see, we are not doing any extra things when receiving the response from ApiService with this Entity class

doannimble: @sleepylee Yes, that's correct. So conclusion : 
- Revert name `UserEntity` to `UserResponse` (in `domain`)
- Create `User` that map from `UserResponse` and expose on `UseCase` (in `domain`).

Is it OK?

sleepylee: sgtm 👍

sleepylee: p/s: my target is that the `Entity` classification can be reserved for the `RoomEntity`, just FYI 😉

doannimble: @sleepylee Thanks for your information. I usually use `Entity` for both `Room` and `Network Response` 😄 .
But it should be only for `Room`.

CoroutineTemplate/domain/src/main/java/co/nimblehq/coroutine/usecase/GetUsersUseCase.kt
```diff
@@ -1,15 +1,16 @@
...
 import javax.inject.Inject
 
 class GetUsersUseCase @Inject constructor(private val userRepository: UserRepository) {
 
-    suspend fun execute(): UseCaseResult<List<UserEntity>> {
+    suspend fun execute(): UseCaseResult<List<User>> {
         return try {
             val response = userRepository.getUsers()
-            UseCaseResult.Success(response)
+            UseCaseResult.Success(response.toUserModel())
```
sleepylee, line 13: you can shorten this method name now as `toUser()`

CoroutineTemplate/domain/src/main/java/co/nimblehq/coroutine/response/UserResponse.kt
```diff
@@ -0,0 +1,60 @@
...
+    return User.Address(
+        street = this.street.orEmpty(),
+        suite = this.suite.orEmpty(),
+        city = this.city.orEmpty(),
+        zipCode = this.zipCode.orEmpty(),
+        geo = this.geo?.toModel()
+    )
+}
+
+private fun UserResponse.Address.Geo.toModel(): User.Address.Geo {
```
sleepylee, line 55: to keep these inner response class different from the model class, how about making:

`UserResponse > AddressResponse > GeoResponse` 
extension: `GeoResponse.toGeo()`; `AddressResponse.toAddress()`?

CoroutineTemplate/domain/src/main/java/co/nimblehq/coroutine/model/User.kt
```diff
@@ -0,0 +1,24 @@
+package co.nimblehq.coroutine.model
+
+data class User(
+    val id: Int?,
+    val name: String,
+    val username: String,
+    val email: String?,
```
sleepylee, line 7: seems like we put the email fallback as empty and never null

sleepylee (commented):
some last touches

CoroutineTemplate/domain/src/main/java/co/nimblehq/coroutine/response/UserResponse.kt
doannimble: @sleepylee It makes sense. Let me update them.

CoroutineTemplate/app/src/main/java/co/nimblehq/coroutine/ui/screens/home/HomeViewModel.kt
```diff
@@ -49,12 +49,14 @@ class HomeViewModel @Inject constructor(
 
     private fun fetchUsers() {
         showLoading()
-        viewModelScope.launch {
-            when (val result = getUsersUseCase.execute()) {
-                is UseCaseResult.Success -> _users.value = result.data
-                is UseCaseResult.Error -> _error.emit(result.exception.message.orEmpty())
+        execute(run {
+            {
```
sleepylee, line 53: this seems redundant with the above `{` 🤔

CoroutineTemplate/domain/src/main/java/co/nimblehq/coroutine/response/UserResponse.kt
```diff
@@ -0,0 +1,60 @@
...
+        @Json(name = "geo") val geoResponse: GeoResponse?
+    ) {
+        data class GeoResponse(
+            @Json(name = "lat") val latitude: String?,
+            @Json(name = "lng") val longitude: String?
+        )
+    }
+}
+
+fun List<UserResponse>.toUser(): List<User> {
```
sleepylee, line 29: `toUsers` - in this case we are mapping List to List (plural) 😄

sleepylee (commented):
few last comments

CoroutineTemplate/domain/src/main/java/co/nimblehq/coroutine/model/User.kt
doannimble: @sleepylee Updated

CoroutineTemplate/app/src/main/java/co/nimblehq/coroutine/ui/screens/home/HomeViewModel.kt
doannimble: @sleepylee It's necessary for `run` block.
But I realize that we don't need to use it in this case. Just remove it.

sleepylee (approved):
lgtm now. Please take notes on the points we went through this PR and I hope you will take care better of your future PRs 🙇  at least target to have <10 comments on a small/medium PR or <20 comments on a big PRs would be a good setup to target to.

CoroutineTemplate/data/src/main/java/co/nimblehq/coroutine/data/service/response/UserResponse.kt
lydiasama: @AVI5HEK Hmm 🤔  I agree at this point. `id = this.id` is clearer.

CoroutineTemplate/domain/src/main/java/co/nimblehq/coroutine/usecase/GetUsersUseCase.kt
```diff
@@ -0,0 +1,18 @@
...
+import javax.inject.Inject
+
+class GetUsersUseCase @Inject constructor(private val userRepository: UserRepository) {
+
+    suspend fun execute(): UseCaseResult<List<User>> {
+        return try {
+            val response = userRepository.getUsers()
+            UseCaseResult.Success(response.toUsers())
+        } catch (e: Exception) {
+            UseCaseResult.Error(e)
```
AVI5HEK, line 15: Can we have custom exceptions for this UseCase in this PR or will we have it sometime later in another PR?

CoroutineTemplate/domain/src/main/java/co/nimblehq/coroutine/response/UserResponse.kt
```diff
@@ -0,0 +1,60 @@
+package co.nimblehq.coroutine.response
+
+import co.nimblehq.coroutine.model.User
+import com.squareup.moshi.Json
+
+data class UserResponse(
```
AVI5HEK, line 6: Why do we have this file and the classes in this module? This should be in the `data` module. 😕

CoroutineTemplate/domain/src/main/java/co/nimblehq/coroutine/repository/UserRepository.kt
```diff
@@ -0,0 +1,8 @@
+package co.nimblehq.coroutine.repository
+
+import co.nimblehq.coroutine.response.UserResponse
+
+interface UserRepository {
+
+    suspend fun getUsers(): List<UserResponse>
```
AVI5HEK, line 7: `Repository` should have only `domain` reference and `RepositoryImpl` should do the mappings (`data` models to `domain` models).
```suggestion
    suspend fun getUsers(): List<User>
```

CoroutineTemplate/domain/src/main/java/co/nimblehq/coroutine/model/User.kt
```diff
@@ -0,0 +1,24 @@
...
+data class User(
+    val id: Int?,
+    val name: String,
+    val username: String,
+    val email: String,
+    val address: Address?,
+    val phone: String,
+    val website: String
+)  {
+    data class Address(
```
AVI5HEK, line 12: Blank line before the first line inside the class as a convention. 😅

CoroutineTemplate/domain/src/main/java/co/nimblehq/coroutine/model/User.kt
```diff
@@ -0,0 +1,24 @@
...
+        val street: String,
+        val suite: String,
+        val city: String,
+        val zipCode: String,
+        val geo: Geo?
+    ) {
+        data class Geo(
```
AVI5HEK, line 19: Blank line before this one too.

CoroutineTemplate/domain/build.gradle
```diff
@@ -1,41 +1,19 @@
...
     implementation(
-        project(':data'),
-
-        "androidx.core:core-ktx:$androidx_core_ktx_version",
-        "androidx.security:security-crypto:$android_crypto_version",
-        "com.google.dagger:hilt-android:$hilt_version",
-        "org.jetbrains.kotlinx:kotlinx-coroutines-android:$kotlinx_coroutines_version"
+        "com.squareup.moshi:moshi-kotlin:$moshi_version",
+        "com.squareup.retrofit2:converter-moshi:$retrofit_version",
+        "com.squareup.moshi:moshi:$moshi_version",
```
AVI5HEK, line 15: These dependencies should be in the `data` module since all the networking and storage-related stuff is the responsibility of `data`.

CoroutineTemplate/data/src/main/java/co/nimblehq/coroutine/repositoryimpl/UserRepository.kt
```diff
@@ -0,0 +1,14 @@
...
+import co.nimblehq.coroutine.response.UserResponse
+import co.nimblehq.coroutine.repository.UserRepository
+
+class UserRepositoryImpl constructor(
+    private val apiService: ApiService,
+) : UserRepository {
+
+    override suspend fun getUsers(): List<UserResponse> {
+        return apiService.getUsers()
+    }
```
AVI5HEK, line 13:
```suggestion
    override suspend fun getUsers(): List<User> {
        return apiService.getUsers().toUsers()
    }
```

CoroutineTemplate/app/src/main/java/co/nimblehq/coroutine/ui/screens/home/HomeViewModel.kt
```diff
@@ -15,7 +15,7 @@ import javax.inject.Inject
 
 interface Output {
 
-    val users: StateFlow<List<UserEntity>>
+    val users: StateFlow<List<User>>
```
AVI5HEK, line 18: We're supposed to have a different model for the `presentation` layer. We map the `domain` model to the `presentation` model inside `ViewModel`s. How about we do it in this PR? 💡 
```suggestion
    val users: StateFlow<List<UserUiModel>>
```

AVI5HEK (commented):
My take on the latest changes. 🚀

CoroutineTemplate/domain/src/main/java/co/nimblehq/coroutine/usecase/GetUsersUseCase.kt
doannimble: @AVI5HEK I think we will need another PR to do this.

CoroutineTemplate/domain/src/main/java/co/nimblehq/coroutine/response/UserResponse.kt
doannimble: @AVI5HEK It should be in `data` module. However, we're aiming to return the original data from `Repository`, that's the reason why there is a bit different compare with the standard.

CoroutineTemplate/domain/src/main/java/co/nimblehq/coroutine/repository/UserRepository.kt
doannimble: @AVI5HEK If we do that, we have to `try/cach`  and return the wrap Result inside `RepositoryImpl` as my previous implementation?

CoroutineTemplate/domain/build.gradle
doannimble: @AVI5HEK Actually, they are adapter for converter factories, you can use them for many purposes. In the current implementation, we're putting the `Response at `domain`, hence, it depends on our last decision to keep the `Response` or not in `domain`

CoroutineTemplate/app/src/main/java/co/nimblehq/coroutine/ui/screens/home/HomeViewModel.kt
doannimble: @AVI5HEK Honestly, I did not see any benefit of this mapping, could you please explain more use cases in real project that we need this step? 🙏

lydiasama:
@doannimble Could you please kindly click on the `Resolve Convwesation`🥺 ? Sometimes I feel hard to know if you resolved it already. 😅

doannimble:
@manh-t Nope, we cannot. As we will get the error on build :

`public final class MainViewModel extends co.nimblehq.coroutine.ui.base.BaseViewModel {
             ^
  @HiltViewModel annotated class should contain exactly one @Inject annotated constructor.
  [Hilt] Processing did not complete. See error above for details.`

PR #138, opened 2021-10-13
[#111] Gradle scripts with Kotlin DSL - Part 2 (Coroutine Template)

hoangnguyen92dn:
https://github.com/nimblehq/android-templates/issues/111

## What happened 👀

Kotlin DSL is now supported to handle Gradle scripts with Kotlin syntax. It's more familiar than Groovy-based syntax for Android developers.

-> Convert some common Groovy-based Gradle scripts into Kotlin DSL scripts.
 
## Insight 📝

- Create new `buildSrc` folder to support Kotlin-DSL
- Migrate Gradle scripts to `gradle.kts` supported Kotlin on `Coroutine` project

👉  Part 3 🏃 :
- Migrate `jacoco.gradle` & `codequality.gradle` to `.kts`
 
## Proof Of Work 📹

🚀 CI build should be passed ✅


chornerman (approved):

CoroutineTemplate/build.gradle.kts
```diff
@@ -0,0 +1,63 @@
+// Top-level build file where you can add configuration options common to all sub-projects/modules.
+
+buildscript {
+
```
lydiasama, line 4: Redundant line break?

CoroutineTemplate/buildSrc/src/main/java/Configurations.kt
```diff
@@ -0,0 +1,15 @@
...
+
+object BuildType {
+    const val DEBUG = "debug"
+    const val RELEASE = "release"
+}
+
+object Module {
+    const val DATA = ":data"
+    const val DOMAIN = ":domain"
+}
```
lydiasama, line 15: Should we create these objects? I feel like it uses only one place and can be just a string.
Even the Module at [this line](https://github.com/nimblehq/android-templates/pull/138/files#diff-c24cd0177f073063f92b7261a67d0cf5ab26020289bfe8d940b844b413a54b52R2), we didn't use this object. 🤔

CoroutineTemplate/data/build.gradle.kts
```diff
@@ -0,0 +1,61 @@
...
+        versionName = Versions.ANDROID_VERSION_NAME
+
+        testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
+        consumerProguardFiles("consumer-rules.pro")
+    }
+
+    buildTypes {
+        getByName(BuildType.RELEASE) {
+            isMinifyEnabled = false
+            proguardFiles(
```
chornerman, line 21: Same here 🙏🏼

CoroutineTemplate/app/build.gradle.kts
```diff
@@ -0,0 +1,116 @@
...
+        versionCode = Versions.ANDROID_VERSION_CODE
+        versionName = Versions.ANDROID_VERSION_NAME
+    }
+
+    buildTypes {
+        getByName(BuildType.RELEASE) {
+            isMinifyEnabled = true
+            isDebuggable = false
+            isShrinkResources = true
+            proguardFiles(
```
chornerman, line 39: (Optional) Make this statement one line 🙏🏼
`proguardFiles(getDefaultProguardFile("proguard-android.txt"), "proguard-rules.pro")`

luongvo: too long I think :)

luongvo (approved):
lgtm

CoroutineTemplate/buildSrc/src/main/java/Configurations.kt
hoangnguyen92dn: @lydiasama 
> Even the Module at this line, we didn't use this object.

I did take a FIXME note to mention why we don't use the constants instead. We will need the solution to resolve it in the future.

CoroutineTemplate/build.gradle.kts
hoangnguyen92dn: @lydiasama It's from the old `build.gradle`. I will remove it.

CoroutineTemplate/app/build.gradle.kts
hoangnguyen92dn: @luongvo I think it's good to make it single line 🤔

CoroutineTemplate/buildSrc/src/main/java/Configurations.kt
lydiasama: @hoangnguyen92dn That's why I don't think we need to use it as an object 😄 . Shall we use a simple String instead? The same idea to the `Flavor` and `BuildType`.

hoangnguyen92dn: @lydiasama Let's keep it consistent with current approach on RxJava Template as well 🙏

chornerman (approved):
lgtm 🎉

PR #142, opened 2021-11-02
[chore] Bump version to 3.3.0

lydiasama:
## What happened 👀

Bump version to 3.3.0 for development

### DIFF
--- a/RxJavaTemplate/buildSrc/src/main/java/Versions.kt
+++ b/RxJavaTemplate/buildSrc/src/main/java/Versions.kt
@@ -2,7 +2,7 @@ object Versions {
      const val BUILD_GRADLE_VERSION = "4.2.2"
 
      const val ANDROID_VERSION_CODE = 9
-     const val ANDROID_VERSION_NAME = "3.2.0"
+     const val ANDROID_VERSION_NAME = "3.3.0"
      const val ANDROID_MIN_SDK_VERSION = 23
      const val ANDROID_TARGET_SDK_VERSION = 28
      const val ANDROID_COMPILE_SDK_VERSION = 28
--- a/CoroutineTemplate/build.gradle
+++ b/CoroutineTemplate/build.gradle
@@ -9,7 +9,7 @@ buildscript {
         android_target_sdk_version = 30
 
         android_version_code = 1
-        android_version_name = '3.2.0'
+        android_version_name = '3.3.0'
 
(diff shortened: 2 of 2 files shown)


RxJavaTemplate/buildSrc/src/main/java/Versions.kt
```diff
@@ -1,60 +1,60 @@
 object Versions {
-     const val BUILD_GRADLE_VERSION = "4.2.2"
```
doannimble, line 2: @lydiasama Is it an unrelated change 😄  ?
Same for others.

lydiasama: It's from my AS formatting automatically. 😱

luongvo: @hoangnguyen92dn why did we use 5 spaces to tab in your last PR?

lydiasama: I think my format is correct, we use 4 spaces instead of 5. @doannimble

luongvo (approved):
lgtm

chornerman (approved):

RxJavaTemplate/buildSrc/src/main/java/Versions.kt
doannimble: @lydiasama Yeah, I know your format is correct and comes from AS.
As I used to be warned it when pushing code with auto-format code from 7-11 project, then I just wanna make a confirmation about this one :D.

doannimble (approved):

RxJavaTemplate/buildSrc/src/main/java/Versions.kt
lydiasama: @doannimble Ok. Then I'll generate another ticket to format this file. It's just a format imo. 😌

doannimble: @lydiasama Totaly agree with you, it does not a matter or a big thing. But it would be great if we can make our process consistent even it looks a bit cumbersome 🙏

lydiasama:
@doannimble Done. 🙆🏻‍♀️

lydiasama:
@doannimble Can I have your approval again?

doannimble (approved):

Issue #148, opened 2021-11-30
Add android-common-ktx library


hoangnguyen92dn:
## Why

We published the first patch version of `android-common-ktx` with Android's component extensions for usage. By adding this library, we will maintain the extensions outside `android-templates` scope.

## Who Benefits?

Android developers

Tuubz:
If I'm not mistaken, we can close this issue? 😁

PR #149, opened 2021-11-30
[#148] Add android-common-ktx library

hoangnguyen92dn:
https://github.com/nimblehq/android-templates/issues/148

## What happened 👀

We published the first patch version of `android-common-ktx` with Android's component extensions for usage. By adding this library, we will maintain the extensions outside `android-templates` scope.

## Insight 📝

- Add `android-common-ktx` on RxJava and Coroutine Template
- Remove unused extensions

## Proof Of Work 📹

N/A


lydiasama:
@hoangnguyen92dn I wonder should we vote on the issue before creating the PR? 😄

hoangnguyen92dn:
@lydiasama 🙆‍♂️  So we would pending this PR for voting...

Tuubz:
@hoangnguyen92dn I see this pull request is marked as "Draft", I'm guessing it's ready for review? 😁

CoroutineTemplate/app/src/main/AndroidManifest.xml
```diff
@@ -12,7 +13,8 @@
         android:networkSecurityConfig="@xml/network_security_config"
         android:roundIcon="@mipmap/ic_launcher_round"
         android:supportsRtl="true"
-        android:theme="@style/AppTheme">
+        android:theme="@style/AppTheme"
+        tools:replace="android:theme">
```
Tuubz, line 17: Just curious, what's the reason we had to make this change? 🙏

hoangnguyen92dn: @Tuubz The project build fails without this change. I am finding solution for it, that's why the PR is masked as Draft 🙈

lydiasama (approved):

doannimble (approved):
LGTM

Wadeewee:
I'm curious about if I would like to create a new function in `ContextExt` Let say we need to implement for open gallery So I need to add in `android-common-ktx` and then we may be to increase version and then applied it to `android-templates` right 🤔   Is this our expectations?, Correct me if I'm wrong 🙇‍♀️

doannimble:
> I'm curious about if I would like to create a new function in `ContextExt` Let say we need to implement for open gallery So I need to add in `android-common-ktx` and then we may be to increase version and then applied it to `android-templates` right 🤔 Is this our expectations?, Correct me if I'm wrong 🙇‍♀️

@Wadeewee You can create your own in case the `ktx-lib` does not support, or you can contribute it into the lib and increase the version 😄 
We do not intend to put everything into the lib, just a few common things 🤔

Wadeewee:
> > I'm curious about if I would like to create a new function in `ContextExt` Let say we need to implement for open gallery So I need to add in `android-common-ktx` and then we may be to increase version and then applied it to `android-templates` right 🤔 Is this our expectations?, Correct me if I'm wrong 🙇‍♀️
> 
> @Wadeewee You can create your own in case the `ktx-lib` does not support, or you can contribute it into the lib and increase the version 😄 We do not intend to put everything into the lib, just a few common things 🤔

Ah, I agree @doannimble  We should put just common things  so that I think my sample case ☝️ should add in `ktx-lib` then like `getAppSettingsIntent` case 🚀

Wadeewee (approved):

Tuubz (approved):

chornerman (approved):

luongvo (approved):

Issue #133, opened 2021-09-30
Adding the test step for the newproject.sh test script


sleepylee:
## Why

As we discussed recently in the Retrospective Meeting, the current process to initialize a project via the `newproject.sh` script is not being tested when the CI runs. This causes us into trouble when there are changes under the hood of the template implementation that breaks the script or are executed but the project is not buildable.

The solution does not only do we run the unit test on the template like a normal project, but we have to also test our init script that works as expected.

## Who Benefits?

Users

sleepylee:
@nimblehq/android-chapter does anyone want to take care of this ticket?

PR #151, opened 2021-12-14
[#133] Adding the test step for the newproject.sh test script

hoangnguyen92dn:
https://github.com/nimblehq/android-templates/issues/133

## What happened 👀

As we discussed recently in the Retrospective Meeting, the current process to initialize a project via the newproject.sh script is not being tested when the CI runs. This causes us into trouble when there are changes under the hood of the template implementation that breaks the script or are executed but the project is not buildable.

The solution does not only do we run the unit test on the template like a normal project, but we have to also test our init script that works as expected.

## Insight 📝

Here is the solution to verify `newproject.sh` script:
- Firstly, I added `gradle` command to build the project and execute UnitTest on the generated project to make sure the script works properly.
- The second step is adding a new workflow to execute `newproject.sh` for 2 templates on Github Actions.

## Proof Of Work 📹

- All checks should be passed ✅ 

- Executing script on local

https://user-images.githubusercontent.com/6950766/146744440-b565e0ec-dab6-407d-bff8-09050ea8b5f3.mov

### DIFF
--- a/.github/workflows/verify_newproject_script.yml
+++ b/.github/workflows/verify_newproject_script.yml
@@ -1,0 +1,40 @@
+name: Verify newproject script
+
+on:
+  pull_request:
+    types: [ opened, edited, reopened, synchronize ]
+    branches:
+      - 'feature/**'
+      - 'chore/**'
+      - 'bug/**'
+
+jobs:
+  verify_newproject_script:
+    name: Verify newproject script
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - name: Set up JDK 1.8
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+
+      - name: Checkout source code
+        uses: actions/checkout@v2.3.2
+
+      - name: Cache Gradle
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.gradle/caches/modules-*
+            ~/.gradle/caches/jars-*
+            ~/.gradle/caches/build-cache-*
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*') }}
+          restore-keys: |
+            ${{ runner.os }}-gradle-
+
+      - name: Verify generating new project from RxTemplate
+        run: ./newproject.sh -t rx -p co.myproject.example -n "MyProjectExampleRx"
+
+      - name: Verify generating new project from CoroutineTemplate
+        run: ./newproject.sh -t crt -p co.myproject.example -n "MyProjectExampleCoroutine"
--- a/.github/workflows/verify_newproject_script.yml
+++ b/.github/workflows/verify_newproject_script.yml
@@ -2,7 +2,7 @@ name: Verify newproject script
 
 on:
   pull_request:
-    types: [ opened, edited, reopened, synchronize ]
+    types: [ opened, reopened, synchronize ]
     branches: [ develop ]
 
 jobs:
--- a/.github/workflows/verify_newproject_script.yml
+++ b/.github/workflows/verify_newproject_script.yml
@@ -3,10 +3,7 @@ name: Verify newproject script
 on:
   pull_request:
     types: [ opened, edited, reopened, synchronize ]
-    branches:
-      - 'feature/**'
-      - 'chore/**'
-      - 'bug/**'
+    branches: [ develop ]
 
 jobs:
   verify_newproject_script:
--- a/newproject.sh
+++ b/newproject.sh
@@ -108,13 +108,13 @@ NAME_NO_SPACES=$(echo "$appname" | sed "s/ //g")
 # Copy main folder
 cp -R $OLD_NAME $NAME_NO_SPACES
 
-# clean the old build
+# Clean the old build
 ./$NAME_NO_SPACES/gradlew -p ./$NAME_NO_SPACES clean
-# get rid of idea settings
+# Get rid of idea settings
 rm -rf $NAME_NO_SPACES/.idea
-# get rid of gradle cache
+# Get rid of gradle cache
 rm -rf $NAME_NO_SPACES/.gradle
-# get rid of the git history
+# Get rid of the git history
 rm -rf $NAME_NO_SPACES/.git
 
 # Rename folder structure
--- a/newproject.sh
+++ b/newproject.sh
@@ -108,15 +108,14 @@ NAME_NO_SPACES=$(echo "$appname" | sed "s/ //g")
 # Copy main folder
 cp -R $OLD_NAME $NAME_NO_SPACES
 
+# clean the old build
+./$NAME_NO_SPACES/gradlew -p ./$NAME_NO_SPACES clean
 # get rid of idea settings
 rm -rf $NAME_NO_SPACES/.idea
 # get rid of gradle cache
 rm -rf $NAME_NO_SPACES/.gradle
 # get rid of the git history
 rm -rf $NAME_NO_SPACES/.git
-# get rid of the build
-rm -rf $NAME_NO_SPACES/build
-rm -rf $NAME_NO_SPACES/app/build
 
 # Rename folder structure
 renameFolderStructure() {
@@ -155,12 +154,6 @@ DATA_PACKAGE_DIR=$( renameFolderStructure $DATA_PACKAGE_DIR )
 DOMAIN_PACKAGE_DIR="domain/src/main/java"
 DOMAIN_PACKAGE_DIR=$( renameFolderStructure $DOMAIN_PACKAGE_DIR )
 
-if [ $template = "rx" ]
-then
-  COMMON_RX_PACKAGE_DIR="common-rx/src/main/java"
-  COMMON_RX_PACKAGE_DIR=$( renameFolderStructure $COMMON_RX_PACKAGE_DIR )
-fi
-
 # Rename android test folder structure
 APP_ANDROIDTEST_DIR="app/src/androidTest/java"
 if [ -d APP_ANDROIDTEST_DIR ]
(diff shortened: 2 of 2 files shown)


sleepylee:
could you please update the PoW, @hoangnguyen92dn - since this script will be run only to the later PR when it's merged, the current statement from the PoW isn't valid to see if the PR is working as expected

newproject.sh
```diff
@@ -146,21 +146,75 @@ renameFolderStructure() {
...
+    DATA_TEST_DIR=$( renameFolderStructure $DATA_TEST_DIR )
+fi
+
+DOMAIN_TEST_DIR="domain/src/test/java"
+if [ -d DOMAIN_TEST_DIR ]
+then
+    DOMAIN_TEST_DIR=$( renameFolderStructure $DOMAIN_TEST_DIR )
+fi
+
+if [ $template = "rx" ]
```
sleepylee, line 211: this seems to be repeated - shouldn't we group the sequent actions in the same `if`?

sleepylee:
besides, I just realize that we are not committing build/ or app/build/ so these were redundant from the original script 
https://github.com/nimblehq/android-templates/pull/151/files#diff-19da73e7c5533edd3285c63a0b88034c9e92dfd1e71a28dbccb78154cdef2b28R119

hoangnguyen92dn:
> since this script will be run only to the later PR when it's merged, the current statement from the PoW isn't valid to see if the PR is working as expected

@sleepylee The workflow will be executed include Verify the script step, so I think it's acceptable rather than verifying once merged to `develop` 🤔 

image


> besides, I just realized that we are not committing build/ or app/build/ so these were redundant from the original script

I don't fully get your review 🤔  I added the build and execute all the UnitTest to make sure the generated project is buildable and ready for use. Then on CI, these steps will help us to alert the code changes that affected the script.

sleepylee:
@hoangnguyen92dn 

>The workflow will be executed include Verify the script step...
okay, so the result is published on this current PR as well, which is ok 👌  but I would still recommend to make some screenshot of how the scripts passed from your local running. When we say "just check the CI state" it's not directive enough as Reviewers might not know what workflow should they look for if they are not reading all the changes. Besides, such comment would work when we do a refactor or adding test, writing a test feature like now we should elaborate better what has been changed.

> I don't fully get your review 🤔 ....
re: removing `build/` and `app/build/` please follow the line I pointed from the script - the change was not included in this PR, I just found that from the original implementation it's not necessary.

hoangnguyen92dn:
> removing build/ and app/build/ please follow the line I pointed from the script - the change was not included in this PR, I just found that from the original implementation it's not necessary.

Got it. In my opinion, I think we can leave it there, just in case users build the template before executing the script, then the generated project will be clean to start development.

sleepylee:
@hoangnguyen92dn then maybe "gradlew clean" is a better approach? 🤔

doannimble (approved):
LGTM

newproject.sh
```diff
@@ -185,5 +251,14 @@ APPLICATION_CLASS_PATH="${OLD_APPLICATION_CLASS_PATH/$OLD_NAME/$NAME_NO_SPACES}"
 mv $OLD_APPLICATION_CLASS_PATH $APPLICATION_CLASS_PATH
 echo "✅  Completed"
 
+echo "=> 🛠️ Building generated project..."
+cd $appname
+./gradlew assembleDebug
+echo "✅  Build success"
+
+echo "=> 🚓 Executing all unittest..."
```
Tuubz, line 259: Typo: unit tests? 😊

newproject.sh
```diff
@@ -185,5 +251,14 @@ APPLICATION_CLASS_PATH="${OLD_APPLICATION_CLASS_PATH/$OLD_NAME/$NAME_NO_SPACES}"
...
+./gradlew testStagingDebugUnitTest
+echo "✅  All test passed"
```
Tuubz, line 261: Typo: tests? 😊

newproject.sh
```diff
@@ -169,14 +223,26 @@ echo "✅  Completed"
 echo "=> 🔎 Replacing package and package name within files..."
 PACKAGE_NAME_ESCAPED="${packagename//./\.}"
 OLD_PACKAGE_NAME_ESCAPED="${OLD_PACKAGE//./\.}"
-LC_ALL=C find $WORKING_DIR/$NAME_NO_SPACES -type f -exec sed -i "" "s/$OLD_PACKAGE_NAME_ESCAPED/$PACKAGE_NAME_ESCAPED/g" {} +
-LC_ALL=C find $WORKING_DIR/$NAME_NO_SPACES -type f -exec sed -i "" "s/$OLD_NAME/$NAME_NO_SPACES/g" {} +
+if [[ "$OSTYPE" == "darwin"* ]] #MacOS
```
Tuubz, line 226: Just curious why you had to add this, what issue did you run in to? 🙏

Tuubz (commented):
Cool! ⭐

Tuubz:
> >> removing build/ and app/build/ please follow the line I pointed from the script - the change was not included in this PR, I just found that from the original implementation it's not necessary.
>>
>> Got it. In my opinion, I think we can leave it there, just in case users build the template before executing the script, then the generated project will be clean to start development.
>
> @hoangnguyen92dn then maybe "gradlew clean" is a better approach? 🤔

I don't see where `gradlew clean` is being called as @sleepylee suggested, was there a change in the decision?

.github/workflows/verify_newproject_script.yml
```diff
@@ -0,0 +1,37 @@
+name: Verify newproject script
+
+on:
+  pull_request:
+    types: [ opened, edited, reopened, synchronize ]
```
lydiasama, line 5: I think we can omit the `edited` because the edited type will trigger to run this workflow when we update the title or description of the PR. IMO just opened, reopened, and synchronize are sufficient.

hoangnguyen92dn: @lydiasama Make sense 👍

newproject.sh
hoangnguyen92dn: @Tuubz The original command was working well when I execute the script on local. Unfortunately, the machine on CI is running with Ubuntu instead, and the `find sed` command doesn't work, you can see the fail run [here](https://github.com/nimblehq/android-templates/runs/4542929431?check_suite_focus=true) so that's why I had to update to detect the OS to execute the command accordingly.

hoangnguyen92dn:
> I don't see where `gradlew clean` is being called as @sleepylee suggested, was there a change in the decision?

@Tuubz Sorry I forgot to push the latest update. The `gradlew clean` added in here
https://github.com/nimblehq/android-templates/pull/151/files#diff-19da73e7c5533edd3285c63a0b88034c9e92dfd1e71a28dbccb78154cdef2b28R112

.github/workflows/verify_newproject_script.yml
Tuubz: Oh, what if the pull request gets `updated`, will it be triggered again? I think it should 🙏

hoangnguyen92dn: @Tuubz 👍

Tuubz: I'm surprised to see that there's no option for:

```
on:
  pull_request:
    types: [ updated ]
```

But then I had a look in to [this documentation](https://docs.github.com/en/enterprise-server@3.2/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#webhook-payload-object-29):

> synchronize: Triggered when a pull request's head branch is updated. For example, when the head branch is updated from the base branch, when new commits are pushed to the head branch, or when the base branch is changed.

Do you think `synchronize` will be triggered whenever the pull request is updated? 🤔

Wadeewee (approved):

.github/workflows/verify_newproject_script.yml
hoangnguyen92dn: > Do you think synchronize will be triggered whenever the pull request is updated? 🤔

`updated` in your mind is new commits pushed, right? If so, I have the same idea about `synchronize`, the event will be triggered if there is new commits pushed on the PR.

Tuubz: Updated could be: New commit, rebase then force push, append old commit then force push, ...

If `synchronize` does it for us, then the following code might be redundant:

```
push:
    paths:
      - 'newproject.sh'
    branches:
      - 'feature/**'
      - 'chore/**'
      - 'bug/**'
```

Please correct me if I misunderstood 🙏

hoangnguyen92dn: @Tuubz So basically, we don't need to trigger the workflow on `pull_request` if the `newproject.sh` is not in the list of change files 🤔  What do you think if we just need to trigger on `push`?

hoangnguyen92dn: Following this answer: https://github.community/t/how-to-trigger-an-action-on-push-or-pull-request-but-not-both/16662/2

@Tuubz So from our case, the event triggers should be like this:

```
on:
  pull_request:
    types: [ opened, reopened, synchronize ]
    branches: [ develop ]
  push:
    branches:
      - 'feature/**'
      - 'chore/**'
      - 'bug/**'
```

Tuubz: Hmmm, from my understanding `synchronize` is the same as:

```
push:
    branches:
      - 'feature/**'
      - 'chore/**'
      - 'bug/**'
```

Because `synchronize` will trigger whenever there's an update the pull request. 🤔 

Reference: [StackOverflow](https://stackoverflow.com/questions/65093647/in-github-actions-how-can-i-trigger-on-push-but-only-if-a-pr-is-active)

hoangnguyen92dn: @Tuubz Hmm... So we only want to trigger only not both. I think I would get rid of

```
push:
    branches:
      - 'feature/**'
      - 'chore/**'
      - 'bug/**'
```

🚀

.github/workflows/verify_newproject_script.yml
```diff
@@ -0,0 +1,37 @@
+name: Verify newproject script
+
+on:
+  pull_request:
+    types: [ opened, reopened, synchronize ]
+    branches: [ develop ]
```
Tuubz, line 6: Just curious, any reason we added `branches: [ develop ]` ? I'm thinking if it would be better to let the run script on all branches 🙏 

Rest LGTM 🚀

hoangnguyen92dn: @Tuubz We only need to trigger when we merge the new updates target to `develop` branch. I am curious why do we need to execute on `main` or `release`? 😄

Tuubz: I was thinking about the following case:

1. Create pull request `ONE`, with source branch `A` and target branch `develop`
2. Create pull request `TWO`, with target branch `A`

Then the verify script would not run for pull request `TWO` until we merge pull request `ONE`

Not a big issue, I guess 🙏

Tuubz (approved):

Issue #121, opened 2021-08-30
Add Wiki content for must-have and nice-to-have features


Tuubz:
## Why

Our android-templates repository is becoming a dumping playground, which beats its original purpose to quickly start a new project with all the must-have requirements. Other nice-to-have features need to be considered as an optional setting when starting a new project. 

Anyhow, before we decide on how to integrate/separate nice-to-have features in to/from our code base just by running a command, we need to define it in words first, on how to differentiate a must-have feature from a nice-to-have feature.

Read more: [thread](https://nimble-co.slack.com/archives/G41UCTGHL/p1629954695014500) 

## Who Benefits?

Once the @nimblehq/android-chapter are aligned on the same guidelines, we can smoothen up the decision making.

Later on this can be moved on to other chapters too 🚀 

PR #153

Tuubz:
I'll aim to finish by end of January 🚀

PR #153, opened 2022-01-17
[#121] Update wiki-content for essential, optional and rejected features

Tuubz:
#121

## What happened 👀

To prevent this repository from becoming a dumping playground, we need to define some ground rules. Originally we were planning to define must-have vs nice-to-have features, but after some discussions, we realized there's a third category when the feature should not be included in the template at all. So I've revised the wording to the following:

- **Essential**
- **Optional**
- **Rejected**

In the future, we aim to include all **essential** features by default and have flags for the **optional** features, while **rejected** features are ignored. For more information, please check the code changes.

## Insight 📝

To know more about how we came to this agreement, please check the following threads:

- https://nimble-co.slack.com/archives/G41UCTGHL/p1629781993001700
- https://nimble-co.slack.com/archives/G41UCTGHL/p1629954695014500
- https://nimble-co.slack.com/archives/G41UCTGHL/p1642413515010600
- https://nimble-co.slack.com/archives/G41UCTGHL/p1642573191009000

## Proof Of Work 📹
Screen Shot 2565-03-22 at 11 24 25

### DIFF
--- a/docs/Home.md
+++ b/docs/Home.md
@@ -2,7 +2,7 @@ Welcome to the android-templates wiki!
 
 This repository generates a new project based on our preferences, by running a simple script. 
 
-Ex: `./newproject.sh -t crt -p co.myproject.example -n "MyProjectExampleCoroutine"`
+Example: `./newproject.sh -t crt -p co.myproject.example -n "MyProjectExampleCoroutine"`
 
 This script must include all essentials by default, while optional features can be appended with flags. To prevent this repository from becoming a dumping playground, features can be rejected too.
 
@@ -10,33 +10,31 @@ Planning to contribute? Let's have a look at the criteria:
 
 How do we decide if a feature is **essential**? 👍
 
-- Our team requires it to maintain code quality.
-- Our team requires it to maintain a good developer experience.
-- Our team requires it to increase the productivity of the team.
-- It is always used or configured in the projects we've worked on.
-- It is recommended to use or configure based on Google's standards.
+- It is always implemented in the projects we've worked on.
+- It allows developers to avoid boilerplate setup, maintain the code quality and have good experience for themselves.
+- It is a Google recommended component.
 
 How do we decide if a feature is **optional**? 🚩
 
-- It is regularly used or configured in projects we've worked on.
-- It is regularly used or configured as an alternative to the default choice.
+- It is regularly implemented in the project's we've worked on.
+- It is an alternative to a newer dominant option.
 
 How do we decide if a feature is **rejected**? 👎
 
-- It is barely used or configured in projects we've worked on.
-- It is barely used or configured as an alternative to the default choice.
+- It is barely implemented in the project's we've worked on due to being a specific project requirement.
+- It has been replaced completely by a newer dominant option.
 
 Let's have a look at some examples 🔎
 
 **Essential**:
+- [Timber](https://github.com/JakeWharton/timber): A logging library which is always used in all the projects we've worked on.
+- [Firebase App Distribution](https://firebase.google.com/docs/app-distribution): Continuous delivery always requires a boilerplate setup.
 - [Detekt](https://github.com/detekt/detekt): Code smell analysis is important to maintain code quality.
 - [Hilt](https://developer.android.com/training/dependency-injection/hilt-android): Dependency injection provides a good developer experience.
-- [Firebase App Distribution](https://firebase.google.com/docs/app-distribution): Continuous delivery increases the productivity of the team.
-- [Timber](https://github.com/JakeWharton/timber): A logging library which is always used in all the projects we've worked on.
 - [Navigation Component](https://developer.android.com/guide/navigation/navigation-getting-started): A suite of libraries for in-app navigation which is recommended by Google.
 
 **Optional**:
-- [Lottie](https://github.com/airbnb/lottie-android): Loading animations with a JSON file is regularly used, but not every project needs it.
+- [RxPermissions](https://github.com/tbruyelle/RxPermissions): Permissions powered by RxJava are regularly used, but not every project needs it.
 - [Bitrise](https://www.bitrise.io/): A CI/CD provider which is regularly used, but [GitHub Actions](https://github.com/features/actions) is our default choice.
 
 **Rejected**:
@@ -45,6 +43,6 @@ Let's have a look at some examples 🔎
 
 Keep in mind, the features are based on _our team's_ requirements. In case the client has different requirements or requests, we can consider adding them as optional features if they occur regularly.
 
-Know that the above examples are not definitive as new and existing libraries keep on emerging and evolving.
+Please note that the above examples are not definitive as new and existing libraries keep on emerging and evolving.
 
 Still unsure where your future contribution belongs? Let's discuss! 🚀
--- a/docs/Home.md
+++ b/docs/Home.md
@@ -6,34 +6,34 @@ Ex: `./newproject.sh -t crt -p co.myproject.example -n "MyProjectExampleCoroutin
 
 This script must include all essentials by default, while optional features can be appended with flags. To prevent this repository from becoming a dumping playground, features can be rejected too.
 
-Planning to contribute? Let's have a look at the criteria
+Planning to contribute? Let's have a look at the criteria:
 
 How do we decide if a feature is **essential**? 👍
 
-- Our team requires it to maintain code quality
-- Our team requires it to maintain a good developer experience
-- Our team requires it to increase the productivity of the team
-- It is always used or configured in the projects we've worked on
-- It is recommended to use or configure based on Google's standards
+- Our team requires it to maintain code quality.
+- Our team requires it to maintain a good developer experience.
+- Our team requires it to increase the productivity of the team.
+- It is always used or configured in the projects we've worked on.
+- It is recommended to use or configure based on Google's standards.
 
 How do we decide if a feature is **optional**? 🚩
 
-- It is regularly used or configured in projects we've worked on
-- It is regularly used or configured as an alternative to the default choice
+- It is regularly used or configured in projects we've worked on.
+- It is regularly used or configured as an alternative to the default choice.
 
 How do we decide if a feature is **rejected**? 👎
 
-- It is barely used or configured in projects we've worked on
-- It is barely used or configured as an alternative to the default choice
+- It is barely used or configured in projects we've worked on.
+- It is barely used or configured as an alternative to the default choice.
 
 Let's have a look at some examples 🔎
 
 **Essential**:
-- [Jacoco](https://github.com/jacoco/jacoco): Code coverage is important to maintain code quality.
+- [Detekt](https://github.com/detekt/detekt): Code smell analysis is important to maintain code quality.
 - [Hilt](https://developer.android.com/training/dependency-injection/hilt-android): Dependency injection provides a good developer experience.
 - [Firebase App Distribution](https://firebase.google.com/docs/app-distribution): Continuous delivery increases the productivity of the team.
-- [Glide](https://github.com/bumptech/glide): Image loading is always used in all the projects we've worked on.
-- [Compose](https://developer.android.com/jetpack/compose): A modern toolkit for building native UI which is recommended by Google
+- [Timber](https://github.com/JakeWharton/timber): A logging library which is always used in all the projects we've worked on.
+- [Navigation Component](https://developer.android.com/guide/navigation/navigation-getting-started): A suite of libraries for in-app navigation which is recommended by Google.
 
 **Optional**:
 - [Lottie](https://github.com/airbnb/lottie-android): Loading animations with a JSON file is regularly used, but not every project needs it.
--- a/docs/Home.md
+++ b/docs/Home.md
@@ -1,13 +1,50 @@
 Welcome to the android-templates wiki!
 
-The purpose of our android-templates repository is to quickly start a new project with all the **must-have** requirements. Other **nice-to-have** features need to be considered as an optional setting when starting a new project.
+This repository generates a new project based on our preferences, by running a simple script. 
 
-Planning to add a new feature? Ask yourself the following questions to define if it's a **must-have** requirement:
+Ex: `./newproject.sh -t crt -p co.myproject.example -n "MyProjectExampleCoroutine"`
 
-- Is the feature frequently used in other projects? (Ex: Glide, Retrofit, ...)
-- Does the feature increase productivity of the team? (Ex: Firebase, Proxyman, ...)
+This script must include all essentials by default, while optional features can be appended with flags. To prevent this repository from becoming a dumping playground, features can be rejected too.
 
-Do _both_ apply to your feature? If not, we will define it as a **nice-to-have** feature:
+Planning to contribute? Let's have a look at the criteria
 
-- Ex: [Jetpack Compose](https://developer.android.com/jetpack/compose) simplifies writing code for UI, but is not frequently used (yet)
-- Ex: [Lottie](https://lottiefiles.com/) is frequently used, but does not provide any other value unless JSON animations are needed
+How do we decide if a feature is **essential**? 👍
+
+- Our team requires it to maintain code quality
+- Our team requires it to maintain a good developer experience
+- Our team requires it to increase the productivity of the team
+- It is always used or configured in the projects we've worked on
+- It is recommended to use or configure based on Google's standards
+
+How do we decide if a feature is **optional**? 🚩
+
+- It is regularly used or configured in projects we've worked on
+- It is regularly used or configured as an alternative to the default choice
+
+How do we decide if a feature is **rejected**? 👎
+
+- It is barely used or configured in projects we've worked on
+- It is barely used or configured as an alternative to the default choice
+
+Let's have a look at some examples 🔎
+
+**Essential**:
+- [Jacoco](https://github.com/jacoco/jacoco): Code coverage is important to maintain code quality.
+- [Hilt](https://developer.android.com/training/dependency-injection/hilt-android): Dependency injection provides a good developer experience.
+- [Firebase App Distribution](https://firebase.google.com/docs/app-distribution): Continuous delivery increases the productivity of the team.
+- [Glide](https://github.com/bumptech/glide): Image loading is always used in all the projects we've worked on.
+- [Compose](https://developer.android.com/jetpack/compose): A modern toolkit for building native UI which is recommended by Google
+
+**Optional**:
+- [Lottie](https://github.com/airbnb/lottie-android): Loading animations with a JSON file is regularly used, but not every project needs it.
+- [Bitrise](https://www.bitrise.io/): A CI/CD provider which is regularly used, but [GitHub Actions](https://github.com/features/actions) is our default choice.
+
+**Rejected**:
+- [Skeleton Layout](https://github.com/Faltenreich/SkeletonLayout): A progress indicator with visual feedback, it has a specific use case that is barely used.
+- [Mockito](https://github.com/mockito/mockito): A mocking framework which is barely used, because [Mockk](https://mockk.io/) is our default choice.
+
+Keep in mind, the features are based on _our team's_ requirements. In case the client has different requirements or requests, we can consider adding them as optional features if they occur regularly.
+
+Know that the above examples are not definitive as new and existing libraries keep on emerging and evolving.
+
+Still unsure where your future contribution belongs? Let's discuss! 🚀
--- a/docs/Home.md
+++ b/docs/Home.md
@@ -1,3 +1,13 @@
-Welcome to the android-templates wiki! 
+Welcome to the android-templates wiki!
 
-More coming soon...
+The purpose of our android-templates repository is to quickly start a new project with all the **must-have** requirements. Other **nice-to-have** features need to be considered as an optional setting when starting a new project.
+
+Planning to add a new feature? Ask yourself the following questions to define if it's a **must-have** requirement:
+
+- Is the feature frequently used in other projects? (Ex: Glide, Retrofit, ...)
+- Does the feature increase productivity of the team? (Ex: Firebase, Proxyman, ...)
+
+Do _both_ apply to your feature? If not, we will define it as a **nice-to-have** feature:
+
+- Ex: [Jetpack Compose](https://developer.android.com/jetpack/compose) simplifies writing code for UI, but is not frequently used (yet)
+- Ex: [Lottie](https://lottiefiles.com/) is frequently used, but does not provide any other value unless JSON animations are needed


Tuubz:
Next step, code integration part for:

-  Add optional flags to the generate script command
-  Add/remove nice-to-have features to the code

I'll create an issue for it tomorrow 💯

docs/Home.md
```diff
@@ -1,3 +1,13 @@
...
+The purpose of our android-templates repository is to quickly start a new project with all the **must-have** requirements. Other **nice-to-have** features need to be considered as an optional setting when starting a new project.
+
+Planning to add a new feature? Ask yourself the following questions to define if it's a **must-have** requirement:
+
+- Is the feature frequently used in other projects? (Ex: Glide, Retrofit, ...)
+- Does the feature increase productivity of the team? (Ex: Firebase, Proxyman, ...)
+
+Do _both_ apply to your feature? If not, we will define it as a **nice-to-have** feature:
+
+- Ex: [Jetpack Compose](https://developer.android.com/jetpack/compose) simplifies writing code for UI, but is not frequently used (yet)
```
sleepylee, line 12: it doesn't simplify, it is "a different way to write UI programmatically". The way you interpreted it makes it sounds like a nice-to-have not by itself 😅

docs/Home.md
```diff
@@ -1,3 +1,13 @@
-Welcome to the android-templates wiki! 
+Welcome to the android-templates wiki!
 
-More coming soon...
\ No newline at end of file
+The purpose of our android-templates repository is to quickly start a new project with all the **must-have** requirements. Other **nice-to-have** features need to be considered as an optional setting when starting a new project.
+
+Planning to add a new feature? Ask yourself the following questions to define if it's a **must-have** requirement:
+
+- Is the feature frequently used in other projects? (Ex: Glide, Retrofit, ...)
```
sleepylee, line 7: frequently used or having the side setup that is usually verbose to replicate is also considerable.

docs/Home.md
```diff
@@ -1,3 +1,13 @@
...
+- Does the feature increase productivity of the team? (Ex: Firebase, Proxyman, ...)
```
sleepylee, line 8: Firebase is huge, and if it's about Analysis we don't always have it.
Maybe you are mentioning Firebase Distribution or Crashlytics? Let's be explicit

sleepylee, line 8: Firebase is huge, and if it's about Analysis we don't always have it.
Maybe you are mentioning Firebase Distribution or Crashlytics? Let's be explicit

docs/Home.md
```diff
@@ -1,3 +1,13 @@
...
+
+Do _both_ apply to your feature? If not, we will define it as a **nice-to-have** feature:
+
+- Ex: [Jetpack Compose](https://developer.android.com/jetpack/compose) simplifies writing code for UI, but is not frequently used (yet)
+- Ex: [Lottie](https://lottiefiles.com/) is frequently used, but does not provide any other value unless JSON animations are needed
```
sleepylee, line 13: I think we can interpret it a bit more direct and less political here (it is this, but not...unless...)

How about

```suggestion
- Ex: [Lottie](https://lottiefiles.com/) a library for animation exclusively with a specific json format file. It has a specific use case and requirements that not every project is having.
```

docs/Home.md
Tuubz: > Jetpack Compose is Android's modern toolkit for building native UI. It simplifies and accelerates UI development on Android bringing your apps to life with less code, powerful tools, and intuitive Kotlin APIs

I got it from the official documentation actually 😅 But I'll find another example anyways, as in the future it will be used more frequently in our projects 🚀

docs/Home.md
Tuubz: >  having the side setup that is usually verbose to replicate

Could you give an example of this, please? 🙇

Tuubz:
@Wadeewee I'll re-think about what examples to give, thanks for the suggestions 👍

docs/Home.md
sleepylee: for example OkHttp we usually have to configure:
- Debug level
- Timeout for read/write
- Interceptor
...

Tuubz: Then I would say they are frequently used. Debug level, time-outs, and interceptors are frequently used in other projects. So I think we keep it like it is, to keep it minimized. Instead, I could mention it as one of the examples for a must-have feature (with some more explanation, like we did for Lottie). What do you think? 🙏

sleepylee: well...it's also good if you can specify on the setup as well; frequently used setups could be different from frequently used lib - as a lib could have many features that we don't use, so my point here is also to welcome people open PR to make configuration changes that we can make used of too...not only about adding a new lib should be considered as a must-have contribution.

Tuubz: What about if we rephrase it like this?

>  Is the feature frequently used/configured in other projects?

Tuubz: Or we need to avoid using "feature" as it might not clearly refer to configurations as you mentioned 🤔

Tuubz:
> @Tuubz It's a good idea but I'm not sure about `must-have` sometime it depend on client which is they want us to implement or depend on type of project as well because some project may be don't need API so we no need to use `Retrofit`, I feel like all of your sample are `nice-to-have`, I think for `must-have` it should be generic that not depend on feature something like these `kotlin`, `mocck`, `kotest`, `timber` What do you think? 🤔

@Wadeewee Originally, I agreed with your statement, but [after some discussions](https://nimble-co.slack.com/archives/G41UCTGHL/p1642573191009000) I've changed my mind 🙏 

It's true that the project depends on the client's requirements, but most of the time, they leave it up to us to decide which technologies to use for a new project, which we believe are best for the team and project 🚀  In terms of the Retrofit example, I still think it's an essential part. Correct me if I'm wrong, but I think it's unlikely that a project will not require an API. 🙇  

Anyhow, I've updated the content. Please have another look when available 🤩

Tuubz:
> @Tuubz It's a good idea but I'm not sure about must-have sometime it depend on client which is they want us to implement or depend on type of project as well because some project may be don't need API so we no need to use Retrofit, I feel like all of your sample are nice-to-have, I think for must-have it should be generic that not depend on feature something like these kotlin, mocck, kotest, timber What do you think? 🤔

@Wadeewee Originally, I agreed with your statement, but after some discussions I've changed my mind 🙏

It's true that the project depends on the client's requirements, but most of the time, they leave it up to us to decide which technologies to use for a new project, which we believe are best for the team and project 🚀 In terms of the Retrofit example, I still think it's an essential part. Correct me if I'm wrong, but I think it's unlikely that a project will not require an API. 🙇

Anyhow, I've updated the content. Please have another look when available 🤩

docs/Home.md
```diff
@@ -1,3 +1,50 @@
...
+- It is barely used or configured in projects we've worked on
+- It is barely used or configured as an alternative to the default choice
+
+Let's have a look at some examples 🔎
+
+**Essential**:
+- [Jacoco](https://github.com/jacoco/jacoco): Code coverage is important to maintain code quality.
+- [Hilt](https://developer.android.com/training/dependency-injection/hilt-android): Dependency injection provides a good developer experience.
+- [Firebase App Distribution](https://firebase.google.com/docs/app-distribution): Continuous delivery increases the productivity of the team.
+- [Glide](https://github.com/bumptech/glide): Image loading is always used in all the projects we've worked on.
```
Wadeewee, line 35: @Tuubz I think guide it's an `optional` we can add it later when we have to implement feature with `image/video`  🤔 

But I think would it be good if we add `RxJava and Coroutine` in `essential` as we know we must to choose one of both before create project and what about `navigation` component, Is it essential too like `Compose` or `Hilt` What do you think 🙏

docs/Home.md
```diff
@@ -1,3 +1,50 @@
...
 
-More coming soon...
\ No newline at end of file
+This repository generates a new project based on our preferences, by running a simple script. 
+
+Ex: `./newproject.sh -t crt -p co.myproject.example -n "MyProjectExampleCoroutine"`
+
+This script must include all essentials by default, while optional features can be appended with flags. To prevent this repository from becoming a dumping playground, features can be rejected too.
+
+Planning to contribute? Let's have a look at the criteria
```
lydiasama, line 9:
```suggestion
Planning to contribute? Let's have a look at the criteria:
```

docs/Home.md
```diff
@@ -1,3 +1,50 @@
...
+
+How do we decide if a feature is **essential**? 👍
+
+- Our team requires it to maintain code quality
+- Our team requires it to maintain a good developer experience
+- Our team requires it to increase the productivity of the team
+- It is always used or configured in the projects we've worked on
+- It is recommended to use or configure based on Google's standards
```
lydiasama, line 17: Shall we add a period `.` after each sentence? 🤔  Same to all bullets below.

docs/Home.md
```diff
@@ -1,3 +1,50 @@
...
+**Essential**:
+- [Jacoco](https://github.com/jacoco/jacoco): Code coverage is important to maintain code quality.
+- [Hilt](https://developer.android.com/training/dependency-injection/hilt-android): Dependency injection provides a good developer experience.
+- [Firebase App Distribution](https://firebase.google.com/docs/app-distribution): Continuous delivery increases the productivity of the team.
+- [Glide](https://github.com/bumptech/glide): Image loading is always used in all the projects we've worked on.
+- [Compose](https://developer.android.com/jetpack/compose): A modern toolkit for building native UI which is recommended by Google
+
+**Optional**:
+- [Lottie](https://github.com/airbnb/lottie-android): Loading animations with a JSON file is regularly used, but not every project needs it.
+- [Bitrise](https://www.bitrise.io/): A CI/CD provider which is regularly used, but [GitHub Actions](https://github.com/features/actions) is our default choice.
```
lydiasama, line 40: I wonder about our default choice, do we include the default choice to the `essential` too?

docs/Home.md
```diff
@@ -1,3 +1,50 @@
...
+
+- It is barely used or configured in projects we've worked on
+- It is barely used or configured as an alternative to the default choice
+
+Let's have a look at some examples 🔎
+
+**Essential**:
+- [Jacoco](https://github.com/jacoco/jacoco): Code coverage is important to maintain code quality.
+- [Hilt](https://developer.android.com/training/dependency-injection/hilt-android): Dependency injection provides a good developer experience.
+- [Firebase App Distribution](https://firebase.google.com/docs/app-distribution): Continuous delivery increases the productivity of the team.
```
lydiasama, line 34: I assume that the Firebase integration is our default choice, right? 🤔  Have you faced any projects that don't use Firebase so far? @Tuubz

docs/Home.md
Tuubz: @Wadeewee Thank you for your feedback 🚀 

> I think guide it's an optional we can add it later when we have to implement feature with image/video

In my opinion, I think `Glide` is `essential`. Because it is always used in all our previous projects. I think it's very unlikely that a future project will not have any image loading. Please correct me if I'm wrong 🙇 

> But I think would it be good if we add RxJava and Coroutine in essential as we know we must to choose one of both before create project

This is a bit of a special case, RxJava should actually be re-factored to be `optional`. While `Coroutine` should be `essential` and set as our default, as it's recommended by Google. What do you think?

> what about navigation component, Is it essential too like Compose or Hilt

Yes, the navigation component is also essential. This is because it is recommended by Google's standards 🙏

docs/Home.md
Tuubz: No, we don't need to add the default choice to `essential`, because all the essentials are our default already 🙏 

I've added the default choice here as an example, to prove that Bitrise is an `optional`

docs/Home.md
Tuubz: Correct, `Firebase App Distribution` is our default choice 🚀 

In my experience, the client for 7-11 chat project wanted to get the app distributed by e-mail (with help of Bitrise)

That's the only reference I have 😁

docs/Home.md
Wadeewee: >This is a bit of a special case, RxJava should actually be re-factored to be optional. While Coroutine + LiveData (or Flow) should be essential and set as our default, as it's recommended by Google. What do you think?
>
>Yes, navigation component is also essential. This is because it is recommended by Google's standards 

@Tuubz I agreed with this so let add `Coroutine + Flow` and  `navigation ` in `essential` and  add  `RxJava`  in `optional` but for `LiveData` I think it old and we don't use it in `CoroutineTemplate` 🙏  and one more thing, Would it be great if we reference the example by the dependencies that we're using in our both template 🙇‍♀️

Tuubz: > I agreed with this so let add Coroutine + Flow and navigation in essential and add RxJava in optional but for LiveData I think it old and we don't use it in CoroutineTemplate 🙏

For essential features, there are five criteria, so my idea is to give _one_ example for each criterion:

- I could replace the `Compose` example with `Navigation Component`, would that be better?
- Since `Coroutine` & `RxJava` are special cases and will be refactored in the future, I think it's better to avoid these examples (for now), to avoid confusion. What do you think? 🙏 

> Would it be great if we reference the example by the dependencies that we're using in our both template

Ex: Are you suggesting we should avoid using `Jacoco` as an example, because it's currently only used as a dependency for `RxJava`, but not for `Coroutine`? 🙇

Wadeewee: > I could replace the Compose example with Navigation Component, would that be better?
Since Coroutine & RxJava are special cases and should be refactored in the future, I think it's better to avoid these examples (for now), to avoid confusion. What do you think? 🙏

@Tuubz Could you please explain more bit about `Since Coroutine & RxJava are special cases and should be refactored in the future` Didn't get it much 🙏 

> Ex: Are you suggesting we should avoid using Jacoco as an example, because it's currently only used in RxJava, but not for Coroutine? 🙇 I guess we could replace this example with Detekt instead 🚀
>

@Tuubz Sound good to me from now we can consider to add `Jacoco` later 🚀

Wadeewee: Ah, I see @Tuubz thanks for explanation 🙇‍♀️

docs/Home.md
```diff
@@ -1,3 +1,50 @@
...
+- It is barely used or configured as an alternative to the default choice.
+
+Let's have a look at some examples 🔎
+
+**Essential**:
+- [Detekt](https://github.com/detekt/detekt): Code smell analysis is important to maintain code quality.
+- [Hilt](https://developer.android.com/training/dependency-injection/hilt-android): Dependency injection provides a good developer experience.
+- [Firebase App Distribution](https://firebase.google.com/docs/app-distribution): Continuous delivery increases the productivity of the team.
+- [Timber](https://github.com/JakeWharton/timber): A logging library which is always used in all the projects we've worked on.
+- [Navigation Component](https://developer.android.com/guide/navigation/navigation-getting-started): A suite of libraries for in-app navigation which is recommended by Google.
```
doannimble, line 36: @Tuubz I think `Navigation Component` should be `Optional`, as not all projects need it 🤔

docs/Home.md
doannimble: @Tuubz Imo, it would be more understandable if we separate them into: 

- `Must Have` - which cannot be missing and or replaceable.

- `Nice to have` - which can be replaceable, but so far, it is our best choice.

- `Optional` - which can be replaced by other ones anytime.

docs/Home.md
sleepylee: I think of the opposite...if it is optional then you will have to provide the primary way to navigate in an application...I don't think the traditional fragment backstack manipulation is preferred right?

Tuubz: @doannimble Could you perhaps give an example of which project didn't need it?

In my opinion, anything that Google recommends using should be `Essential`. So as they recommend using "[Single Activity with multiple fragments](https://youtu.be/9O1D_Ytk0xg?t=79)", navigation component comes with.

sleepylee:
@doannimble this proposal has a few flaws that you are leaving for further confusion:

> Must Have - which cannot be missing and or replaceable.

there is nothing `not replaceable` forever...so it's either the only option for now (navigation component), or there are 2 ways to do things but 1 must be the default (coroutine over rxjava).

> Nice to have - which can be replaceable, but so far, it is our best choice.

Now when you read again, you actually reference ` but so far, it is our best choice.` to the first choice

> Optional - which can be replaced by other ones anytime.

and lastly, this same repetition `which can be replaced` will create confusion to categorize type with type 2. 


Please understand the idea first, having an opinion and cross-check on your understanding before proposing a new idea.

docs/Home.md
Tuubz: @doannimble Please have a look at [this thread](https://nimble-co.slack.com/archives/G41UCTGHL/p1642578956011700?thread_ts=1642573191.009000&cid=G41UCTGHL) where we discuss the naming, hopefully, it can clear up the naming choice  🙏

docs/Home.md
```diff
@@ -1,3 +1,50 @@
...
+
+Planning to contribute? Let's have a look at the criteria:
+
+How do we decide if a feature is **essential**? 👍
+
+- Our team requires it to maintain code quality.
+- Our team requires it to maintain a good developer experience.
+- Our team requires it to increase the productivity of the team.
+- It is always used or configured in the projects we've worked on.
+- It is recommended to use or configure based on Google's standards.
```
sleepylee, line 17: I think we can reorder this a bit so it has a parallelism structure with the other option below. What I meant is that the first consideration must be about `how often it has been chosen or agreed to be picked` in a project.

The first 3 can be shortened to reduce the repetition in the writing, be precise.
>- Our team requires it to maintain code quality.
>- Our team requires it to maintain a good developer experience.
>- Our team requires it to increase the productivity of the team.

lastly, suggesting a more directive way of writing:
> It is a Google recommended component.

docs/Home.md
```diff
@@ -1,3 +1,50 @@
...
+
+How do we decide if a feature is **optional**? 🚩
+
+- It is regularly used or configured in projects we've worked on.
+- It is regularly used or configured as an alternative to the default choice.
```
sleepylee, line 22: recommend avoiding repetition. This can be understood simply as: 
- We used it too, but maybe now there is a newer dominant option (coroutine over rxjava so rxjava becomes optional).

docs/Home.md
```diff
@@ -1,3 +1,50 @@
...
+
+How do we decide if a feature is **rejected**? 👎
+
+- It is barely used or configured in projects we've worked on.
+- It is barely used or configured as an alternative to the default choice.
```
sleepylee, line 27: same here, we need to be more directive and succinct: Components that we don't apply to most of the project, come as a specific project requirement. Right? 🤔

sleepylee: and the idea here is to unify the ways/components our team should be familiar with and spend effort on maintaining, so by rejecting the uncommon idea, we can have more focus on the real matter.

docs/Home.md
doannimble: @sleepylee @Tuubz  We can use fragment backstack with activity if it works and does not give us an issue. You guys can see at the end of the video [5:30](https://www.youtube.com/watch?v=9O1D_Ytk0xg&t=330s)  that if you're using Multi-Activity development and it works, stick with it. 

But if you're running into issues and bugs then to use Single Activity development.

The old way is not to be preferred, but it does not mean the `Navigation Component` is a must-have IMO 🙏

sleepylee: we cannot leave the place for doubt like that... please read the description again on the essential part: the feature targetted as default or essential is the one that increase our dev experience and productivity. This is not a playground or fully crowd open source that we have to leave space for interpretation. We must own it and make the solid rule in the end.
As you can see, every project is now moving to navigation component; will you start the next project without it? Just answer that question to yourself then you might have the clear answer.

lydiasama: @doannimble As for the android-template is for `us` to use it to initiate the android project that we're gonna work on it further. So I think this should combine the approach we usually use in the client project. 🙏🏻  So I think the `Navigation Component` is an Essential one. What do you think?

docs/Home.md
Tuubz: @sleepylee Yes, there is some repetition: `Our team requires it to...` 

Do you suggest minimizing it in one sentence? I'm open to hearing your approach 🙏

sleepylee: yes, I'd say we write it in a crystal-clear thought:
something like:


> - It is always implemented in the projects we have worked on or will work on.
> - It allows developers to increase the productivity from boilerplate setup, maintain the quality, and have a good experience for themselves.
>- It is a Google-recommended component.

what do you think?

Tuubz: Thanks for the recommendations, I agree with the first and third bullet points.

For the second bullet point, we can't only refer to `developers`. 

Ex: `Chucker` or `Firebase App Distribution` also increases the productivity of our project managers.

sleepylee: developers have the most benefit from it, in this codebase PM comes after so I would refrain to prioritize them in this topic, it will generate more conflict to resolve later when someone wants to bring something that has less value to dev but more value to the PM (btw, Product Manager, we don't have Project Manager role), how do you think?

Tuubz: If I'm not mistaken, it was Julien that requested for Chucker.

As for Firebase App Distribution, is mostly beneficial for testers (our PM).

Please correct me if I got the wrong idea here 🙏

sleepylee: nope, it was from our side to initialize Chucker after we learned from iOS. Eventually, we found it very useful for dev ourselves as well.

For Firebase App Distribution, it is to enhance our productivity because it's a requirement that often appears in our process, so we want to reduce the setup time. Again, the purpose of the feature is for everyone, the template is here to serve us: save our time on the setup.

Tuubz: I agree, that it saves _our_ time on the setup. But both features were implemented for the team's purpose:

- We can check LogCat, so Chucker is not essential for us developers. It is essential for PMs however
- PMs require Firebase App Distribution for testing. So it's for their needs that we set it up 

🙏

sleepylee: I still don't get your point from the latest argument. As been said, I don't deny that PM benefits from it, but I suggested to not involving them in this context because it can generate more debates in the future.

Now talking about Firebase or Chucker: 
- Do developers use them? Yes.
- Do we want to save time to set them up? Yes

So can we just talk about developers in here? I think yes too?

sleepylee: > We can check LogCat, so Chucker is not essential for us developers. It is essential for PMs however.

LogCat is a console display, while Chucker allows you to debug while not connecting to AS. I guess you haven't had this experience before in AllOnline but we did have some good use cases with it in DM where we started using it more often. While it is not a blocker for our development, it has proved its usefulness to us, as well as the PM, so that's why it got promoted here to reduce the setup time.

sleepylee: For firebase: to shape the understanding of "developer-oriented" way: it is a requirement that often come from the PM; while it benefits the PMs, we don't necessarily bring PM here as the reason. The real reason or the real benefit **to the developers** is that we don't have to re-setup it, again and again, all the time. Does this sound different from your original thought? 🤔

docs/Home.md
doannimble: Got it, thank you @Tuubz @sleepylee @lydiasama 🙏

docs/Home.md
```diff
@@ -1,3 +1,50 @@
...
+- [Lottie](https://github.com/airbnb/lottie-android): Loading animations with a JSON file is regularly used, but not every project needs it.
+- [Bitrise](https://www.bitrise.io/): A CI/CD provider which is regularly used, but [GitHub Actions](https://github.com/features/actions) is our default choice.
+
+**Rejected**:
+- [Skeleton Layout](https://github.com/Faltenreich/SkeletonLayout): A progress indicator with visual feedback, it has a specific use case that is barely used.
+- [Mockito](https://github.com/mockito/mockito): A mocking framework which is barely used, because [Mockk](https://mockk.io/) is our default choice.
+
+Keep in mind, the features are based on _our team's_ requirements. In case the client has different requirements or requests, we can consider adding them as optional features if they occur regularly.
+
+Know that the above examples are not definitive as new and existing libraries keep on emerging and evolving.
```
minhnimble, line 48: Using the word `Know that` here doesn't seem to be appropriate for the given context, I suggest to update it to `Please note that` for better wording on politely letting the readers be aware of what is going on. What do you think?
```suggestion
Please note that the above examples are not definitive as new and existing libraries keep on emerging and evolving.
```

Tuubz:
@minhnimble @lydiasama @Wadeewee @sleepylee Updated 🚀

minhnimble (approved):
LGTM 🚀

docs/Home.md
```diff
@@ -1,3 +1,48 @@
...
+
+**Essential**:
+- [Timber](https://github.com/JakeWharton/timber): A logging library which is always used in all the projects we've worked on.
+- [Firebase App Distribution](https://firebase.google.com/docs/app-distribution): Continuous delivery always requires a boilerplate setup.
+- [Detekt](https://github.com/detekt/detekt): Code smell analysis is important to maintain code quality.
+- [Hilt](https://developer.android.com/training/dependency-injection/hilt-android): Dependency injection provides a good developer experience.
+- [Navigation Component](https://developer.android.com/guide/navigation/navigation-getting-started): A suite of libraries for in-app navigation which is recommended by Google.
+
+**Optional**:
+- [Lottie](https://github.com/airbnb/lottie-android): Loading animations with a JSON file is regularly used, but not every project needs it.
```
luongvo, line 37: This could be considered as `Rejected`, the same as `Skeleton Layout` too IMO. Instead of mentioning `Lottie` as `Optional`, I think we can list `Glide - an image loader` here because almost project needs an image loader, but not always `Glide` 🤓

luongvo (commented):
Rest lgtm

docs/Home.md
sleepylee: >I think we can list Glide - an image loader here because almost project needs an image loader, but not always Glide 

@luongvo do you mean

>....but not always `Lottie`
?

because 10/10 recent projects we had we always use Glide 🤔

luongvo: @sleepylee Okay if 10/10 recent projects we had we always use Glide so Glide is an essential feature, I may be wrong on this part.

On another side, I still don't think we can consider `Lottie` as Optional, it should be the same as `Skeleton Layout` in example.

Tuubz: @luongvo At this moment, I think I have used `Lottie` in at least of half of my projects, while for `Skeleton` it's zero.

Anyways, I'm totally open to finding a more clear example. Do you have any suggestion? 🙇

sleepylee: @luongvo yep probably we can reject it too as Lottie's setup is relatively simple with a library import. Possibly for those cases, depending on the project's need - developers should setup themselves instead of us maintaining it.

sleepylee: @Tuubz I think we have Skeleton loading view in The 1 + DeeMoney (not sure, maybe Toyota Wallet also) 🤔

Tuubz: I'll revise the rules between "optional" & "rejected" a bit more, so it's more clear and strict + adapt the examples accordingly 👍

luongvo: @Tuubz @sleepylee How about listing [RxPermissions](https://github.com/tbruyelle/RxPermissions) as an Optional feature here instead? I see most of our projects use it and it's quite convenience for our permission handling steps too 🤔

sleepylee: @luongvo that works 👍

Tuubz:
@luongvo @sleepylee @lydiasama @Wadeewee Updated 🤩

luongvo (approved):
All good, LGTM

lydiasama (approved):

Wadeewee (approved):

Tuubz:
@doannimble @sleepylee Any final words? Please do, if any 😁

doannimble (approved):
LGTM 🚀

Issue #157, opened 2022-02-01
Fix resources to minimize `Danger` noise


Tuubz:
## Why

We use `Danger` as a code quality tool, [here's how we integrated it](https://github.com/nimblehq/android-templates/pull/129).

Currently, it's creating a lot of noise whenever we create a pull request. So let's start by resolving the `resources` warnings first:

- `Hardcoded string "Next", should use @string resource`
- `Hardcoded string "Jetpack Compose", should use @string resource`
- `The resource R.font.font_circular_family appears to be unused`
- `Hardcoded string "Refresh", should use @string resource`
- `Hardcoded string "Clickable text", should use @string resource`
- `Hardcoded string "Open Camera", should use @string resource`
- `Hardcoded string "Open Post", should use @string resource`
- `The resource R.color.blue_free_speech_90a appears to be unused`
- `The resource R.color.red_violet_70a appears to be unused`
- `The resource R.color.white appears to be unused`
- `The resource R.color.black appears to be unused`
- `The resource R.style.Widget_Template_TextView_Bold appears to be unused`
- `The resource R.string.api_endpoint_example appears to be unused`

I'll create more issues like this, once this issue has been approved. Because as you can see [here](https://github.com/nimblehq/android-templates/pull/156#issuecomment-1026611148), there's a lot more to fix 🙏 

## Who Benefits?

Less noise is better to observe the warnings in our pull requests (I personally skip and scroll the Danger section 😅). So once we resolve all the current Danger warnings, we can create a good habit of only merging future pull requests once there are [no Danger warnings left ](https://github.com/nimblehq/android-templates/pull/156#issuecomment-1026611148) 🚀

PR #164, opened 2022-03-23
[#157] Resolve resource warnings from Danger

Tuubz:
#157 

## What happened 👀

Resolve **resource** warnings from Danger to have better visibility on code smells. As you can see, the amount of warnings have been minimized, we'll tackle the rest in a different ticket (except code coverage)

## Proof Of Work 📹

Check Danger warnings below, you can compare it with **[this pull request](https://github.com/nimblehq/android-templates/pull/167)** for example


doannimble (approved):

CoroutineTemplate/app/src/main/res/layout/fragment_home.xml
```diff
@@ -12,7 +13,8 @@
         android:text="Next"
         app:layout_constraintEnd_toEndOf="parent"
         app:layout_constraintStart_toStartOf="parent"
-        app:layout_constraintTop_toTopOf="parent" />
+        app:layout_constraintTop_toTopOf="parent"
+        tools:ignore="HardcodedText" />
```
Wadeewee, line 17: Why we don't add string in `strings.xml` instead of ignoring `HardcodedText` 🙏

Tuubz: I'll add it to `strings.xml` 🚀

luongvo: I don't think we should try to add those example detail stuff into config like strings or dimension. We should try to keep our Template project neat, clean, and flexible by the `setup` script. 🤔 

FYI: this is the same issue we're discussing when building Flutter Template https://github.com/nimblehq/flutter_templates/issues/58 🤓

luongvo (approved):

CoroutineTemplate/app/src/main/res/layout/fragment_home.xml
Tuubz: We're also planning to restructure the folders in this repository, it would be nice if both flutter and android templates could agree on the same implementation. Most likely this discussion will not be concluded this month, I'll need some more time to think about it too.

@Wadeewee Would it be OK to keep the changes for now? Once we've agreed on how to restructure the project, we can have another look into separating the strings 🙏

Wadeewee: > Would it be OK to keep the changes for now? Once we've agreed on how to restructure the project, we can have another look into separating the strings

@Tuubz Sound good to me 🚀

Wadeewee (approved):

minhnimble (approved):

Issue #179, opened 2022-04-04
Fix: incorrect Jacoco coverage report after migrating to Kotlin DSL


luongvo:
## Issue

After migrating [Jacoco config to Kotlin DSL](https://github.com/nimblehq/android-templates/pull/145), Jacoco is reporting **incorrect** coverage with some redundant packages and classes.

![image](https://user-images.githubusercontent.com/16315358/161608555-ddce7a2d-8d65-49c2-bb84-5baa98866dc7.png)
 
## Expected

- These packages should be excluded from the config
```
    "co/nimblehq/di/**",
    "co/nimblehq/ui/**/di/**",
    "com/bumptech/glide"
```

- There are also some more packages and classes should be excluded

```
dagger.hilt.internal.aggregatedroot.codegen
dagger.hilt.internal.processedrootsentinel.codegen
hilt_aggregated_deps
```

![image](https://user-images.githubusercontent.com/16315358/161608923-550c61d9-d65e-4450-b136-551a7ec853fb.png)

![image](https://user-images.githubusercontent.com/16315358/161609222-751d255a-5c07-4235-a468-9873215cafde.png)

![image](https://user-images.githubusercontent.com/16315358/161609392-072d8e30-de1d-4245-a2db-6da4cfa0a72b.png)

![image](https://user-images.githubusercontent.com/16315358/161609558-79880e98-5508-47eb-833b-e0ffc6c0d2b7.png)

 
## Steps to reproduce

1. Checkout develop at hash `cdd9b100dbe7a458e0bacc3c5897638ba0ff667b`.
2. Run `./gradlew jacocoTestReport`.
3. Open HTML report.

luongvo:
@Tuubz This bug should be added and fixed in 3.7.0 before https://github.com/nimblehq/android-templates/issues/169 🤓

PR #145, opened 2021-11-04
[111] Gradle scripts with Kotlin DSL - Part 3 (Jacoco and Checkstyle gradle)

hoangnguyen92dn:
https://github.com/nimblehq/android-templates/issues/111

## What happened 👀

Kotlin DSL is now supported to handle Gradle scripts with Kotlin syntax. It's more familiar than Groovy-based syntax for Android developers.

-> Convert some common Groovy-based Gradle scripts into Kotlin DSL scripts.
 
## Insight 📝

- Migrate `jacoco.gradle` & `codequality.gradle` to `.kts`

 
## Proof Of Work 📹

🚀 CI build should be passed ✅

PR #180, opened 2022-04-04
[#179] Fix: incorrect Jacoco coverage report after migrating to Kotlin DSL

luongvo:
https://github.com/nimblehq/android-templates/issues/179

## What happened 👀

After migrating [Jacoco config to Kotlin DSL](https://github.com/nimblehq/android-templates/pull/145), Jacoco is reporting **incorrect** coverage with some redundant packages and classes.

## Insight 📝

- In Kotlin DSL, we need to add `**/` prefix for package excludes.
- There are also some more packages and classes that should be excluded, such as Hilt, Nav Component, Data binding, Kotlin enum, etc.

```
dagger.hilt.internal.aggregatedroot.codegen
dagger.hilt.internal.processedrootsentinel.codegen
hilt_aggregated_deps
```

## Proof Of Work 📹

|  After | Before |
| - | - |
| image | image | 
| image | image | 
| image | image | 
| image | image | 
| image | image | 
| image | image |

### DIFF
--- a/RxJavaTemplate/.gitignore
+++ b/RxJavaTemplate/.gitignore
@@ -55,3 +55,6 @@ freeline_project_description.json
 
 # Fastlane
 fastlane/report.xml
+
+# Code coverage
+jacoco.exec
--- a/RxJavaTemplate/app/.gitignore
+++ b/RxJavaTemplate/app/.gitignore
@@ -1,2 +1,1 @@
 /build
-jacoco.exec
--- a/RxJavaTemplate/config/jacoco.gradle.kts
+++ b/RxJavaTemplate/config/jacoco.gradle.kts
@@ -11,12 +11,28 @@ val fileGenerated = setOf(
     "**/*_ViewBinding*.*",
     "**/*Adapter*.*",
     "**/*Test*.*",
-    "android/**/*.*"
+    // Enum
+    "**/*\$Creator*",
+    // Nav Component
+    "**/*_Factory*",
+    "**/*FragmentArgs*",
+    "**/*FragmentDirections*",
+    "**/FragmentNavArgsLazy.kt",
+    "**/*Fragment*navArgs*",
+    "**/*ModuleDeps*.*",
+    "**/*NavGraphDirections*",
+    // Hilt
+    "**/*_HiltComponents*",
+    "**/*_HiltModules*",
+    "**/Hilt_*"
 )
 
 val packagesExcluded = setOf(
-    "**/co/nimblehq/rxjava/di/**",
-    "**/com/bumptech/glide"
+    "**/com/bumptech/glide",
+    "**/dagger/hilt/internal",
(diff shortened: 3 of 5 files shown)


Tuubz:
@luongvo Is it possible for us to exclude these as well? 💭 

Screen Shot 2565-04-05 at 10 12 02

luongvo:
> @luongvo Is it possible for us to exclude these as well? 💭
> 
> Screen Shot 2565-04-05 at 10 12 02

@Tuubz Actually, they are our written code, it's just the way that Jacoco tries to show code coverage for Kotlin Lamba funcs. I don't see a good reason to remove it. 🤔

Tuubz:
> > @luongvo Is it possible for us to exclude these as well? 💭
> > Screen Shot 2565-04-05 at 10 12 02
> 
> @Tuubz Actually, they are our written code, it's just the way that Jacoco tries to show code coverage for Kotlin Lamba funcs. I don't see a good reason to remove it. 🤔

I could be wrong, but the ones that I highlighted in the screenshot are reflected already in `HomeFragment.kt`. 

So it feels a bit redundant, what do you think?

luongvo:
> > > @luongvo Is it possible for us to exclude these as well? 💭
> > > Screen Shot 2565-04-05 at 10 12 02
> > 
> > 
> > @Tuubz Actually, they are our written code, it's just the way that Jacoco tries to show code coverage for Kotlin Lamba funcs. I don't see a good reason to remove it. 🤔
> 
> I could be wrong, but the ones that I highlighted in the screenshot are reflected already in `HomeFragment.kt`.
> 
> So it feels a bit redundant, what do you think?

@Tuubz I see, but because lambda funcs are parts of Kotlin language so we can't exclude lambda function out of the coverage report. That's the support of Jacoco for Kotlin lang, not an issue I think. You can check out more at https://www.jacoco.org/jacoco/trunk/doc/changes.html

Tuubz (approved):
@luongvo I still have a strong urge to remove them from the report (especially because some of them show `Source file "xxx.kt" was not found during generation of report.` once you click on them.)

Example:
Screen Shot 2565-04-05 at 11 01 41

Anyways, your code changes are definitely an improvement from before, so approved 🚀

Maybe it's something we can look in to again later

luongvo:
@Tuubz Thank you, I tried to focus on excluding generated files that affect our coverage statistics rather than excluding some coverage detail that does not affect the coverage of container files. 🤓 

For example:
- DI and Hilt generated files > no need to test, should be excluded to increase the main coverage of our actual written code.
- Lamba and inline functions > need to be tested, should not be excluded from the main coverage (except we can find a way to hide it without changing coverage statistics).

Wadeewee (approved):

Tuubz:
@nimblehq/android-chapter Could we get one more approval here, please? 🤩

doannimble (approved):

Issue #171, opened 2022-03-25
Remove `checkStyle` 


Tuubz:
## Why

We can remove [checkStyle](https://github.com/checkstyle/checkstyle) because it can only be used for Java and has been replaced completely by [detekt](https://github.com/detekt/detekt). 

So let's remove anything related to `checkStyle` accordingly 🎸 

## Who Benefits?

Developers will have a cleaner codebase with less chance for confusion

I think this is **essential**, because it helps with maintaining code quality

luongvo:
@Tuubz This one is optional imo since it's integrated without no issues, removing it will take our effort.

Tuubz:
@luongvo Hmmm, if I'm not mistaken we're not using it anymore (we don't have any Java classes left in our code)

Wouldn't you say it would be essential to have it removed (to avoid confusion and to maintain clean code) 🙏 

If we decide it to be optional, it would mean that we could add a flag to our script to include `checkStyle` 🤔

luongvo:
> If we decide it to be optional, it would mean that we could add a flag to our script to include checkStyle 🤔

@Tuubz Hmm, I forgot this point, so I agree to set this task to be `essential`, we should remove `checkStyle` but it's not in rush.

PR #189, opened 2022-04-21
[#171] Remove `checkStyle`

Tuubz:
#171 

## What happened 👀

We can remove [checkStyle](https://github.com/checkstyle/checkstyle) because it can only be used for Java and has been replaced completely by [detekt](https://github.com/detekt/detekt).

## Proof Of Work 📹

App & CI are still working as expected


luongvo (approved):
lgtm, that's enough I think👍

hoangnguyen92dn (approved):
I think there should be a `checkstyle` plugin applied to `build.gradle` but actually it is already added in DSL...

The rest LGTM 🚀

sleepylee (approved):

Issue #200, opened 2022-05-03
[KScript] Initialize project folder


Tuubz:
### Description
Initialize project folder, by copying all the files inside `CoroutineTemplate` and rename according to **app-name**

### Acceptance Criteria
- Show `"=> 🐢  Initializing new project..."` in the logs
- Initialize project folder by copying `CoroutineTemplate` **recursively** 
- Rename project folder according to **trimmed** `app-name`
   - If `app-name="My Project"`, then the project folder must be named `"MyProject"`

### Example
```
object NewProject {

    ...

    private fun validateArguments(args: Array<String>) {
      if (args.areValid() {
         initializeProjectFolder()
      }
    }

    private fun initializeProjectFolder() {
         // copy `CoroutineTemplate` recursively & rename folder name
    }
}

NewProject.generate(args)
```

### Reference
- https://github.com/nimblehq/android-templates/compare/kscript-poc?expand=1
- https://github.com/nimblehq/android-templates/blob/97ff612678ff9514a7a82a67fdf1ad7ff4c9c5bc/newproject.sh#L108-L109

PR #216, opened 2022-05-09
[#200] Initialize new project directory and copy files recursively

kaungkhantsoe:
[#200](https://github.com/nimblehq/android-templates/issues/200)

## What happened 👀

- Show "=> 🐢 Initializing new project..." in the logs
- Initialize project folder by copying CoroutineTemplate recursively
- Rename project folder according to trimmed app-name
   - If app-name="My Project", then the project folder must be named "MyProject"

## Insight 📝

- Create new folder for the project and name the project without space
- Copy files to new folder recursively
- Enable `gradlew` to be `executable`

## Proof Of Work 📹

https://user-images.githubusercontent.com/32578035/167353504-8e7f22e0-24b2-4a18-a9f3-8f9c5e9dd10e.mov

### DIFF
--- a/scripts/new_project.kts
+++ b/scripts/new_project.kts
@@ -1,14 +1,27 @@
+import java.io.File
+
 object NewProject {
 
+    private const val ARGUMENT_DELIMITER = "="
     private const val KEY_APP_NAME = "app-name"
     private const val KEY_PACKAGE_NAME = "package-name"
-    private const val ARGUMENT_DELIMITER = "="
+    private const val TEMPLATE_FOLDER_NAME = "CoroutineTemplate"
 
     private var appName = ""
+    private val appNameWithoutSpace: String
+        get() = appName.replace(" ", "")
     private var packageName = ""
 
     fun generate(args: Array<String>) {
         handleArguments(args)
+        initializeNewProjectFolder()
+    }
+
+    private fun initializeNewProjectFolder() {
+        showMessage("=> 🐢 Initializing new project...")
+        copyFiles(fromPath = TEMPLATE_FOLDER_NAME, toPath = appNameWithoutSpace)
+        // Set gradlew file as executable, because copying files from one folder to another doesn't copy file permissions correctly (= read, write & execute).
+        File("../$appNameWithoutSpace/gradlew")?.setExecutable(true)
     }
 
     private fun handleArguments(args: Array<String>) {
@@ -18,6 +31,19 @@ object NewProject {
                 KEY_APP_NAME -> appName = value
                 KEY_PACKAGE_NAME -> packageName = value
             }
+        }
+    }
+
+    private fun showMessage(message: String) {
+        println("\n${message}\n")
+    }
+
+    private fun copyFiles(fromPath: String, toPath: String) {
+        val targetFolder = File("../$toPath")
+        val sourceFolder = File("../$fromPath")
+        sourceFolder.copyRecursively(targetFolder, true) { file, exception ->
+            showMessage(exception?.message ?: "Error copying files")
+            return@copyRecursively OnErrorAction.TERMINATE
         }
     }
 }


scripts/new_project.kts
```diff
@@ -1,25 +1,68 @@
...
+                    showMessage(exception?.message)
+                    return@copyRecursively OnErrorAction.TERMINATE
+                }
+            }
+        }
+    }
 }
 
-NewProject.generate(args)
+NewProject.generate(args)
```
Tuubz, line 68: Please keep the new line at the end of the file 🚀

scripts/new_project.kts
```diff
@@ -1,25 +1,68 @@
...
         }
     }
+
+    private fun showMessage(message: String?) {
+        runBlocking {
+            println("\n${message?.trimIndent()}\n")
+        }
+    }
+
+    private suspend fun copyFiles(to: String) {
```
Tuubz, line 54: Should we rename `to: String` to `targetPath: String`? 🙏

scripts/new_project.kts
```diff
@@ -1,25 +1,68 @@
+#!/usr/bin/env kscript
+
+@file:MavenRepository("central", "https://repo.maven.apache.org/maven2/")
+@file:DependsOn("org.jetbrains.kotlinx:kotlinx-coroutines-core:1.6.1")
+
+import kotlinx.coroutines.*
+import java.io.File
+import kotlin.coroutines.coroutineContext
```
Tuubz, line 8: Could we sort the imports alphabetically? It will look a bit more organized that way 😁

scripts/new_project.kts
```diff
@@ -1,25 +1,68 @@
...
+
 object NewProject {
 
     private const val KEY_APP_NAME = "app-name"
     private const val KEY_PACKAGE_NAME = "package-name"
     private const val ARGUMENT_DELIMITER = "="
+    private const val COROUTINE_FOLDER_NAME = "CoroutineTemplate"
```
Tuubz, line 15: Should we keep the naming more general? Ex: `TEMPLATE_FOLDER_NAME`

As most likely we'll rename `CoroutineTemplate` to `template` in the future

scripts/new_project.kts
```diff
@@ -1,25 +1,68 @@
...
+    private const val COROUTINE_PACKAGE_THIRD_SEG = "coroutine"
```
Tuubz, line 16: Seems like this is unused, can we remove it? 🐟

scripts/new_project.kts
```diff
@@ -1,25 +1,68 @@
...
         handleArguments(args)
+
+        showMessage("=> 🐢 Initializing new project...")
+        copyFiles(to = appNameWithoutSpace)
+        /*
+         * Set gradlew file as executable
+         * Copying file from one folder to another lost it's previous execution right.
+         * Need to set gradlew as executable to run
+         */
+        File("$appNameWithoutSpace/gradlew")?.setExecutable(true)
```
Tuubz, line 32: Shall we put this into a new function, for example: `initializeProjectFolder()` / `initializeProjectDirectory()` ? 🙏

scripts/new_project.kts
```diff
@@ -1,25 +1,68 @@
...
 
     private const val KEY_APP_NAME = "app-name"
     private const val KEY_PACKAGE_NAME = "package-name"
     private const val ARGUMENT_DELIMITER = "="
+    private const val COROUTINE_FOLDER_NAME = "CoroutineTemplate"
+    private const val COROUTINE_PACKAGE_THIRD_SEG = "coroutine"
 
     private var appName = ""
     private var packageName = ""
+    private var appNameWithoutSpace = ""
```
Tuubz, line 20: Would it be nicer to have `appNameWithoutSpace` below `appName` 🤩 ?

scripts/new_project.kts
```diff
@@ -1,25 +1,68 @@
...
 
-    fun generate(args: Array<String>) {
+    fun generate(args: Array<String>) = runBlocking {
         handleArguments(args)
+
+        showMessage("=> 🐢 Initializing new project...")
```
Tuubz, line 25: Optional: Put it as a `const` -> Example: `MESSAGE_INITIALIZE_PROJECT`

scripts/new_project.kts
```diff
@@ -1,25 +1,68 @@
...
+        runBlocking {
+            println("\n${message?.trimIndent()}\n")
+        }
+    }
+
+    private suspend fun copyFiles(to: String) {
+        withContext(coroutineContext) {
+            async(Dispatchers.IO) {
+                val target = File(to)
+                val file = File("../$COROUTINE_FOLDER_NAME")
```
Tuubz, line 58: What about: `val targetFile` & `val sourceFile` 🚀

scripts/new_project.kts
```diff
@@ -1,25 +1,68 @@
...
+    private fun showMessage(message: String?) {
+        runBlocking {
+            println("\n${message?.trimIndent()}\n")
+        }
+    }
+
+    private suspend fun copyFiles(to: String) {
+        withContext(coroutineContext) {
+            async(Dispatchers.IO) {
+                val target = File(to)
```
Tuubz, line 57: Should this be `File("../$to")`? 

So we can copy it on to the `root` level instead of inside `scripts` 💭

scripts/new_project.kts
```diff
@@ -1,25 +1,68 @@
+#!/usr/bin/env kscript
+
+@file:MavenRepository("central", "https://repo.maven.apache.org/maven2/")
+@file:DependsOn("org.jetbrains.kotlinx:kotlinx-coroutines-core:1.6.1")
```
Tuubz, line 4: Could we also sort these alphabetically? (`@file:DependsOn` in front of `@file:MavenRepository`) 🙇

scripts/new_project.kts
```diff
@@ -1,25 +1,68 @@
...
+
+    private fun showMessage(message: String?) {
+        runBlocking {
+            println("\n${message?.trimIndent()}\n")
+        }
+    }
+
+    private suspend fun copyFiles(to: String) {
+        withContext(coroutineContext) {
+            async(Dispatchers.IO) {
```
Tuubz, line 56: By using coroutines, does it mean the script can execute faster than without? 🙏 

If yes, should we move this to `fun generate(...)` instead? So all functions are executed on `Dispatchers.IO`

scripts/new_project.kts
Tuubz: What about we merge them together? `toPath`? :sexy_girl

I got inspired by `File.java`:
Screen Shot 2565-05-09 at 17 22 09

scripts/new_project.kts
Tuubz: Yeah, no biggie. Although, by moving all of this into a function (ex: `initializeProjectFolder()`), it should be clear already that we're initializing the project 🙏

scripts/new_project.kts
Tuubz: Hmmmm, please allow me to search a bit more on this. I'm not so familiar with using coroutines in a script context. Do you have any reference/link I can look into by any chance? 🙏

scripts/new_project.kts
Tuubz: @kaungkhantsoe Oops! I believe it should be like this if we want to sort it alphabetically, right?

```
import java.io.File
import kotlin.coroutines.coroutineContext
import kotlinx.coroutines.*
``

File: scripts/new_project.kts [hunk 13]
```diff
@@ -1,25 +1,72 @@
...
+import java.io.File
+
 object NewProject {
 
+    private const val ARGUMENT_DELIMITER = "="
+    private const val COROUTINE_FOLDER_NAME = "CoroutineTemplate"
     private const val KEY_APP_NAME = "app-name"
     private const val KEY_PACKAGE_NAME = "package-name"
-    private const val ARGUMENT_DELIMITER = "="
+    private const val TEMPLATE_FOLDER_NAME = "coroutine"
```
Tuubz, line 16: Hmmm, I think there was a misunderstanding here 🙇  My suggestion was to rename `COROUTINE_FOLDER_NAME` to `TEMPLATE_FOLDER_NAME` and remove line 16 as it's not being used anywhere 🙏

scripts/new_project.kts
```diff
@@ -1,25 +1,72 @@
...
         }
     }
+
+    private fun showMessage(message: String?) {
+        runBlocking {
+            println("\n${message?.trimIndent()}\n")
+        }
+    }
+
+    private suspend fun copyFiles(toPath: String) {
```
Tuubz, line 58: Should we refactor it to `copyFiles(fromPath: String, toPath: String)`? It will be more flexible this way

`copyFiles(fromPath = "../$COROUTINE_FOLDER_NAME", toPath = "../$appNameWithoutSpace")`

Tuubz (commented):
A few more styling/renaming comments from my side 😁

scripts/new_project.kts
manh-t: @kaungkhantsoe when we run long-running tasks in an interface, I agree that we should use `IO` thread. But in this case, `copyRecursively` function is using `for loop` for the execution, and other complex tasks are neither executed nor are running on an interface. I think we can get rid of `coroutines` in the script 🙇  Please let me know your opinion too @nimblehq/android-chapter

scripts/new_project.kts
Tuubz: @kaungkhantsoe You can resolve this, by enabling "Ensure every saved file ends with a line break:

Screen Shot 2565-05-11 at 10 21 55

scripts/new_project.kts
Tuubz: @kaungkhantsoe Could you please share a screen recording that demonstrates why you chose to go with coroutines? I recall you mentioned to me that you faced an issue if we don't use coroutines 🙏

scripts/new_project.kts
```diff
@@ -1,25 +1,71 @@
...
+                    appName = value
+                    appNameWithoutSpace = appName.trim().replace(" ", "")
+                }
                 KEY_PACKAGE_NAME -> packageName = value
             }
         }
     }
+
+    private fun showMessage(message: String?) {
+        runBlocking {
```
Wadeewee, line 52: Do we need to use `runBlocking` here as we already execute show the message under the thread that we created in method `copyFiles` 🤔

Wadeewee:
https://github.com/nimblehq/android-templates/pull/216#discussion_r869831728

@kaungkhantsoe I also agree with @manh-t I don't think we need a suspendable computation for the execution script it should be simple🙏 Please correct me If I missed something 🙇‍♀️

scripts/new_project.kts
```diff
@@ -1,25 +1,55 @@
...
 
+    private const val ARGUMENT_DELIMITER = "="
     private const val KEY_APP_NAME = "app-name"
     private const val KEY_PACKAGE_NAME = "package-name"
-    private const val ARGUMENT_DELIMITER = "="
+    private const val TEMPLATE_FOLDER_NAME = "CoroutineTemplate"
 
     private var appName = ""
+    private val appNameWithoutSpace: String
+        get() = appName.trim().replace(" ", "")
```
Tuubz, line 14: Oh, could be it we don't actually need `trim()`, as we are already removing all the whitespaces with  `replace(" ", "")` ? 🙏

scripts/new_project.kts
```diff
@@ -1,25 +1,55 @@
...
+                KEY_APP_NAME -> {
+                    appName = value
+                }
                 KEY_PACKAGE_NAME -> packageName = value
             }
         }
     }
+
+    private fun showMessage(message: String) {
+        println("\n${message.trimIndent()}\n")
```
Tuubz, line 42: Do we need `trimIndent()`? 🙏

scripts/new_project.kts
```diff
@@ -1,25 +1,55 @@
...
     }
 
     private fun handleArguments(args: Array<String>) {
         args.forEach {
             val (key, value) = it.split(ARGUMENT_DELIMITER)
             when (key) {
-                KEY_APP_NAME -> appName = value
+                KEY_APP_NAME -> {
+                    appName = value
+                }
```
Tuubz, line 35: This could be reverted 🙏

scripts/new_project.kts
```diff
@@ -1,25 +1,55 @@
+#!/usr/bin/env kscript
```
Tuubz, line 1: Do we still need this? If I'm not mistaken we no longer need this, unless we import dependencies

scripts/new_project.kts
```diff
@@ -1,25 +1,71 @@
...
     private var packageName = ""
 
-    fun generate(args: Array<String>) {
+    fun generate(args: Array<String>) = runBlocking {
         handleArguments(args)
+
+        initializeNewProjectDirectory()
+    }
+
+    private suspend fun initializeNewProjectDirectory() {
```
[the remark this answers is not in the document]
Tuubz: @kaungkhantsoe Missed one here 👋 😁

Tuubz (commented):
Last one from my side, the rest looks perfect for me ⭐

Tuubz (approved):
🚀

scripts/new_project.kts
```diff
@@ -1,25 +1,51 @@
...
+        // Set gradlew file as executable, because copying files from one folder to another doesn't copy file permissions correctly (= read, write & execute).
+        File("../$appNameWithoutSpace/gradlew")?.setExecutable(true)
     }
 
     private fun handleArguments(args: Array<String>) {
         args.forEach {
             val (key, value) = it.split(ARGUMENT_DELIMITER)
             when (key) {
-                KEY_APP_NAME -> appName = value
+                KEY_APP_NAME ->  appName = value
```
Wadeewee, line 31: Please revert this change format, REST LGTM 🚀

Wadeewee (approved):

Tuubz:
One last thing here, do you think we should clean up the git commit history a bit? 🙇 

I'm fine if we squash them all into one commit, or be more descriptive

Screen Shot 2565-05-12 at 14 39 53

hoangnguyen92dn (approved):

Tuubz:
@manh-t One more approval from you and we'll merge this 🚀

manh-t (approved):

Issue #201, opened 2022-05-03
[KScript] Clean project folder


Tuubz:
### Description
Clean the project folder to start fresh.

### Acceptance Criteria
Inside the project folder:
- Run clean build with Gradle: `./gradlew clean`
- Remove `.idea` folder
- Remove `.gradle` folder
- Remove `.git` folder

**NOTE**: Don't run clean build and remove folders in `CoroutineTemplate`

### Example
```
object NewProject {

    ...

    private fun validateArguments(args: Array<String>) {
      if (args.areValid() {
         initializeProjectFolder()
         cleanProjectFolder()
      }
    }

    private fun cleanProjectFolder() {
         // run clean build & remove unucessary folders
    }
}

NewProject.generate(args)
```

### Reference
- https://github.com/nimblehq/android-templates/compare/kscript-poc?expand=1
- https://github.com/nimblehq/android-templates/blob/97ff612678ff9514a7a82a67fdf1ad7ff4c9c5bc/newproject.sh#L111-L118

PR #218, opened 2022-05-10
[#201] Clean project folder

kaungkhantsoe:
[#201](https://github.com/nimblehq/android-templates/issues/201)

## What happened 👀

- Clean the project folder to start fresh.
- Run clean build with Gradle: ./gradlew clean
- Remove .idea folder
- Remove .gradle folder
- Remove .git folder

## Insight 📝

- Run clean build bash command
- Remove .idea, .gradle, and .git files from new project folder

## Proof Of Work 📹

https://user-images.githubusercontent.com/32578035/167616609-d74b4723-43ce-4ec3-9b8d-31e6a6def322.mov

### DIFF
--- a/scripts/new_project.kts
+++ b/scripts/new_project.kts
@@ -11,17 +11,22 @@ object NewProject {
     private val appNameWithoutSpace: String
         get() = appName.replace(" ", "")
     private var packageName = ""
+    private val projectPath: String
+        get() = rootPath + appNameWithoutSpace
+    private val rootPath: String
+        get() = System.getProperty("user.dir").replace("scripts", "")
 
     fun generate(args: Array<String>) {
         handleArguments(args)
         initializeNewProjectFolder()
+        cleanNewProjectFolder()
     }
 
     private fun initializeNewProjectFolder() {
         showMessage("=> 🐢 Initializing new project...")
-        copyFiles(fromPath = TEMPLATE_FOLDER_NAME, toPath = appNameWithoutSpace)
+        copyFiles(fromPath = rootPath + TEMPLATE_FOLDER_NAME, toPath = projectPath)
         // Set gradlew file as executable, because copying files from one folder to another doesn't copy file permissions correctly (= read, write & execute).
-        File("../$appNameWithoutSpace/gradlew")?.setExecutable(true)
+        File(projectPath + File.separator + "gradlew")?.setExecutable(true)
     }
 
     private fun handleArguments(args: Array<String>) {
@@ -39,12 +44,27 @@ object NewProject {
     }
 
     private fun copyFiles(fromPath: String, toPath: String) {
-        val targetFolder = File("../$toPath")
-        val sourceFolder = File("../$fromPath")
+        val targetFolder = File(toPath)
+        val sourceFolder = File(fromPath)
         sourceFolder.copyRecursively(targetFolder, true) { file, exception ->
             showMessage(exception?.message ?: "Error copying files")
             return@copyRecursively OnErrorAction.TERMINATE
         }
+    }
+
+    private fun cleanNewProjectFolder() {
+        executeCommand("sh $projectPath${File.separator}gradlew -p $projectPath clean")
+        executeCommand("sh $projectPath${File.separator}gradlew -p $projectPath${File.separator}buildSrc clean")
+        listOf(".idea",".gradle","buildSrc${File.separator}.gradle",".git").forEach {
+            File( "$projectPath${File.separator}$it")?.let { targetFile ->
+                targetFile.deleteRecursively()
+            }
+        }
+    }
+
+    private fun executeCommand(command: String) {
+        val process = Runtime.getRuntime().exec(command)
+        process.inputStream.reader().forEachLine { println(it) }
     }
 }
 


scripts/new_project.kts
```diff
@@ -54,6 +58,27 @@ object NewProject {
         }
     }
 
+    private suspend fun cleanOldBuildFiles() {
+        withContext(coroutineContext) {
+            async {
+                // Clean old build
+                "sh \"../$appNameWithoutSpace\"/gradlew -p \"../$appNameWithoutSpace\" clean".exec()
```
Tuubz, line 65: Hmmm, I'm not a big fan of this approach, as it requires to escape `"` with `\`

Have you tried [my approach](https://github.com/nimblehq/android-templates/compare/kscript-poc?expand=1#diff-bc4c132346cb8367e875b5d0cf2ff92b86d04b2b7ad9bbef0dd1674c9a649ceaR116-R122), it requires a lot less code.

If yes, any difference? 🙏

scripts/new_project.kts
```diff
@@ -54,6 +58,27 @@ object NewProject {
...
+                    val filePath = it.toString()
+                    if (
+                        filePath.endsWith(".idea") ||
+                        filePath.endsWith(".gradle") ||
+                        filePath.endsWith(".git")
+                    ) {
+                        File(filePath)?.deleteRecursively()
+                    }
+                }
+            }
```
Tuubz, line 78: Instead of looping through all the folders one by one, would it be better to specify the path directly, similar to [this function](https://github.com/nimblehq/android-templates/compare/kscript-poc?expand=1#diff-bc4c132346cb8367e875b5d0cf2ff92b86d04b2b7ad9bbef0dd1674c9a649ceaR78-R84) instead? 💭

scripts/new_project.kts
```diff
@@ -54,6 +58,27 @@ object NewProject {
         }
     }
 
+    private suspend fun cleanOldBuildFiles() {
```
Tuubz, line 61: What about `cleanNewProjectFolder()` ? 🙏

scripts/new_project.kts
kaungkhantsoe: I've tried running with `./` before but it says `permission denied`. Also I could not run the bash using the proposed approach. Please let me know if anyone can. 🙏

Tuubz: Hmmm, I see. I'll try again soon, maybe I overlooked something when I tried before.  🚀

scripts/new_project.kts
Tuubz: Oh, I see 👍  From my understanding is that `./gradlew clean` will remove `buildSrc` anyways, so the only one left would be `../MyProject/.gradle`. Please correct me if I'm wrong 🙏 

> Is there a specific set of files that we want to delete?

Looking at [the current implemenaton in newproject.sh](https://github.com/nimblehq/android-templates/blob/99b064dc893ab678dbb7d0e86249ed39c1bcff36/newproject.sh#L114-L118), seems like they remove a specific set of folders:

- `.idea` 
- `.gradle`
- `.git`

Tuubz: > From my understanding is that ./gradlew clean will remove buildSrc anyways

My bad, I see your point now. There's indeed another `./gradle` folder inside `buildSrc` 💭 

But seems like they didn't remove it it in the previous script

scripts/new_project.kts
Tuubz: @kaungkhantsoe I could be wrong, but this is working for me. Please try it out 🤩  

```
private fun cleanOldBuildFiles() {
    executeCommand("./../$appNameWithoutSpace/gradlew -p ../$appNameWithoutSpace clean")
    executeCommand("./../$appNameWithoutSpace/gradlew -p ../$appNameWithoutSpace/buildSrc clean")
     ...
{

private fun executeCommand(command: String) {
    val process = Runtime.getRuntime().exec(command)
    InputStreamReader(process.inputStream).forEachLine {
        println(it)
    }
}
```

kaungkhantsoe: @Tuubz I could run with it. I am not sure why I couldn't run it before 🧐. I've updated the code using your approach.

scripts/new_project.kts
```diff
@@ -1,4 +1,6 @@
+import java.io.BufferedReader
```
Tuubz, line 1: Redundant import 😁

Tuubz (approved):
Looks clean! 🚀

scripts/new_project.kts
```diff
@@ -46,6 +48,22 @@ object NewProject {
...
+        File("../$appNameWithoutSpace/.gradle")?.deleteRecursively()
+        File("../$appNameWithoutSpace/buildSrc/.gradle")?.deleteRecursively()
+        File("../$appNameWithoutSpace/.git")?.deleteRecursively()
+    }
+
+    private fun executeCommand(command: String) {
+        val process = Runtime.getRuntime().exec(command)
+        InputStreamReader(process.inputStream).forEachLine {
+            println(it)
+        }
```
manh-t, line 65: small improvement:
```suggestion
        val process = Runtime.getRuntime().exec(command)
        process.inputStream.reader().forEachLine {
            println(it)
        }
```

scripts/new_project.kts
```diff
@@ -46,6 +48,22 @@ object NewProject {
...
         }
     }
+
+    private fun cleanNewProjectFolder() {
+        executeCommand("./../$appNameWithoutSpace/gradlew -p ../$appNameWithoutSpace clean")
+        executeCommand("./../$appNameWithoutSpace/gradlew -p ../$appNameWithoutSpace/buildSrc clean")
+        File("../$appNameWithoutSpace/.idea")?.deleteRecursively()
+        File("../$appNameWithoutSpace/.gradle")?.deleteRecursively()
+        File("../$appNameWithoutSpace/buildSrc/.gradle")?.deleteRecursively()
+        File("../$appNameWithoutSpace/.git")?.deleteRecursively()
```
manh-t, line 58: relative-path like `./../` is error-prone and it may cause issues if we update the project path in the future 💭  (we have to find all mentioned paths and update them 🤪 )
I would suggest getting the current working directory + the new project directory instead. What do you think?

scripts/new_project.kts
Tuubz: Nice, maybe we don't need the import `import java.io.InputStreamReader` then

scripts/new_project.kts
Tuubz: @manh-t Fair point 👍 

Could you show us some example code for better understanding, please?

manh-t: @Tuubz @kaungkhantsoe currently the new project will be stored in the root `android-templates`. We can define the project path and then append the directories:
```
val root = File("").absolutePath.toString().replace(File.separator + "scripts", "")
val projectPath = root + File.separator + appNameWithoutSpace
listOf(".idea", ".gradle", "buildSrc" + File.separator + ".gradle", ".git").forEach {
    val targetFile = File(projectPath + File.separator + it)
    if (targetFile.exists()) targetFile.deleteRecursively()
}
```
Notice that the template can be init from any platform, so `File.separator` is recommended instead of `"/"` 😁

Tuubz: > Notice that the template can be init from any platform, so File.separator is recommended instead of "/" 😁

@manh-t Just curious, any idea which platform doesn't use `"/"` as a file separtor?

manh-t: @Tuubz windows for sure `"\"`

Tuubz: Aaaah, I see. Thanks 😁 

Yes, let's do it with the `File.separator` then 🚀

kaungkhantsoe: Updated 🚀

Wadeewee (approved):
LGTM 🚀

scripts/new_project.kts
```diff
@@ -39,13 +40,31 @@ object NewProject {
...
+    }
+
+    private fun executeCommand(command: String) {
+        val process = Runtime.getRuntime().exec(command)
+        process.inputStream.reader().forEachLine { println(it) }
+    }
+
+    private fun getRootPath(): String {
+        return System.getProperty("user.dir").replace("scripts", "")
+    }
```
Tuubz, line 67: I see there's a lot of redundancy on `getRootPath() + appNameWithoutSpace`

Should we bundle them to a top-level variable so we can re-use it anywhere?

kaungkhantsoe: Updated again

Tuubz: Sorry to keep on going on, but wouldn't something like this be a cleaner solution 🙇 
```
    private val rootPath: String
        get() = System.getProperty("user.dir").replace("scripts", "")

    private val projectPath: String
        get() = rootPath + appNameWithoutSpace
```

I'm open to other suggestions 🙏

kaungkhantsoe: It's okay. I was wondering about the same thing too. Updated ✌️

scripts/new_project.kts
```diff
@@ -38,14 +39,32 @@ object NewProject {
         println("\n${message}\n")
     }
 
-    private fun copyFiles(fromPath: String, toPath: String) {
-        val targetFolder = File("../$toPath")
-        val sourceFolder = File("../$fromPath")
+    private fun copyFiles(fromPath: String) {
+        val targetFolder = File(getRootPathOfNewProjectFolder())
+        val sourceFolder = File(getRootPathOfNewProjectFolder().replace(appNameWithoutSpace, "") + fromPath)
```
manh-t: I think we should keep the old implementation instead since we may re-use this function for other purposes. Currently it's hard for setting to copy files from a specific folder to another folder 😅  For example, my case is to copy files from the old package path `co.nimblehq.coroutine` to a new package path `com.test.abc`.

scripts/new_project.kts
```diff
@@ -39,13 +44,27 @@ object NewProject {
...
+        val sourceFolder = File(fromPath)
         sourceFolder.copyRecursively(targetFolder, true) { file, exception ->
             showMessage(exception?.message ?: "Error copying files")
             return@copyRecursively OnErrorAction.TERMINATE
         }
     }
+
+    private fun cleanNewProjectFolder() {
+        executeCommand("sh " + projectPath + File.separator + "gradlew -p " + projectPath + " clean")
+        executeCommand("sh " + projectPath + File.separator + "gradlew -p " + projectPath + "/buildSrc clean")
```
Tuubz, line 57: Last one from me 😍 

```suggestion
        executeCommand("sh $projectPath${File.separator} gradlew -p $projectPath${File.separator}buildSrc clean")

```

scripts/new_project.kts
```diff
@@ -38,14 +39,32 @@ object NewProject {
...
         }
     }
+
+    private fun cleanNewProjectFolder() {
+        executeCommand("sh " + getRootPathOfNewProjectFolder() + File.separator + "gradlew -p " + getRootPathOfNewProjectFolder() + " clean")
+        executeCommand("sh " + getRootPathOfNewProjectFolder() + File.separator + "gradlew -p " + getRootPathOfNewProjectFolder() + "/buildSrc clean")
+        File(getRootPathOfNewProjectFolder() + File.separator + ".idea")?.deleteRecursively()
+        File(getRootPathOfNewProjectFolder() + File.separator + ".gradle")?.deleteRecursively()
+        File(getRootPathOfNewProjectFolder() + File.separator + "buildSrc" + File.separator + ".gradle")?.deleteRecursively()
+        File(getRootPathOfNewProjectFolder() + File.separator + ".git")?.deleteRecursively()
```
manh-t: can we refactor these lines to avoid duplicated code please? 🙏 
```suggestion
        listOf(".idea", ".gradle", "buildSrc" + File.separator + ".gradle", ".git").forEach {
            val targetFile = File(getRootPathOfNewProjectFolder() + File.separator + it)
            if (targetFile.exists()) targetFile.deleteRecursively()
        }
```

Tuubz (commented):
Also, shall we squash the re-factor commits in to one refactor commit? 🙏

scripts/new_project.kts
Tuubz: @manh-t Updated 🚀

manh-t (approved):

doannimble (approved):

Issue #204, opened 2022-05-03
[KScript] Rename folder structure for `androidTest`


Tuubz:
## Description
Rename folder structures inside the project folder for `androidTest` according to `package-name`

**NOTE**: `CoroutineTemplate` currently doesn't contain `androidTest`

### Acceptance Criteria
- Check if `androidTest` exists before renaming **OR** add empty `androidTest` to `CoroutineTemplate` 
- Rename folder structures in the project folder for `androidTest`
- If `package-name="co.myproject.example"`, then the result must look like this:

![Image](https://user-images.githubusercontent.com/18277915/166436237-ef2a9c0b-9824-4f3a-89fa-7d123fb77b48.png)

### Example
```
object NewProject {

    ...

    private fun renameFolderStructures() {
       // rename folder structure for `app`
       // rename folder structure for `data`
       // rename folder structure for `domain`
    }
}

NewProject.generate(args)
```

### Reference
- https://github.com/nimblehq/android-templates/compare/kscript-poc?expand=1
- https://github.com/nimblehq/android-templates/blob/39a3ee5fc5f1fa259661db5f26234f9f86342aaf/newproject.sh#L157-L174

Issue #203, opened 2022-05-03
[KScript] Rename folder structure for `app`, `data` & `domain`


Tuubz:
### Description
Rename folder structures inside the project folder for `app`, `data` & `domain` according to `package-name`

### Acceptance Criteria
- Show `"=> 🔎  Renaming folder structure..."` in the log
- Rename folder structures in the project folder for `app`, `data` & `domain`
- If `package-name="co.myproject.example"`, then the result must look like this:

![Image](https://user-images.githubusercontent.com/18277915/166435910-8f011475-8ae9-4563-896a-6657a28afba0.png)

### Example
```
object NewProject {

    ...

    private fun renameFolderStructures() {
       // rename folder structure for `app`
       // rename folder structure for `data`
       // rename folder structure for `domain`
    }
}

NewProject.generate(args)
```

### Reference
- https://github.com/nimblehq/android-templates/compare/kscript-poc?expand=1
- https://github.com/nimblehq/android-templates/blob/39a3ee5fc5f1fa259661db5f26234f9f86342aaf/newproject.sh#L147-L155

Issue #202, opened 2022-05-03
[KScript] Prepare function to rename folder structures


Tuubz:
### Description
Rename folder structures inside the project folder according to `package-name`.

### Acceptance Criteria
- Prepare a function to rename folder structures, don't execute it anywhere **yet**
- If `package-name="co.myproject.example"`, then the result must look like this:

![Image](https://user-images.githubusercontent.com/18277915/166434721-15ff6901-fb8b-4cbf-be87-abea43318484.png)

### Example
```
object NewProject {

    ...

    private fun renameFolderStructure() {
         // rename folder structure according to `package-name`
    }
}

NewProject.generate(args)
```

### Reference
- https://github.com/nimblehq/android-templates/compare/kscript-poc?expand=1
- https://github.com/nimblehq/android-templates/blob/97ff612678ff9514a7a82a67fdf1ad7ff4c9c5bc/newproject.sh#L120-L142

PR #220, opened 2022-05-13
[#202][#203][#204][#205] Rename folder structure for existing modules

manh-t:
https://github.com/nimblehq/android-templates/issues/202
https://github.com/nimblehq/android-templates/issues/203
https://github.com/nimblehq/android-templates/issues/204
https://github.com/nimblehq/android-templates/issues/205

## What happened 👀

Rename folder structures inside the project folder

## Insight 📝

- We list the modules that need to be restructure via `modules`.

## Proof Of Work 📹

https://user-images.githubusercontent.com/60863885/168269918-eb3aebce-1c9b-489b-987f-595416a7cfd4.mp4

### DIFF
--- a/scripts/new_project.kts
+++ b/scripts/new_project.kts
@@ -54,46 +54,27 @@ object NewProject {
                 .maxDepth(2)
                 .filter { it.isDirectory && it.name == "java" }
                 .forEach { javaDirectory ->
-                    val javaPath = javaDirectory.absolutePath
-                    val currentPackagePath =
-                        javaPath + File.separator + TEMPLATE_PACKAGE_NAME.replace(
-                            DOT_SEPARATOR,
-                            File.separator
+                    val oldDirectory = File(
+                        javaDirectory, TEMPLATE_PACKAGE_NAME.replace(
+                            oldValue = DOT_SEPARATOR,
+                            newValue = File.separator
                         )
-                    val newPackagePath =
-                        javaPath + File.separator + packageName.replace(
-                            DOT_SEPARATOR,
-                            File.separator
+                    )
+                    val newDirectory = File(
+                        javaDirectory, packageName.replace(
+                            oldValue = DOT_SEPARATOR,
+                            newValue = File.separator
                         )
-                    if (File(currentPackagePath).exists()) {
-                        copyCurrentPackageToNewPackageAndClean(
-                            currentPackagePath = currentPackagePath,
-                            tempPath = javaPath + File.separator + "temp_folder",
-                            newPackagePath = newPackagePath
-                        )
-                    }
+                    )
+
+                    val tempDirectory = File(javaDirectory, "temp_directory")
+                    oldDirectory.copyRecursively(tempDirectory)
+                    oldDirectory.parentFile?.parentFile?.deleteRecursively()
+                    newDirectory.mkdirs()
+                    tempDirectory.copyRecursively(newDirectory)
+                    tempDirectory.deleteRecursively()
                 }
         }
-    }
-
-    private fun copyCurrentPackageToNewPackageAndClean(
-        currentPackagePath: String,
-        tempPath: String,
-        newPackagePath: String
-    ) {
-        copyFiles(fromPath = currentPackagePath, toPath = tempPath)
-        deleteFilesOrDirectories(
-            currentPackagePath.replace(
-                TEMPLATE_PACKAGE_NAME.replace(DOT_SEPARATOR, File.separator),
-                TEMPLATE_PACKAGE_NAME.split(DOT_SEPARATOR).first()
-            )
-        )
-        copyFiles(fromPath = tempPath, toPath = newPackagePath)
-        deleteFilesOrDirectories(tempPath)
-    }
-
-    private fun deleteFilesOrDirectories(fromPath: String) {
-        File(fromPath).deleteRecursively()
     }
 
     private fun showMessage(message: String) {
--- a/scripts/new_project.kts
+++ b/scripts/new_project.kts
@@ -3,10 +3,13 @@ import java.io.File
 object NewProject {
 
     private const val ARGUMENT_DELIMITER = "="
+    private const val DOT_SEPARATOR = "."
     private const val KEY_APP_NAME = "app-name"
     private const val KEY_PACKAGE_NAME = "package-name"
     private const val TEMPLATE_FOLDER_NAME = "CoroutineTemplate"
     private const val TEMPLATE_PACKAGE_NAME = "co.nimblehq.coroutine"
+
+    private val modules = listOf("app", "data", "domain")
 
     private var appName = ""
     private val appNameWithoutSpace: String
@@ -21,6 +24,7 @@ object NewProject {
         handleArguments(args)
         initializeNewProjectFolder()
         cleanNewProjectFolder()
+        repackage()
         renamePackageNameWithinFiles()
     }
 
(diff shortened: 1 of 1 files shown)


scripts/new_project.kts
```diff
@@ -2,27 +2,42 @@ import java.io.File
 
 object NewProject {
 
+    private const val APP_MODULE = "app"
```
manh-t, line 5: oh, this is redundant, I forgot to remove. Will update

manh-t:
rebased 🙏

scripts/new_project.kts
```diff
@@ -34,18 +51,102 @@ object NewProject {
         }
     }
 
+    private fun repackage() {
+        showMessage("=> 🔎 Rename the package folders...")
+        modules.forEach { module ->
+            val srcPath = newProjectPath + File.separator + module + File.separator + "src"
+            for (file in File(srcPath).listFiles()) {
```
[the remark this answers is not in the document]
manh-t: @Tuubz unfortunately we cannot use `walk` in this case since `walk` will loop and find all nested directories in the module, but we only need the directories inside `src` folder 😅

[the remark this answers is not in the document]
manh-t: @Tuubz but what if there is more than a folder named "java" in our project? It could be a false case 😄  Right now I just want to update the folder `src/[main/staging/production]/java`, the other cases like `src/[main/staging/production]/java/.../java` is not the expected result.
Let's try to test by creating a `java` folder, and the result is:
```
walk /Users/tranmanh/Desktop/nimble/github/android-templates/TestProject/app/src/test/java
walk /Users/tranmanh/Desktop/nimble/github/android-templates/TestProject/app/src/main/java
walk /Users/tranmanh/Desktop/nimble/github/android-templates/TestProject/app/src/main/java/co/nimblehq/coroutine/extension/java
walk /Users/tranmanh/Desktop/nimble/github/android-templates/TestProject/app/src/debug/java
walk /Users/tranmanh/Desktop/nimble/github/android-templates/TestProject/data/src/test/java
walk /Users/tranmanh/Desktop/nimble/github/android-templates/TestProject/data/src/main/java
walk /Users/tranmanh/Desktop/nimble/github/android-templates/TestProject/domain/src/test/java
walk /Users/tranmanh/Desktop/nimble/github/android-templates/TestProject/domain/src/main/java
```
Yeah, `/Users/tranmanh/Desktop/nimble/github/android-templates/TestProject/app/src/main/java/co/nimblehq/coroutine/extension/java` is not the folder we want to update 🤪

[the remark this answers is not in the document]
Tuubz: @manh-t Oh, you're right

In that case, we can use `maxDepth(2)` 😁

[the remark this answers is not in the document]
manh-t: @Tuubz that makes sense now 🎉

manh-t:
@Tuubz @kaungkhantsoe updated 🙏

scripts/new_project.kts
```diff
@@ -41,6 +45,57 @@ object NewProject {
...
+                            DOT_SEPARATOR,
+                            File.separator
+                        )
+                    if (File(currentPackagePath).exists()) {
+                        copyCurrentPackageToNewPackageAndClean(
+                            currentPackagePath = currentPackagePath,
+                            tempPath = javaPath + File.separator + "temp_folder",
+                            newPackagePath = newPackagePath
+                        )
+                    }
```
Tuubz, line 74: This solution works well, but I was hoping we could make use of `renameTo()` so we don't require `temp_folder`.

What do you think about this solution? 🙇 

If it works we can minimize the code 🔥 

```suggestion
                   val oldDirectory = File(
                        javaDirectory, TEMPLATE_PACKAGE_NAME.replace(
                            oldValue = DOT_SEPARATOR,
                            newValue = File.separator
                        )
                    )
                    val newDirectory = File(
                        javaDirectory, packageName.replace(
                            oldValue = DOT_SEPARATOR,
                            newValue = File.separator
                        )
                    )
                    newDirectory.mkdirs()
                    oldDirectory.renameTo(newDirectory)
                    oldDirectory.parentFile.parentFile.deleteRecursively()
```

kaungkhantsoe:
LGTM 🚀 but it would be better if can get rid of `temp_folder` like @Tuubz said.

kaungkhantsoe (approved):

scripts/new_project.kts
manh-t: @Tuubz this solution looks good, but it's failed in case of the new package path is similar to the old package path, ex `co.nimblehq.testproject` 🙏   The new package folder will be deleted 😓 
Here's the result when I run the above solution 😂  `java` folder is empty.
image

That's why I create a `temp` folder to store all files and then delete the old package folder.
Do we have other solutions for this case? 👉  Yes, it requires the calculation to delete the correct folder (delete only `testproject` in this case) but I think we can go with my simple solution (use `temp` folder to store files temporarily) instead.
If you have other ideas, please let me know 😁  I'm open to discussing more.

Tuubz: Good catch, I missed that one! 👍 

OK, I agree that the temp folder is necessary to make this work

However, I do have an urge to simplify it a bit, for easier future maintainability 🙇 

What do you think about this solution? (It also makes use of `temp_folder`):

```
File(srcPath)
        .walk()
        .maxDepth(2)
        .filter { it.isDirectory && it.name == "java" }
        .forEach { javaDirectory ->
            val oldDirectory = File(
                javaDirectory, TEMPLATE_PACKAGE_NAME.replace(
                    oldValue = DOT_SEPARATOR,
                    newValue = File.separator
                )
            )
            val newDirectory = File(
                javaDirectory, packageName.replace(
                    oldValue = DOT_SEPARATOR,
                    newValue = File.separator
                )
            )

            val tempDirectory = File(javaDirectory, "temp_directory")
            oldDirectory.copyRecursively(tempDirectory)
            oldDirectory.parentFile.parentFile.deleteRecursively()
            newDirectory.mkdirs()
            tempDirectory.copyRecursively(newDirectory)
            tempDirectory.deleteRecursively()
        }
```

Last suggestion ❤️

scripts/new_project.kts
```diff
@@ -41,6 +45,38 @@ object NewProject {
...
+                    )
+                    val newDirectory = File(
+                        javaDirectory, packageName.replace(
+                            oldValue = DOT_SEPARATOR,
+                            newValue = File.separator
+                        )
+                    )
+
+                    val tempDirectory = File(javaDirectory, "temp_directory")
+                    oldDirectory.copyRecursively(tempDirectory)
```
Wadeewee, line 71: Can we execute `copyRecursively` in `copyFiles` method  or adapt parameter inside there 🤔  So we can know when error

Wadeewee (approved):

scripts/new_project.kts
```diff
@@ -41,6 +45,38 @@ object NewProject {
...
+                    oldDirectory.parentFile?.parentFile?.deleteRecursively()
+                    newDirectory.mkdirs()
+                    tempDirectory.copyRecursively(newDirectory)
+                    tempDirectory.deleteRecursively()
```
Wadeewee, line 75: Should we handle error when try to `deleteRecursively` I just saw we handle error in `copyFiles` 🙏

Wadeewee (commented):
Rest LGTM 🚀

scripts/new_project.kts
Tuubz: I can't think of any case where it would be an error unless `tempDirectory` doesn't exist.

But here we can be sure that `tempDirectory` as we created it ourselves, 

So I don't think we need it, what do you think? 🙏

scripts/new_project.kts
kaungkhantsoe: > @Tuubz this solution looks good, but it's failed in case of the new package path is similar to the old package path, ex `co.nimblehq.testproject` 🙏 The new package folder will be deleted 😓 Here's the result when I run the above solution 😂 `java` folder is empty. image
> 
> That's why I create a `temp` folder to store all files and then delete the old package folder. Do we have other solutions for this case? 👉 Yes, it requires the calculation to delete the correct folder (delete only `testproject` in this case) but I think we can go with my simple solution (use `temp` folder to store files temporarily) instead. If you have other ideas, please let me know 😁 I'm open to discussing more.

The folder is deleted because of the thread I think. The script got executed line by line and the old folder got deleted while it is renaming the new folder.

scripts/new_project.kts
Wadeewee: Ah, I see 👍  No need then 🎉

scripts/new_project.kts
kaungkhantsoe: We can use the `copyFiles()` if we wish now, like`copyFiles(oldDirectory.absolutePath,tempDirectory.absolutePath)` isn't it? I think the current implementation is also acceptable since it is the native android api. cc @Tuubz @Wadeewee

manh-t: oh, we should use `copyFiles` method in this case. They are the same 😓 @kaungkhantsoe please help me refactor this in your PR 💘 Many thanks 💯

Issue #209, opened 2022-05-03
[KScript] Run Gradle scripts


Tuubz:
## Description
Run Gradle scripts to ensure everything is still working as expected

### Acceptance Criteria
Run and show output in the logs for the following scripts:
- `./gradlew assembleDebug` > Show `"=> 🛠️  Building project..."` in the log
- `./gradlew testStagingDebugUnitTest` >  Show `"=> 🚓  Running tests..."` in the log
- Show `"=> 🚀  Done! The project is ready for development"`

### Example
```
object NewProject {

    ...

    private fun runGradleScript() {
       // run gradle script and show output in logs
    }
}

NewProject.generate(args)
```

### Reference
- https://github.com/nimblehq/android-templates/compare/kscript-poc?expand=1
- https://github.com/nimblehq/android-templates/blob/39a3ee5fc5f1fa259661db5f26234f9f86342aaf/newproject.sh#L251-L257

PR #224, opened 2022-05-20
[#209] Build project and run tests

kaungkhantsoe:
[#209](https://github.com/nimblehq/android-templates/issues/209)

## What happened 👀

Run and show output in the logs for the following scripts:

./gradlew assembleDebug > Show "=> 🛠️ Building project..." in the log
./gradlew testStagingDebugUnitTest > Show "=> 🚓 Running tests..." in the log
Show "=> 🚀 Done! The project is ready for development"

## Insight 📝

- Run `sh gradlew -p project assembleDebug` for building
- Run `.h gradlew -p project testStagingDebugUnitTest` for unit test
- Refactor the code stracture to see the main method first then util methods

## Proof Of Work 📹

https://user-images.githubusercontent.com/32578035/169455515-98683cd9-40f9-4f8e-9d51-a6489a1d6df3.mp4

### DIFF
--- a/scripts/new_project.kts
+++ b/scripts/new_project.kts
@@ -14,6 +14,7 @@ object NewProject {
     private var appName = ""
     private val appNameWithoutSpace: String
         get() = appName.replace(" ", "")
+    private val fileSeparator = File.separator
     private var packageName = ""
     private val projectPath: String
         get() = rootPath + appNameWithoutSpace
@@ -26,13 +27,7 @@ object NewProject {
         cleanNewProjectFolder()
         renamePackageNameFolders()
         renamePackageNameWithinFiles()
-    }
-
-    private fun initializeNewProjectFolder() {
-        showMessage("=> 🐢 Initializing new project...")
-        copyFiles(fromPath = rootPath + TEMPLATE_FOLDER_NAME, toPath = projectPath)
-        // Set gradlew file as executable, because copying files from one folder to another doesn't copy file permissions correctly (= read, write & execute).
-        File(projectPath + File.separator + "gradlew")?.setExecutable(true)
+        buildProjectAndRunTests()
     }
 
     private fun handleArguments(args: Array<String>) {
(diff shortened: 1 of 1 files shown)


Tuubz (commented):
Appreciate the code re-factor 🎸

scripts/new_project.kts
```diff
@@ -90,35 +124,11 @@ object NewProject {
...
-            File("$projectPath${File.separator}$it")?.let { targetFile ->
-                targetFile.deleteRecursively()
-            }
-        }
-    }
-
     private fun executeCommand(command: String) {
         val process = Runtime.getRuntime().exec(command)
         process.inputStream.reader().forEachLine { println(it) }
     }
```
Tuubz, line 130: I think we should terminate/kill the process if something fails 💭 

Let's say, we run the Gradle script for running tests and if one of the tests fails, it will still show: `🚀 Done! The project is ready for development`

---

My suggestion is to do something like this:

```
private fun executeCommand(command: String) {
    val process = Runtime.getRuntime().exec(command)
    process.inputStream.reader().forEachLine { println(it) }
    val exitValue = process.waitFor()
    if (exitValue != 0) System.exit(exitValue)
}
```

So if the command resulted in an error, then kill the whole process, otherwise proceed 🤖

scripts/new_project.kts
```diff
@@ -57,26 +69,48 @@ object NewProject {
...
                     tempDirectory.deleteRecursively()
                 }
         }
     }
 
+    private fun renamePackageNameWithinFiles() {
+        showMessage("=> 🔎 Renaming package name within files...")
+        File(projectPath)
+            ?.walk()
+            .filter { it.isFile && it.name != "debug.keystore" && it.name != "gradle-wrapper.jar"}
```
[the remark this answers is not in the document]
kaungkhantsoe: As usual, it got corrupted after rename function. 😅 So I have neglected that file from the method. I think we should only rewrite `.class`, `.kt`, and `.xml` files because whenever we try to rewrite the other types of files, they got corrupted.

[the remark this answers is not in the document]
Tuubz: Hmmm, good point 👍 

Then I wonder if we should change it to: `.filter { it.endsWith(".kt") || it.endsWith(".xml") }` instead

Any reason we should include `.class` too? 🙏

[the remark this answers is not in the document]
kaungkhantsoe: Nope. I just explained it as an example 😆

scripts/new_project.kts
kaungkhantsoe: Shall we only terminate the task or shall we also prompt like `"Something went wrong!!"` before we terminate the process? @Tuubz

Tuubz: Yeah, adding an additional error message sounds like a good plan 🚀

kaungkhantsoe: Can you suggest what we should prompt? @Tuubz 🙏

Tuubz: `❌ Something went wrong!` works for m 😁

Tuubz (approved):
🔥 🔥 🔥

manh-t (approved):
No issues from me

sleepylee (approved):

Wadeewee (approved):
LGTM  🚀

Issue #208, opened 2022-05-03
[KScript] Rename application class


Tuubz:
## Description
Rename application class within the project folder according to `app-name`

### Acceptance Criteria
- Show `"=> 🔎  Renaming application class..."` in the log
- Rename application class according to `app-name`, including `AndroidManifest.xml`
- If `app-name="My Project"`, then the result must be `MyProjectApplication` and look like this:

![Image](https://user-images.githubusercontent.com/18277915/166439026-81d69f58-b8c0-410f-aebf-c07b74382fa6.png)

### Example
```
object NewProject {

    ...

    private fun renameApplicationClass() {
       // rename application class
    }
}

NewProject.generate(args)
```

### Reference
- https://github.com/nimblehq/android-templates/compare/kscript-poc?expand=1
- https://github.com/nimblehq/android-templates/blob/97ff612678ff9514a7a82a67fdf1ad7ff4c9c5bc/newproject.sh#L245-L249

PR #225, opened 2022-05-20
[#208] Rename application class

Tuubz:
#208 

## What happened 👀

Rename application class according to `app-name` argument, there are two processes:
1. Rename application class file (Ex: `CoroutineTemplateApplication` to `MyProjectApplication`)
2. Rename application class in `AndroidManifest.xml`

## Proof Of Work 📹

https://user-images.githubusercontent.com/18277915/169508816-15042a0c-6553-489f-b70c-b262cd2323b0.mov

### DIFF
--- a/scripts/new_project.kts
+++ b/scripts/new_project.kts
@@ -6,6 +6,7 @@ object NewProject {
     private const val DOT_SEPARATOR = "."
     private const val KEY_APP_NAME = "app-name"
     private const val KEY_PACKAGE_NAME = "package-name"
+    private const val TEMPLATE_APPLICATION_CLASS_NAME = "CoroutineTemplateApplication"
     private const val TEMPLATE_FOLDER_NAME = "CoroutineTemplate"
     private const val TEMPLATE_PACKAGE_NAME = "co.nimblehq.coroutine"
 
@@ -14,6 +15,8 @@ object NewProject {
     private var appName = ""
     private val appNameWithoutSpace: String
         get() = appName.replace(" ", "")
+    private val applicationClassName: String
+        get() = "${appNameWithoutSpace}Application"
     private val fileSeparator = File.separator
     private var packageName = ""
     private val projectPath: String
@@ -27,6 +30,7 @@ object NewProject {
         cleanNewProjectFolder()
         renamePackageNameFolders()
         renamePackageNameWithinFiles()
+        renameApplicationClass()
         buildProjectAndRunTests()
     }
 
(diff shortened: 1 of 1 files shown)


scripts/new_project.kts
```diff
@@ -103,6 +107,32 @@ object NewProject {
...
+                    delimiter = fileSeparator,
+                    replacement = "$applicationClassName.kt"
+                )
+                val newApplicationFile = File(newApplicationPath)
+                templateApplicationFile.renameTo(newApplicationFile)
+            }
+
+        File(projectPath)
+            ?.walk()
+            .filter { it.name.endsWith(".kt") || it.name.endsWith(".xml") }
```
manh-t, line 126: @Tuubz can we specific `AndroidManifest.xml` file like above? 🙏  is there any other `.kt` files I didn't notice here? 😁

Tuubz: Aaah, good question 👍 

There's `AndroidManifest.xml` but also the application class itself:

Screen Shot 2565-05-20 at 17 39 01

Do you think we should change it? 🙏

kaungkhantsoe (approved):
LGTM 🚀

scripts/new_project.kts
manh-t: ah ha, it's good now. Thanks for explanation 🌮

manh-t (approved):

scripts/new_project.kts
Wadeewee: @Tuubz Which file of `.kt` that we have to rename here 🙏

Wadeewee, line 126: @Tuubz Which file of .kt that we have to rename here 🙏

Tuubz: The application class file, ex: `CoroutineTemplateApplication.kt` 🚀

Wadeewee: What about in line `121` 🤔 Is it duplicate 🙏

Tuubz: Oh, I see 👍 

The first part is to rename the file name, but the second part is for inside the file:

Screen Shot 2565-05-20 at 17 39 01

Wadeewee: Oh, I got it now  🚀

Wadeewee: @Tuubz I think we can do like this :

Screen Shot 2565-05-23 at 19 15 41

What do you think? 🙏

Wadeewee (commented):
Rest LGTM 🚀

scripts/new_project.kts
Tuubz: Sounds good, I'll re-factor 🚀 

I agree, it would be better for easier understanding

Tuubz:
@Wadeewee Updated 🚀

Wadeewee (approved):

Issue #199, opened 2022-05-03
[KScript] Show error message for missing/invalid arguments


Tuubz:
### Description
Show error message for invalid or missing arguments

**NOTE**: Skip template argument as `RxJavaTemplate` is deprecated

### Acceptance Criteria
- Validate `package-name` and `app-name` arguments 
- Show error message in logs, if argument is missing or invalid
- See cases below:

**Missing package name** 
- Example: `kscript new_project.kts app-name=zzz`
- Error message: `ERROR: No package name has been provided`
- Explanation: `package-name` is required

**Invalid package name** 
- Example: `kscript new_project.kts package-name=y;y;y`
- Error message: `ERROR: Invalid package name`
- Explanation: `package-name` must follow regex pattern: `^[a-z][a-z0-9_]*(\.[a-z0-9_]+)+[0-9a-z_]$`

**Missing app name** 
- Example: `kscript new_project.kts package-name=yyy`
- Error message: `ERROR: No app name has been provided`
- Explanation: `app-name` is required

**Invalid app name** 
- Example: `kscript new_project.kts app-name=My?Project<`
- Error message: `ERROR: Invalid app name`
- Explanation: `package-name` must not include special characters

If there are any other cases, please include them here too 🚀 

### Example
```
object NewProject {

    fun generate(args: Array<String>) {
         validateArguments(args)
    }

    private fun validateArguments(args: Array<String>) {
        args.forEach {
            // validate argument and show error message if needed
        }
    }
}

NewProject.generate(args)
```

### Reference
https://github.com/nimblehq/android-templates/blob/39a3ee5fc5f1fa259661db5f26234f9f86342aaf/newproject.sh#L62-L101

PR #226, opened 2022-05-23
[#199] Show error message for missing/invalid arguments

Wadeewee:
#199

## What happened 👀

Show error message for missing/invalid arguments.

## Insight 📝

- Create `APP_PATTERN` => `^[A-Z][a-zA-Z0-9\s]*$` to validate an `app-name`.
- Create `PACKAGE_PATTERN` => `^[a-z]+(\.[a-z][a-z0-9]*)+$` to validate `package-name`.

## Proof Of Work 📹

https://user-images.githubusercontent.com/28002315/169935760-0c7edc5e-ab2e-42cf-89ec-be43087339a0.mov

https://user-images.githubusercontent.com/28002315/169935911-2f31819a-6d97-4253-918d-7dd45a4b70cf.mov

https://user-images.githubusercontent.com/28002315/169745130-5b602388-e627-413d-b208-c5d4f0cd235b.mov

https://user-images.githubusercontent.com/28002315/169745519-3bc33520-5c9b-4a82-b7a0-fa763305c9e2.mov

https://user-images.githubusercontent.com/28002315/169746184-89fa56c6-1ece-45b1-a152-5a280ebd267f.mov

### DIFF
--- a/scripts/new_project.kts
+++ b/scripts/new_project.kts
@@ -10,7 +10,7 @@ object NewProject {
     private const val TEMPLATE_FOLDER_NAME = "CoroutineTemplate"
     private const val TEMPLATE_PACKAGE_NAME = "co.nimblehq.coroutine"
 
-    private const val APP_PATTERN = "^[A-Z\$][a-zA-Z0-9\$]*\$"
+    private const val APP_PATTERN = "^[A-Z\$][a-zA-Z0-9(\\s)\$]*\$"
     private const val PACKAGE_PATTERN = "^[a-z]+(\\.[a-z][a-z0-9]*)*\$"
 
     private val modules = listOf("app", "data", "domain")
@@ -23,7 +23,7 @@ object NewProject {
         get() = "${appNameWithoutSpace}Application"
 
     private val projectPath: String
-        get() = rootPath + appName
+        get() = rootPath + appName.replace(oldValue = " ", newValue = "")
 
     private val rootPath: String
         get() = System.getProperty("user.dir").replace("scripts", "")
@@ -33,7 +33,7 @@ object NewProject {
     }
 
     private fun handleArguments(args: Array<String>) {
-        val agrumentError = "ERROR: Invalid Agrument name: Ensure define argruments => app-name={\"MyProject\"} package-name={com.sample.myproject}"
+        val agrumentError = "ERROR: Invalid Agrument name: Ensure define argruments => app-name={\"MyProject\"} or {\"My Project\"} { package-name={com.sample.myproject}"
         when (args.size) {
             1 -> when {
                 args.first().startsWith(KEY_APP_NAME) -> showMessage("ERROR: No package has been provided")
@@ -56,7 +56,7 @@ object NewProject {
         if (APP_PATTERN.toRegex().containsMatchIn(value)) {
             appName = value
         } else {
-            showMessage("ERROR: Invalid App Name: $value (needs to follow standard pattern {AppName})")
+            showMessage("ERROR: Invalid App Name: $value (needs to follow standard pattern {MyProject} or {My Project})")
         }
     }
 
--- a/scripts/new_project.kts
+++ b/scripts/new_project.kts
@@ -10,54 +10,72 @@ object NewProject {
     private const val TEMPLATE_FOLDER_NAME = "CoroutineTemplate"
     private const val TEMPLATE_PACKAGE_NAME = "co.nimblehq.coroutine"
 
+    private const val APP_PATTERN = "^[A-Z\$][a-zA-Z0-9\$]*\$"
+    private const val PACKAGE_PATTERN = "^[a-z]+(\\.[a-z][a-z0-9]*)*\$"
+
     private val modules = listOf("app", "data", "domain")
+    private val fileSeparator = File.separator
 
     private var appName = ""
-    private val appNameWithoutSpace: String
-        get() = appName.replace(" ", "")
+    private var packageName = ""
+
     private val applicationClassName: String
         get() = "${appNameWithoutSpace}Application"
-    private val fileSeparator = File.separator
-    private var packageName = ""
+
     private val projectPath: String
-        get() = rootPath + appNameWithoutSpace
+        get() = rootPath + appName
+
     private val rootPath: String
         get() = System.getProperty("user.dir").replace("scripts", "")
 
     fun generate(args: Array<String>) {
         handleArguments(args)
-        initializeNewProjectFolder()
-        cleanNewProjectFolder()
-        renamePackageNameFolders()
-        renamePackageNameWithinFiles()
-        renameApplicationClass()
-        buildProjectAndRunTests()
     }
 
     private fun handleArguments(args: Array<String>) {
-        args.forEach {
-            val (key, value) = it.split(ARGUMENT_DELIMITER)
-            when (key) {
-                KEY_APP_NAME -> appName = value
-                KEY_PACKAGE_NAME -> packageName = value
+        val agrumentError = "ERROR: Invalid Agrument name: Ensure define argruments => app-name={\"MyProject\"} package-name={com.sample.myproject}"
+        when (args.size) {
+            1 -> when {
+                args.first().startsWith(KEY_APP_NAME) -> showMessage("ERROR: No package has been provided")
+                args.first().startsWith(KEY_PACKAGE_NAME) -> showMessage("ERROR: No app name has been provided")
+                else -> showMessage(agrumentError)
             }
+            2 -> args.map { arg ->
+                val (key, value) = arg.split(ARGUMENT_DELIMITER)
+                when (key) {
+                    KEY_APP_NAME -> validAppName(value)
+                    KEY_PACKAGE_NAME -> validPackageName(value)
+                    else -> showMessage(agrumentError)
+                }.also { executeNextSteps() }
+            }
+            else -> showMessage("ERROR: Require app-name and package-name to initialize the new project")
         }
     }
 
-    private fun initializeNewProjectFolder() {
-        showMessage("=> 🐢 Initializing new project...")
-        copyFiles(fromPath = rootPath + TEMPLATE_FOLDER_NAME, toPath = projectPath)
-        // Set gradlew file as executable, because copying files from one folder to another doesn't copy file permissions correctly (= read, write & execute).
-        File(projectPath + fileSeparator + "gradlew")?.setExecutable(true)
+    private fun validAppName(value: String) {
+        if (APP_PATTERN.toRegex().containsMatchIn(value)) {
+            appName = value
+        } else {
+            showMessage("ERROR: Invalid App Name: $value (needs to follow standard pattern {AppName})")
+        }
     }
 
-    private fun cleanNewProjectFolder() {
-        executeCommand("sh $projectPath${fileSeparator}gradlew -p $projectPath clean")
-        executeCommand("sh $projectPath${fileSeparator}gradlew -p $projectPath${fileSeparator}buildSrc clean")
-        listOf(".idea", ".gradle", "buildSrc$fileSeparator.gradle", ".git").forEach {
-            File("$projectPath$fileSeparator$it")?.let { targetFile ->
-                targetFile.deleteRecursively()
-            }
+    private fun validPackageName(value: String) {
+        if (PACKAGE_PATTERN.toRegex().containsMatchIn(value)) {
+            packageName = value
+        } else {
+            showMessage("ERROR: Invalid Package Name: $value (needs to follow standard pattern {com.example.package})")
+        }
+    }
+
+    private fun executeNextSteps() {
+        if (appName.isNotEmpty() && packageName.isNotEmpty()) {
+            initializeNewProjectFolder()
+            cleanNewProjectFolder()
+            renamePackageNameFolders()
+            renamePackageNameWithinFiles()
+            renameApplicationClass()
+            buildProjectAndRunTests()
         }
     }
 
@@ -171,6 +189,10 @@ object NewProject {
         var sourceText = sourceFile.readText()
         sourceText = sourceText.replace(oldValue, newValue)
         sourceFile.writeText(sourceText)
+    }
+
+    private fun showMessage(message: String) {
+        println("\n${message}\n")
     }
 }
 
--- a/scripts/new_project.kts
+++ b/scripts/new_project.kts
@@ -10,8 +10,8 @@ object NewProject {
     private const val TEMPLATE_FOLDER_NAME = "CoroutineTemplate"
     private const val TEMPLATE_PACKAGE_NAME = "co.nimblehq.coroutine"
 
-    private const val PATTERN_APP = "^[A-Z\$][a-zA-Z0-9(\\s)\$]*\$"
-    private const val PATTERN_PACKAGE = "^[a-z]+(\\.[a-z][a-z0-9]*)*\$"
+    private const val PATTERN_APP = "^[A-Z][a-zA-Z0-9\\s]*$"
+    private const val PATTERN_PACKAGE = "^[a-z]+(\\.[a-z][a-z0-9]*)+$"
 
     private val modules = listOf("app", "data", "domain")
     private val fileSeparator = File.separator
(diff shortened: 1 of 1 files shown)


scripts/new_project.kts
```diff
@@ -9,23 +9,67 @@ object NewProject {
...
+            }
+            else -> showMessage("ERROR: Require app-name and package-name to initialize the new project")
+        }
+    }
+
+    private fun validAppName(value: String) {
+        if (APP_PATTERN.toRegex().containsMatchIn(value)) {
+            appName = value
+        } else {
+            showMessage("ERROR: Invalid App Name: $value (needs to follow standard pattern {AppName})")
```
kaungkhantsoe, line 54: I think we should terminate the process here with `System.exit(0)`. What do you think?

scripts/new_project.kts
```diff
@@ -9,23 +9,67 @@ object NewProject {
...
+        }
+    }
+
+    private fun validPackageName(value: String) {
+        if (PACKAGE_PATTERN.toRegex().containsMatchIn(value)) {
+            packageName = value
+        } else {
+            showMessage("ERROR: Invalid Package Name: $value (needs to follow standard pattern {com.example.package})")
```
kaungkhantsoe, line 62: Ditto here

scripts/new_project.kts
kaungkhantsoe: Also, current implementation doesn't allow running the script as `app-name='My App'`. Please correct me if I am wrong, we allow `'My App'` as a valid `app-name` before right? 🙏 cc @Tuubz

scripts/new_project.kts
kaungkhantsoe: `validPackageName` doesn't show the error if we run like `package-name=comtest.mytest` 🙏

scripts/new_project.kts
Tuubz: Correct, we should still allow spaces in `app-name` and surround it with quotes 🙏 

Reference: https://nimble-co.slack.com/archives/C03DFQD942W/p1652084187174279

scripts/new_project.kts
Tuubz: If I'm not mistaken, a package name with two packages is allowed in the previous script 💭  

So I think it's correct to not show an error for this case, what do you think? 🙏

manh-t: @Tuubz @kaungkhantsoe according to [the documentation](https://developer.android.com/studio/build/configure-app-module#set-application-id), `It must have at least two segments (one or more dots).` 😄

kaungkhantsoe: I see. Then it is okay 🚀

scripts/new_project.kts
Wadeewee: Ah, I remembered @kaungkhantsoe @Tuubz , Will update 🙏

Wadeewee: > I think we should terminate the process here with System.exit(0). What do you think?

@kaungkhantsoe I don't think we need to terminate the process because we didn't start the process yet. We will execute the next step**(e.g. initialize, clean package)** if all of the validation is valid and then start the process 🙏 cc @Tuubz @manh-t

scripts/new_project.kts
```diff
@@ -9,23 +9,67 @@ object NewProject {
     private const val TEMPLATE_FOLDER_NAME = "CoroutineTemplate"
     private const val TEMPLATE_PACKAGE_NAME = "co.nimblehq.coroutine"
 
+    private const val APP_PATTERN = "^[A-Z\$][a-zA-Z0-9(\\s)\$]*\$"
+    private const val PACKAGE_PATTERN = "^[a-z]+(\\.[a-z][a-z0-9]*)*\$"
```
Tuubz, line 13: Should we rename it like this, for better grouping? 🙏 
```suggestion
    private const val PATTERN_APP = "^[A-Z\$][a-zA-Z0-9(\\s)\$]*\$"
    private const val PATTERN_PACKAGE = "^[a-z]+(\\.[a-z][a-z0-9]*)*\$"
```

scripts/new_project.kts
Tuubz: Hmmm, the process starts as soon as we run the script

I do agree with @kaungkhantsoe's idea, to stop the script/process as soon as there is an error.

Then we don't need this check:

``` private fun executeNextSteps() {
        if (appName.isNotEmpty() && packageName.isNotEmpty()) {
           ....
        }
    }
```

@Wadeewee Did we miss something here? 🙇

scripts/new_project.kts
```diff
@@ -9,23 +9,67 @@ object NewProject {
...
+
     private val modules = listOf("app", "data", "domain")
 
     private var appName = ""
-    private val appNameWithoutSpace: String
-        get() = appName.replace(" ", "")
```
Tuubz, line 16: Could we please rebase? I don't think we can remove this, as we require it for renaming the application class 🙏

scripts/new_project.kts
```diff
@@ -9,23 +9,67 @@ object NewProject {
...
+                val (key, value) = arg.split(ARGUMENT_DELIMITER)
+                when (key) {
+                    KEY_APP_NAME -> validAppName(value)
+                    KEY_PACKAGE_NAME -> validPackageName(value)
+                    else -> showMessage(agrumentError)
+                }.also { executeNextSteps() }
+            }
+            else -> showMessage("ERROR: Require app-name and package-name to initialize the new project")
+        }
+    }
```
manh-t, line 48: @Wadeewee to simplify the condition and avoid nested conditions, I think we can show the error if `app-name` or `package-name` is not provided, no need for a specific check for each case. What do you think? cc @Tuubz @kaungkhantsoe 
Something like:
```suggestion
    private fun handleArguments(args: Array<String>) {
        args.forEach {
              val (key, value) = it.split(ARGUMENT_DELIMITER)
              when (key) {
                    KEY_APP_NAME -> appName = value
                    KEY_PACKAGE_NAME -> packageName = value
              }
        }
        if (appName.isEmpty() or packageName.isEmpty()) {
            showMessage("ERROR: Invalid Agrument name: Ensure define argruments => app-name={\"MyProject\"} or {\"My Project\"} { package-name={com.sample.myproject}")
            // exit process
        }
        validateAppName(appName)
        validatePackageName(packageName)
    }
```

scripts/new_project.kts
```diff
@@ -9,23 +9,67 @@ object NewProject {
...
+
+    private fun validAppName(value: String) {
```
manh-t, line 50: please define a method with verb 😄 
```suggestion
    private fun validateAppName(value: String) {
```

scripts/new_project.kts
```diff
@@ -9,23 +9,67 @@ object NewProject {
...
+        if (APP_PATTERN.toRegex().containsMatchIn(value)) {
+            appName = value
+        } else {
+            showMessage("ERROR: Invalid App Name: $value (needs to follow standard pattern {MyProject} or {My Project})")
+        }
+    }
+
+    private fun validPackageName(value: String) {
```
manh-t, line 58: same here
```suggestion
    private fun validatePackageName(value: String) {
```

manh-t (commented):
@Wadeewee please rebase to the latest change 🙋

scripts/new_project.kts
Tuubz: @Wadeewee Something like this is what I had in mind 🙇 

Hopefully, I didn't miss any cases 

```
fun generate(args: Array<String>) {
    handleArguments(args)
    initializeNewProjectFolder()
    cleanNewProjectFolder()
    renamePackageNameFolders()
    renamePackageNameWithinFiles()
    renameApplicationClass()
    buildProjectAndRunTests()
}

    if (args.size == 0) showErrorMessage("ERROR: No arguments")
    if (args.size > 2) showErrorMessage("ERROR: Too many arguments")
    args.forEach {
        val (key, value) = it.split(ARGUMENT_DELIMITER)
        when (key) {
            KEY_APP_NAME -> validateAppName(value)
            KEY_PACKAGE_NAME -> validatePackageName(value)
            else -> showErrorMessage("ERROR: Invalid argument")
        }
    }
}

private fun validateAppName(value: String) {
    if (value.isEmpty()) {
        showErrorMessage("ERROR: Missing App Name")
    } else if (!PATTERN_APP.toRegex().matches(value)) {
        showErrorMessage("ERROR: Invalid App Name")
    } else {
        appName = value
    }
}

private fun validatePackageName(value: String) {
    if (value.isEmpty()) {
        showErrorMessage("ERROR: Missing Package Name")
    } else if (!PATTERN_PACKAGE.toRegex().matches(value)) {
        showErrorMessage("ERROR: Invalid Package Name")
    } else {
        packageName = value
    }
}

private fun showErrorMessage(message: String, exitCode: Int = 0) {
    println("\n${message}\n")
    System.exit(exitCode)
}
```

scripts/new_project.kts
```diff
@@ -10,37 +10,75 @@ object NewProject {
     private const val TEMPLATE_FOLDER_NAME = "CoroutineTemplate"
     private const val TEMPLATE_PACKAGE_NAME = "co.nimblehq.coroutine"
 
+    private const val PATTERN_APP = "^[A-Z\$][a-zA-Z0-9(\\s)\$]*\$"
+    private const val PATTERN_PACKAGE = "^[a-z]+(\\.[a-z][a-z0-9]*)*\$"
```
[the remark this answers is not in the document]
Wadeewee: @Tuubz It shorter than pattern in `newproject.sh`,  Please let me know if it doesn't work some case  🙏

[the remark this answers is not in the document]
Tuubz: Aah, I see. 👍 

Unfortunately, the current reg-expattern allows package-name with only one package, but the minimum should be two

Ex: 
- `package-name=test` (should not be allowed)
- `package-name=test.example` (should be allowed)

What do you think? 🙏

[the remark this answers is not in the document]
Wadeewee: @Tuubz May be we need to update pattern a bit, Could you please try this `^[a-z]+(\.[a-z][a-z0-9]*)+$` for me  instead of `^[a-z]+(\\.[a-z][a-z0-9]*)*\$`🙏

scripts/new_project.kts
```diff
@@ -10,37 +10,75 @@ object NewProject {
     private const val TEMPLATE_FOLDER_NAME = "CoroutineTemplate"
     private const val TEMPLATE_PACKAGE_NAME = "co.nimblehq.coroutine"
 
+    private const val PATTERN_APP = "^[A-Z\$][a-zA-Z0-9(\\s)\$]*\$"
```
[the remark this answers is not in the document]
Wadeewee: @Tuubz If we don't have `\` if will be just `string` in Kotlin?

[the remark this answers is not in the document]
Wadeewee: > If we don't need it, I think the following is sufficient: ^[a-zA-Z0-9\\s]*$

The first character should be Uppercase 🤔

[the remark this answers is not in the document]
Tuubz: Actually, something like `toyota-android-wallet` should also be allowed 💭 I just tried with the old script, but then the application class name will be renamed to `toyota-android-walletApplication` 🙈 

I think let's require uppercase for now, later we can make adjustments to allow cases like `toyota-android-wallet`.

What do you think? 🙏

scripts/new_project.kts
```diff
@@ -10,37 +10,75 @@ object NewProject {
...
+    private fun validateAppName(value: String) {
+        if (PATTERN_APP.toRegex().containsMatchIn(value)) {
+            appName = value
+        } else {
+            showMessage("ERROR: Invalid App Name: $value (needs to follow standard pattern {MyProject} or {My Project})")
+        }
+    }
+
+    private fun validatePackageName(value: String) {
+        if (PATTERN_PACKAGE.toRegex().containsMatchIn(value)) {
```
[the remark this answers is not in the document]
Wadeewee: Good Question @Tuubz 👍  Here's a differ:
Screen Shot 2565-05-25 at 11 23 07
 
 It works on both function but seems like `matches` is safer to check the whole string 🙏

scripts/new_project.kts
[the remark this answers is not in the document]
Tuubz: > @Tuubz If we don't have \ it will be just string in Kotlin?

Sorry, I don't think I explained myself well enough

From my understanding `\$` means that `$` is allowed as character, but that's not the case, right? 🙏 

So I think the end result should be `"^[A-Z][a-zA-Z0-9\\s]*$"`, please correct me if I'm wrong 🙇

scripts/new_project.kts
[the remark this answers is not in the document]
Tuubz: Nice, that works for me 🚀 

If I'm not mistaken, we could minimize it even a little bit more: `"^[a-z]+(\\.[a-z][a-z0-9])+$"`

scripts/new_project.kts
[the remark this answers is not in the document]
Wadeewee: > From my understanding \$ means that $ is allowed as character, but that's not the case, right? 🙏

@Tuubz You mean (`s` or `S`) not `$` right? 
Screen Shot 2565-05-25 at 13 42 10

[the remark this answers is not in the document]
Tuubz: I see that you also mentioned `\\s`, this part is correct 👍 

But I'm referring to these:

Screen Shot 2565-05-25 at 14 44 10

From my understanding, they should be removed. Because this shouldn't be allowed `app-name="Test($)"` 🙏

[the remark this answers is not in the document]
Wadeewee: >Actually, something like toyota-android-wallet should also be allowed 💭 I just tried with the old script, but then the application class name will be renamed to toyota-android-walletApplication 🙈 I think let's require uppercase for now, later we can make adjustments to allow cases like toyota-android-wallet.
What do you think? 🙏

Yeah, I agreed, Let require uppercase for now and you can create the issue to solved this later if need  🚀

scripts/new_project.kts
manh-t: @Wadeewee I think we should exit the process when there is an error, so there is only 1 error that will display 😁  What do you think?
@Tuubz I know that this implementation is slightly different from the AC, but is there any chance to update the AC as I mention above? 🙇 `no need for a specific check for each case`

Tuubz: @manh-t I agree with you Manh, if you scroll up I also suggested a solution where we exit the process ☝️ 

Anyways, if it's different from the AC, I'm fine with that 👍

scripts/new_project.kts
Wadeewee: @Tuubz Missing this case, when putting only `1` argument e.g `kscript new_project.kts app-name='MyProject')` or `kscript new_project.kts package-name=com.example.myproject`

So I think let's keep the previous one for checking size and then validate the argument, Also by this way we no need to handle the empty cases in each validate function as well. 

What do you think? 🙏

scripts/new_project.kts
```diff
@@ -10,17 +10,24 @@ object NewProject {
     private const val TEMPLATE_FOLDER_NAME = "CoroutineTemplate"
     private const val TEMPLATE_PACKAGE_NAME = "co.nimblehq.coroutine"
 
+    private const val PATTERN_APP = "^[A-Z][a-zA-Z0-9(\\s)]*$"
```
Tuubz, line 13: We should not allow `(` & `)`, right? 🙏 

```suggestion
    private const val PATTERN_APP = "^[A-Z][a-zA-Z0-9\\s]*$"
```

scripts/new_project.kts
```diff
@@ -10,17 +10,24 @@ object NewProject {
...
+    private const val PATTERN_PACKAGE = "^[a-z]+(\\.[a-z][a-z0-9]*)+\$"
```
Tuubz, line 14: Redundant `\`, I think? 😁 

```suggestion
    private const val PATTERN_PACKAGE = "^[a-z]+(\\.[a-z][a-z0-9]*)+$"
```

scripts/new_project.kts
```diff
@@ -35,12 +42,38 @@ object NewProject {
     }
 
     private fun handleArguments(args: Array<String>) {
-        args.forEach {
-            val (key, value) = it.split(ARGUMENT_DELIMITER)
-            when (key) {
-                KEY_APP_NAME -> appName = value
-                KEY_PACKAGE_NAME -> packageName = value
+        val argumentError = "ERROR: Invalid argument name: Ensure define arguments => app-name={\"MyProject\"} or {\"My Project\"} { package-name={com.sample.myproject}"
```
Tuubz, line 45: Redundant `{` 🚀 

```suggestion
        val argumentError = "ERROR: Invalid argument name: Ensure define arguments => app-name={\"MyProject\"} or {\"My Project\"} package-name={com.sample.myproject}"
```

scripts/new_project.kts
```diff
@@ -35,12 +42,38 @@ object NewProject {
...
+        when (args.size) {
+            1 -> when {
+                args.first().startsWith(KEY_APP_NAME) -> showErrorMessage("ERROR: No package name has been provided")
+                args.first().startsWith(KEY_PACKAGE_NAME) -> showErrorMessage("ERROR: No app name has been provided")
+                else -> showErrorMessage(argumentError)
+            }
+            2 -> args.map { arg ->
```
Tuubz, line 52: Would `forEach` be more correct? 🙏 

```suggestion
            2 -> args.forEach { arg ->
```

Tuubz (commented):
Rest LGTM 🚀

scripts/new_project.kts
```diff
@@ -161,8 +190,7 @@ object NewProject {
         process.inputStream.reader().forEachLine { println(it) }
         val exitValue = process.waitFor()
         if (exitValue != 0) {
-            showMessage("❌ Something went wrong!")
-            System.exit(exitValue)
+            showErrorMessage("❌ Something went wrong!")
```
Tuubz, line 193: Woops, sorry last one! 🙇 

```suggestion
            showErrorMessage("❌ Something went wrong!", exitValue)
```

Tuubz (approved):
🚀 🚀 🚀

kaungkhantsoe (approved):
LGTM 🚀

manh-t (approved):

PR #245, opened 2022-06-23
[#100] Add new workflow to generate release note and version tag automatically

### LINKED ISSUE #100: Add new workflow to generate Release note and version tag automatically
## Why

The template needs to be released every month so adding a new workflow to handle release notes and version tag on every new releases

## Who Benefits?

Android chapter

AVI5HEK:
https://github.com/nimblehq/android-templates/issues/100

## What happened 👀

Add a workflow to automatically generate a release note and tag when pushing changes to `main`.

## Insight 📝

- Added `release-drafter.yml` to be triggered when pushing changes to `main`.
- Used [release-drafter/release-drafter](https://github.com/release-drafter/release-drafter) action to generate a draft release. The content would be the same as our current format. Please suggest emojis if you want them in the content.
> Note: the config file _must_ be on the default branch, i.e, `develop` in our case.
- Added `get_version.kts` to get the version from the commit message passed as an argument. Special thanks to @Tuubz for helping with KScript.

## Proof Of Work 📹

image

Closes #100

### DIFF
--- a/.github/release-drafter-config.yml
+++ b/.github/release-drafter-config.yml
@@ -4,11 +4,11 @@ tag-template: '$RESOLVED_VERSION'
 
 categories:
   - title: 'Features'
-    label: 'feature'
+    label: 'type : feature'
   - title: 'Chores'
-    label: 'chore'
+    label: 'type : chore'
   - title: 'Bugs'
-    label: 'bug'
+    label: 'type : bug'
 
 change-template: '- $TITLE'
 
--- a/.github/release-drafter-config.yml
+++ b/.github/release-drafter-config.yml
@@ -3,11 +3,11 @@ name-template: '$RESOLVED_VERSION'
 tag-template: '$RESOLVED_VERSION'
 
 categories:
-  - title: 'Features'
+  - title: '⭐ Features'
     label: 'type : feature'
-  - title: 'Chores'
+  - title: '🧹 Chores'
     label: 'type : chore'
-  - title: 'Bugs'
+  - title: '🐞 Bugs'
     label: 'type : bug'
 
 change-template: '- $TITLE'
--- a/.github/release-drafter-config.yml
+++ b/.github/release-drafter-config.yml
@@ -1,0 +1,16 @@
+name-template: '$RESOLVED_VERSION'
+
+tag-template: '$RESOLVED_VERSION'
+
+categories:
+  - title: 'Features'
+    label: 'feature'
+  - title: 'Chores'
+    label: 'chore'
+  - title: 'Bugs'
+    label: 'bug'
+
+change-template: '- $TITLE'
+
+template: |
+  $CHANGES
--- a/.github/workflows/draft-new-release.yml
+++ b/.github/workflows/draft-new-release.yml
@@ -1,4 +1,4 @@
-name: Draft a new release
+name: Draft new release
 
 on:
   push:
@@ -9,13 +9,15 @@ permissions:
   contents: read
 
 jobs:
-  update_release_draft:
+  draft_new_release:
     permissions:
       contents: write
     runs-on: ubuntu-latest
+    timeout-minutes: 30
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
+
       - name: Install Kscript
         run: |
           curl -s "https://get.sdkman.io" | bash
@@ -23,12 +25,14 @@ jobs:
           sdk install kotlin 1.6.21
           sdk install kscript 4.0.3
           echo $PATH >> $GITHUB_PATH
+
       - name: Set version
         working-directory: scripts
         run: |
           COMMIT_MESSAGE="${{ github.event.head_commit.message }}"
           VERSION=$(kscript get_version.kts "$COMMIT_MESSAGE")
           echo "VERSION=$VERSION" >> $GITHUB_ENV
+
       - name: Draft release
         uses: release-drafter/release-drafter@v5
         with:
--- a/.github/workflows/draft-new-release.yml
+++ b/.github/workflows/draft-new-release.yml
@@ -26,7 +26,7 @@ jobs:
           sdk install kscript 4.0.3
           echo $PATH >> $GITHUB_PATH
 
-      - name: Set version
+      - name: Get version from the latest commit message
         working-directory: scripts
         run: |
           COMMIT_MESSAGE="${{ github.event.head_commit.message }}"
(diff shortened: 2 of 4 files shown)


Tuubz (commented):
🎁

.github/release-drafter-config.yml
```diff
@@ -0,0 +1,16 @@
+name-template: '$RESOLVED_VERSION'
+
+tag-template: '$RESOLVED_VERSION'
+
+categories:
+  - title: 'Features ⭐'
+    label: 'type : feature'
+  - title: 'Chores 🧹'
+    label: 'type : chore'
+  - title: 'Bugs 🐞'
```
Tuubz, line 10: Optional: Put the emojis in front

Hahahh, sorry. Just think it would look better 😂

Tuubz (commented):
Rest LGTM 🔥 🔥 🔥 🔥

scripts/get_version.kts
```diff
@@ -0,0 +1,15 @@
+object GetVersion {
+
+    private const val PATTERN_VERSION = "[0-9]+.[0-9]+.[0-9]+"
```
ryan-conway, line 3: Isn't `.` a meta character meaning `any character`? Looks like this would also match `1a3b3` 😅 Escaping the `.` or wrapping it in `[]` will match a `.` character 😄 
```suggestion
    private const val PATTERN_VERSION = "[0-9]+\\.[0-9]+\\.[0-9]+"
```
```suggestion
    private const val PATTERN_VERSION = "[0-9]+[.][0-9]+[.][0-9]+"
```

scripts/get_version.kts
```diff
@@ -0,0 +1,15 @@
...
+        return PATTERN_VERSION
+            .toRegex()
+            .find(commitMessage)
+            ?.value
+    }
+}
+
+val version = GetVersion.search(commitMessage = args.first())
+
+println(version)
```
lydiasama, line 15: What if the version is null? Shall we default the empty string in this case?
```suggestion
println(version.orEmpty())
```

AVI5HEK: @lydiasama if `version` is `null`, release-drafter resolves the version automatically by incrementing the _patch_ from the last release.

Tuubz: @AVI5HEK I could be wrong, but I think it will print "null" in this case 🙏

lydiasama: @AVI5HEK I meant in case `GetVersion.search(commitMessage = args.first())` returns `null`. 🙏🏻

luongvo: @AVI5HEK @Tuubz can we have unit test for this? 🤔

luongvo (commented):
rest lgtm

.github/workflows/draft-new-release.yml
```diff
@@ -0,0 +1,42 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Install Kscript
+        run: |
+          curl -s "https://get.sdkman.io" | bash
+          source "$HOME/.sdkman/bin/sdkman-init.sh"
+          sdk install kotlin 1.6.21
+          sdk install kscript 4.0.3
```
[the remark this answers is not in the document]
Tuubz: Agree, I'll create an issue for this soon

Would you agree it's not a big priority atm?  🙏

.github/workflows/draft-new-release.yml
```diff
@@ -0,0 +1,42 @@
+name: Draft new release
+
+on:
+  push:
+    branches:
+      - main
+
+permissions:
+  contents: read
```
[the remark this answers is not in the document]
Tuubz: We added this because it's also specified in the example: https://github.com/release-drafter/release-drafter#usage 🚀 

Might be safer to keep it, I'd say

.github/workflows/draft-new-release.yml
```diff
@@ -0,0 +1,42 @@
...
+
+jobs:
+  draft_new_release:
+    permissions:
+      contents: write
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
```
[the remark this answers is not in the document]
Tuubz: That's great 🚀  

I think we could have a separate ticket to change it in all workflows, what do you think?

scripts/get_version.kts
Tuubz: That's actually a good idea 👍 

@AVI5HEK Do you think you can add this by end of today? If not, let's skip it for now

.github/workflows/draft-new-release.yml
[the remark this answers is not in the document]
luongvo: @Tuubz not a big deal, agree 👍

.github/workflows/draft-new-release.yml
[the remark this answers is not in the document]
luongvo: Please create it soon then we can skip it in this PR 🙏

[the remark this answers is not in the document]
Tuubz: Done: https://github.com/nimblehq/android-templates/issues/247 ✅

scripts/get_version.kts
AVI5HEK: @Tuubz @luongvo not sure where and how to write the unit test for this. The file is outside of both the projects. Any insight regarding where to begin would be much appreciated. 🙏🏽

AVI5HEK: @Tuubz 
>I think it will print "null" in this case
That's correct. And if it's null, release-drafter will automatically resolve the version. We can edit the version (and tag) from the draft later on.

.github/workflows/draft-new-release.yml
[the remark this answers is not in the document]
AVI5HEK: So should I change it to v3 in this PR? @Tuubz @luongvo

scripts/get_version.kts
AVI5HEK: @Tuubz that's what I found while testing on my repo. So if somehow the script fails to get the version from the commit message, i.e, `println(null)`, `release-drafter` will set the version to `3.8.1`, I believe. It incremented the _patch_ value in my case.

Tuubz (approved):
🎸 🎸 🎸

scripts/get_version.kts
luongvo: @AVI5HEK Thank you for your insights. I think we can add tests later as we don't have ideas to do it now 😄

luongvo (approved):
I think it's good to go 👨‍🎤

ryan-conway (approved):

Issue #217, opened 2022-05-10
RFC: Increase the connection timeout when making API request


manh-t:
## Issue

Our `CoroutineTemplate` has integrated Proxyman config which helps to update request/ response in API. But I'm facing the `SocketTimeoutException` when updating the request/ response via Proxyman.

## Solution

- Increase the timeout when making API requests so we have enough time to modify the payload.
- They are: `connectTimeout`, `readTimeout`, `writeTimeout` of `OkHttpClient`.
- Recommend increasing the timeout to `30 seconds`.

## Who Benefits?

Developers.

I think this is **essential**.

## What's Next?

- [ ] Increase the timeout for `connectTimeout`, `readTimeout` and `writeTimeout`.

luongvo:
Voted 👍

hoangnguyen92dn:
+1 👍

Tuubz:
Approved 🚀 

Something extra to think about: Do we consider Proxyman as **essential** or **optional**?

manh-t:
@Tuubz I think we should consider Proxyman as `essential` since we use it in most of our projects 🥼 You have integrated Proxyman in https://github.com/nimblehq/android-templates/pull/118 @Tuubz 😃

Tuubz:
@manh-t Correct, I integrated Proxyman before we implemented the idea of essential/optional features 🏑 

A while back [not everyone in the team agreed that Proxyman should be essential](https://nimble-co.slack.com/archives/G41UCTGHL/p1642475278000300?thread_ts=1642413515.010600&cid=G41UCTGHL)

Tuubz:
@thiennguyen0196 

> Increase the timeout when making API requests so we have enough time to modify the payload.

Just to clarify a bit here, whenever we use the "**breakpoint**" feature of Proxyman it cancels the breakpoint after 10 seconds. This is too little time for us to modify the JSON response, that's why we would like to increase it to 30 seconds 🙏 Would be nice if we could keep it at 10 seconds for release variants, but have it at 30 seconds for debug variants 💭 

Anyways, if you get stuck during the set-up of Proxyman. Please let me know, I'll help out 😁

thiennguyen0196:
@Tuubz I will check it out and let you know if I stumble any problems 👌 I have got the idea of this issue and I will try to solve it. Many thanks for the presentation file you sent me, it helps a lot 😆

luongvo:
[...]

PR #250, opened 2022-07-01
[#217] Increase timeout in OkHttpClient to 30 seconds

thiennguyen0196:
https://github.com/nimblehq/android-templates/issues/217

## What happened 👀

Increase the `readTimeout` to at least 30 seconds with debug mode so that we can update the request/ response via Proxyman easier, because default they are 10 seconds which are too short

## Insight 📝

- Run example app `CoroutinesTemplate`
- Set breakpoint proxyman to `https://jsonplaceholder.typicode.com/users`
- Let the app run request `https://jsonplaceholder.typicode.com/users`
- Edit the response body to test it and wait for 10 seconds or more
- Press `execute` button on proxy man breakpoint for it to return edited response
- The response should be 200 success instead of `SocketTimeoutException`

## Proof Of Work 📹

https://user-images.githubusercontent.com/25218255/176863712-378d682a-e7ce-4875-ab35-2dc6c9dbbdf3.mov

### DIFF
--- a/CoroutineTemplate/app/src/main/java/co/nimblehq/coroutine/di/modules/OkHttpClientModule.kt
+++ b/CoroutineTemplate/app/src/main/java/co/nimblehq/coroutine/di/modules/OkHttpClientModule.kt
@@ -7,6 +7,9 @@ import dagger.hilt.InstallIn
 import dagger.hilt.components.SingletonComponent
 import okhttp3.OkHttpClient
 import okhttp3.logging.HttpLoggingInterceptor
+import java.util.concurrent.TimeUnit
+
+private const val READ_TIME_OUT = 30L
 
 @Module
 @InstallIn(SingletonComponent::class)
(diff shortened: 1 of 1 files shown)


CoroutineTemplate/app/src/main/java/co/nimblehq/coroutine/di/modules/OkHttpClientModule.kt
```diff
@@ -18,6 +23,10 @@ class OkHttpClientModule {
             addInterceptor(HttpLoggingInterceptor().apply {
                 level = HttpLoggingInterceptor.Level.BODY
             })
+            connectTimeout(CONNECT_TIME_OUT, TimeUnit.SECONDS)
+            writeTimeout(WRITE_TIME_OUT, TimeUnit.SECONDS)
```
Tuubz, line 27: Could it be we don't need this? I'm thinking if `readTimeout` is sufficient 🙏

thiennguyen0196: You are correct, let me delete redundant time out
The `readTimeout` should be enough

luongvo (approved):

Tuubz (commented):
Rest LGTM 🚀 🚀 🚀

Tuubz (approved):
Rest LGTM 🚀

Issue #244, opened 2022-06-17
[Sample] Rename `CoroutineTemplate` to `template`


Tuubz:
**Description**

Rename `CoroutineTemplate` to `template`, also update its references.

**Acceptance Criteria**
- Rename `CoroutineTemplate` to `template`
- Update `README.md`
- Update `verify_newproject_script.yml`
- Update `review_pull_request.yml`
- Update `run_detekt_and_unit_tests.yml`

PR #227, opened 2022-05-23
[#207] Rename app name in strings.xml

Wadeewee:
#207

## What happened 👀

Rename app name within all strings.xml files inside the project folder according to app-name.

## Proof Of Work 📹

https://user-images.githubusercontent.com/28002315/169774187-3b4967a9-0cd6-4386-b6cd-26f0abcd85ad.mov

PR #258, opened 2022-07-22
[#244] Rename `CoroutineTemplate` to `template`

Tuubz:
#244 

## What happened 👀

- Rename `CoroutineTemplate` folder to `template` (main folder)
- Rename `CoroutineTemplateApplication` to `TemplateApplication` (application class)
- Rename `co.nimblehq.coroutine` to `co.nimblehq.template`(package)
- Rename `"Coroutine Template"` to `"Template"` (`strings.xml`)
- Rename `CoroutineTemplate` to `template` inside workflows, documentation, ...

## Proof Of Work 📹
- CI is still working as expected ✅ 
- App is still buildable ✅ 
- Script is still working correctly ✅


.github/workflows/review_pull_request.yml
```diff
@@ -29,16 +29,16 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-gradle-
 
-      - name: Run Detekt on CoroutineTemplate
-        working-directory: ./CoroutineTemplate
+      - name: Run Detekt on template
```
Wadeewee, line 32: Should it be Upper Camel Case 🤔 
```suggestion
      - name: Run Detekt on Template
```

Wadeewee (commented):
Just a small concern for the rest LGTM 🚀

.github/workflows/review_pull_request.yml
Tuubz: @Wadeewee  I'm thinking to keep it lowercase, then it's aligned with the folder name  -> `template`

What do you think? I'm fine with both approaches 🙏

Wadeewee: @Tuubz I'm not sure, too, but I'm similarly using lowercase. It will be something like this pattern `android-template. `🙏 Actually, I'm okay with both too.

Wadeewee (approved):

chornerman:
Hi, @Tuubz I wonder why the title of the PR tag #224, but the commits tag #227 🤔

Tuubz:
@chornerman Oh damn, I messed that one up badly 🙈 Thank you

chornerman (approved):
LGTM 🚀

lydiasama (approved):

minhnimble (approved):

PR #260, opened 2022-07-29
[#234] Add help function for kscript

### LINKED ISSUE #234: Add help function for initiating new project using kscript
## Why

Old bash script have `--help` function which prints the instruction of how to use the script to create a new project but new script doesn't have one.

## Who Benefits?

Developers who use the script to create a new project.

kaungkhantsoe:
Closes #234

## What happened 👀

Old bash script have --help function which prints the instruction of how to use the script to create a new project but new script doesn't have one.

## Insight 📝

Add additional parameter and check if the passed parameter is `help` and show message

## Proof Of Work 📹
Screen Shot 2022-07-29 at 4 31 55 PM

Screen Shot 2022-08-03 at 4 00 23 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,8 +17,8 @@ Our Android template: **[template](https://github.com/nimblehq/android-templates
 2. Install [Kscript](https://github.com/holgerbrandl/kscript#installation)
 3. Run `kscript new_project.kts` to create a new project with the following arguments:
   ```   
-    package-name=                      New package name (i.e. com.example.package)
-    app-name=                          New app name (i.e. MyApp, "My App")
+    package-name=                      New package name (i.e., com.example.package)
+    app-name=                          New app name (i.e., MyApp, "My App", "my-app")
   ```
 
   Example: `kscript new_project.kts package-name=co.myproject.example app-name="My Project"`
--- a/scripts/new_project.kts
+++ b/scripts/new_project.kts
@@ -73,7 +73,7 @@ Example: kscript new_project.kts package-name=co.myproject.example app-name="My 
         args.forEach { arg ->
             if (arg == KEY_HELP) {
                 showMessage(
-                    message = HELP_MESSAGE,
+                    message = helpMessage,
                     exitAfterMessage = true
                 )
             } else if (arg.startsWith("$KEY_APP_NAME$ARGUMENT_DELIMITER")) {
@@ -86,18 +86,18 @@ Example: kscript new_project.kts package-name=co.myproject.example app-name="My 
                 hasPackageName = true
             } else {
                 showMessage(
-                    message = "ERROR: Invalid argument name: $arg \n$HELP_MESSAGE",
+                    message = "ERROR: Invalid argument name: $arg \n$helpMessage",
                     exitAfterMessage = true
                 )
             }
         }
         when {
             !hasAppName -> showMessage(
-                message = "ERROR: No app name has been provided \n$HELP_MESSAGE",
+                message = "ERROR: No app name has been provided \n$helpMessage",
                 exitAfterMessage = true
             )
             !hasPackageName -> showMessage(
-                message = "ERROR: No package name has been provided \n$HELP_MESSAGE",
+                message = "ERROR: No package name has been provided \n$helpMessage",
                 exitAfterMessage = true
             )
         }
--- a/scripts/new_project.kts
+++ b/scripts/new_project.kts
@@ -113,8 +113,7 @@ object NewProject {
             appName = value.trim()
         } else {
             showMessage(
-                message = "ERROR: Invalid App Name: $value (needs to follow standard pattern {MyProject} or {My Project}) or {my-project} \n" +
-                        "$helpMessage",
+                message = "ERROR: Invalid App Name: $value (needs to follow standard pattern {MyProject} or {My Project}) or {my-project} \n$helpMessage",
                 exitAfterMessage = true
             )
         }
@@ -125,7 +124,7 @@ object NewProject {
             packageName = value.trim()
         } else {
             showMessage(
-                message = "ERROR: Invalid Package Name: $value (needs to follow standard pattern {com.example.package})",
+                message = "ERROR: Invalid Package Name: $value (needs to follow standard pattern {com.example.package}) \n$helpMessage",
                 exitAfterMessage = true
             )
         }
@@ -246,7 +245,8 @@ object NewProject {
         if (exitValue != 0) {
             showMessage(
                 message = "❌ Something went wrong! when executing command: $command",
-                exitAfterMessage = true
+                exitAfterMessage = true,
+                exitValue = exitValue
             )
         }
     }
@@ -272,7 +272,11 @@ object NewProject {
         sourceFile.writeText(sourceText)
     }
 
-    private fun showMessage(message: String, exitAfterMessage: Boolean = false) {
+    private fun showMessage(
+        message: String,
+        exitAfterMessage: Boolean = false,
+        exitValue: Int = 0
+    ) {
         println("\n${message}\n")
         if (exitAfterMessage) System.exit(0)
     }
--- a/scripts/new_project.kts
+++ b/scripts/new_project.kts
@@ -227,9 +227,9 @@ Example: kscript new_project.kts package-name=co.myproject.example app-name="My 
 
     private fun buildProjectAndRunTests() {
         showMessage("=> 🛠️ Building project...")
-        executeCommand("sh $projectPath${fileSeparator}gradlew -p $fileSeparator$projectPath assembleDebug")
+        executeCommand("sh $projectPath${fileSeparator}gradlew -p $projectPath assembleDebug")
         showMessage("=> 🚓 Running tests...")
-        executeCommand("sh $projectPath${fileSeparator}gradlew -p $fileSeparator$projectPath testStagingDebugUnitTest")
+        executeCommand("sh $projectPath${fileSeparator}gradlew -p $projectPath testStagingDebugUnitTest")
         showMessage("=> 🚀 Done! The project is ready for development")
     }
 
--- a/scripts/new_project.kts
+++ b/scripts/new_project.kts
@@ -5,6 +5,7 @@ object NewProject {
     private const val ARGUMENT_DELIMITER = "="
     private const val DOT_SEPARATOR = "."
     private const val KEY_APP_NAME = "app-name"
+    private const val KEY_HELP = "--help"
     private const val KEY_PACKAGE_NAME = "package-name"
     private const val MINUS_SEPARATOR = "-"
     private const val SPACE_SEPARATOR = " "
@@ -15,6 +16,14 @@ object NewProject {
 
     private const val PATTERN_APP = "^([A-Z][a-zA-Z0-9\\s]*)|([a-z][a-z0-9-]*)$"
     private const val PATTERN_PACKAGE = "^[a-z]+(\\.[a-z][a-z0-9]*)+$"
+
+    private val helpMessage = """
+Run kscript new_project.kts to create a new project with the following arguments:
+    package-name=   New package name (i.e., com.example.package)
+    app-name=       New app name (i.e., MyApp, "My App", "my-app")
+
+Example: kscript new_project.kts package-name=co.myproject.example app-name="My Project"
+    """.trimIndent()
 
     private val modules = listOf("app", "data", "domain")
     private val fileSeparator = File.separator
@@ -59,30 +68,58 @@ object NewProject {
     }
 
     private fun handleArguments(args: Array<String>) {
-        val argumentError = "ERROR: Invalid argument name: Ensure define arguments => app-name={\"MyProject\"} or {\"My Project\"} package-name={com.sample.myproject}"
-        when (args.size) {
-            1 -> when {
-                args.first().startsWith(KEY_APP_NAME) -> showErrorMessage("ERROR: No package name has been provided")
-                args.first().startsWith(KEY_PACKAGE_NAME) -> showErrorMessage("ERROR: No app name has been provided")
-                else -> showErrorMessage(argumentError)
+        val argumentError = "ERROR: Invalid argument name \n$helpMessage"
+        when {
+            args.size == 1 -> when {
+                args.first().isHelp() -> showMessage(
+                    message = helpMessage,
+                    exitAfterMessage = true
+                )
+                args.first().startsWith(KEY_APP_NAME) -> showMessage(
+                    message = "ERROR: No package name has been provided",
+                    exitAfterMessage = true
+                )
+                args.first().startsWith(KEY_PACKAGE_NAME) -> showMessage(
+                    message = "ERROR: No app name has been provided",
+                    exitAfterMessage = true
+                )
+                else -> showMessage(
+                    message = argumentError,
+                    exitAfterMessage = true
+                )
             }
-            2 -> args.forEach { arg ->
+            args.size == 2 && args.isBothKeyValuePair() -> args.forEach { arg ->
                 val (key, value) = arg.split(ARGUMENT_DELIMITER)
                 when (key) {
                     KEY_APP_NAME -> validateAppName(value)
                     KEY_PACKAGE_NAME -> validatePackageName(value)
-                    else -> showErrorMessage(argumentError)
+                    else -> showMessage(message = argumentError, exitAfterMessage = true)
                 }
             }
-            else -> showErrorMessage("ERROR: Require app-name and package-name to initialize the new project")
+            else -> showMessage(
+                message = "ERROR: Wrong format to initialize the new project \n$helpMessage",
+                exitAfterMessage = true
+            )
         }
     }
+
+    private fun Array<String>.isBothKeyValuePair(): Boolean {
+        val firstArg = this.first()
+        val secondArg = this[1]
+        return firstArg.contains(ARGUMENT_DELIMITER) && firstArg.contains("help")
+            .not() && secondArg.contains(ARGUMENT_DELIMITER) && secondArg.contains("help").not()
+    }
+
+    private fun String.isHelp(): Boolean = this == KEY_HELP
 
     private fun validateAppName(value: String) {
         if (PATTERN_APP.toRegex().matches(value)) {
             appName = value.trim()
         } else {
-            showErrorMessage("ERROR: Invalid App Name: $value (needs to follow standard pattern {MyProject} or {My Project})")
+            showMessage(
+                message = "ERROR: Invalid App Name: $value (needs to follow standard pattern {MyProject} or {My Project})",
+                exitAfterMessage = true
+            )
         }
     }
 
@@ -90,7 +127,10 @@ object NewProject {
         if (PATTERN_PACKAGE.toRegex().matches(value)) {
             packageName = value.trim()
         } else {
-            showErrorMessage("ERROR: Invalid Package Name: $value (needs to follow standard pattern {com.example.package})")
+            showMessage(
+                message = "ERROR: Invalid Package Name: $value (needs to follow standard pattern {com.example.package})",
+                exitAfterMessage = true
+            )
         }
     }
 
@@ -207,7 +247,10 @@ object NewProject {
         process.inputStream.reader().forEachLine { println(it) }
         val exitValue = process.waitFor()
         if (exitValue != 0) {
-            showErrorMessage("❌ Something went wrong! when executing command: $command", exitValue)
+            showMessage(
+                message = "❌ Something went wrong! when executing command: $command",
+                exitAfterMessage = true
+            )
         }
     }
 
@@ -232,13 +275,9 @@ object NewProject {
         sourceFile.writeText(sourceText)
     }
 
-    private fun showMessage(message: String) {
+    private fun showMessage(message: String, exitAfterMessage: Boolean = false) {
         println("\n${message}\n")
-    }
-
-    private fun showErrorMessage(message: String, exitCode: Int = 0) {
-        println("\n${message}\n")
-        System.exit(exitCode)
+        if (exitAfterMessage) System.exit(0)
     }
 
     private fun String.uppercaseEveryFirstCharacter(): String {
--- a/scripts/new_project.kts
+++ b/scripts/new_project.kts
@@ -68,36 +68,36 @@ Example: kscript new_project.kts package-name=co.myproject.example app-name="My 
     }
 
     private fun handleArguments(args: Array<String>) {
-        val argumentError = "ERROR: Invalid argument name \n$helpMessage"
-        when {
-            args.size == 1 -> when {
-                args.first().isHelp() -> showMessage(
-                    message = helpMessage,
+        var hasAppName = false
+        var hasPackageName = false
+        args.forEach { arg ->
+            if (arg == KEY_HELP) {
+                showMessage(
+                    message = HELP_MESSAGE,
                     exitAfterMessage = true
                 )
-                args.first().startsWith(KEY_APP_NAME) -> showMessage(
-                    message = "ERROR: No package name has been provided",
-                    exitAfterMessage = true
-                )
-                args.first().startsWith(KEY_PACKAGE_NAME) -> showMessage(
-                    message = "ERROR: No app name has been provided",
-                    exitAfterMessage = true
-                )
-                else -> showMessage(
-                    message = argumentError,
-                    exitAfterMessage = true
-                )
-            }
-            args.size == 2 && args.isBothKeyValuePair() -> args.forEach { arg ->
+            } else if (arg.startsWith("$KEY_APP_NAME$ARGUMENT_DELIMITER")) {
                 val (key, value) = arg.split(ARGUMENT_DELIMITER)
-                when (key) {
-                    KEY_APP_NAME -> validateAppName(value)
-                    KEY_PACKAGE_NAME -> validatePackageName(value)
-                    else -> showMessage(message = argumentError, exitAfterMessage = true)
-                }
+                validateAppName(value)
+                hasAppName = true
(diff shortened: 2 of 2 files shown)


scripts/new_project.kts
```diff
@@ -4,7 +4,16 @@ object NewProject {
 
     private const val ARGUMENT_DELIMITER = "="
     private const val DOT_SEPARATOR = "."
+    private const val HELP_MESSAGE = """
+Run kscript new_project.kts to create a new project with the following arguments:
+    package-name=   New package name (i.e. com.example.package)
+    app-name=       New app name (i.e. MyApp, "My App")
+    --help          Display useage message and exit
```
chornerman, line 11: I believe this should be "Display **usage** message and exit"

kaungkhantsoe: @chornerman it is the message for `--help` option when we use it like `kscript new_project.kts --help` to print out how we can use `new_project.kts` to make a sample project.

chornerman: @kaungkhantsoe I just thought that the word `useage` is a typo 😄

kaungkhantsoe: Ahh.. I see will update it. 👍

scripts/new_project.kts
```diff
@@ -62,6 +71,7 @@ object NewProject {
         val argumentError = "ERROR: Invalid argument name: Ensure define arguments => app-name={\"MyProject\"} or {\"My Project\"} package-name={com.sample.myproject}"
         when (args.size) {
             1 -> when {
+                args.first() == KEY_HELP -> showMessage(message = HELP_MESSAGE, exitAfterMessage = true)
```
Tuubz, line 74: What do you think about this suggestion, would it be better? 
```
    if (args.contains(KEY_HELP)) showMessage(message = HELP_MESSAGE, exitAfterMessage = true)
    val argumentError = "ERROR: Invalid argument name: Ensure define arguments => app-name={\"MyProject\"} or {\"My Project\"} package-name={com.sample.myproject}"
    when (args.size) {
        1 -> when {
            args.first().startsWith(KEY_APP_NAME) -> showErrorMessage("ERROR: No package name has been provided")
            args.first().startsWith(KEY_PACKAGE_NAME) -> showErrorMessage("ERROR: No app name has been provided")
            else -> showErrorMessage(argumentError)
        }
        ...
    }
}
```

Then `--help` is independent of the arguments size

scripts/new_project.kts
```diff
@@ -232,8 +242,9 @@ object NewProject {
         sourceFile.writeText(sourceText)
     }
 
-    private fun showMessage(message: String) {
+    private fun showMessage(message: String, exitAfterMessage: Boolean = false) {
         println("\n${message}\n")
+        if (exitAfterMessage) System.exit(0)
```
Tuubz, line 247: Should we re-factor `showMessage()` and `showErrorMessage()` in to one function ? 🙇

scripts/new_project.kts
Wadeewee: Hmm, I'm thinking should we show `HELP_MESSAGE` instead of `argumentError` 🤔

scripts/new_project.kts
```diff
@@ -4,7 +4,16 @@ object NewProject {
 
     private const val ARGUMENT_DELIMITER = "="
     private const val DOT_SEPARATOR = "."
+    private const val HELP_MESSAGE = """
+Run kscript new_project.kts to create a new project with the following arguments:
+    package-name=   New package name (i.e. com.example.package)
+    app-name=       New app name (i.e. MyApp, "My App")
+    --help          Display usage message and exit
```
Wadeewee, line 11: Do we need to show this 👇 when running script `--help` 🤔  Is that mean we will show(suggestion) all commands that can be run?
```suggestion
    --help          Display usage message and exit
```

scripts/new_project.kts
Tuubz: @Wadeewee Hmmmmm, good point 😂 

What do you think @kaungkhantsoe?

scripts/new_project.kts
Tuubz: I think @kaungkhantsoe was following the previous output message: 

Screen Shot 2022-07-29 at 3 46 37 PM

I'm totally fine to remove it, as it doesn't bring much value 🚀

scripts/new_project.kts
Wadeewee: Or should we suggest the user run script `kscript new_project.kts --help` when the argument is an error instead 🤔

scripts/new_project.kts
Wadeewee: I'm fine to keep it too, just asked for sure 🙏

scripts/new_project.kts
kaungkhantsoe: Sorry guys, I totally forgot about this PR. I agree with @Wadeewee. But I would also like to inform user about the arguments were error too. So how about we combine two ideas? We will show the usual help message if it is run with `--help` option and if there was an error, we will append `ERROR: Invalid argument name:` to the help message. So that the user can know the error state and also learn how to make a correct one. @Tuubz @Wadeewee

Tuubz: Works for me 👍

scripts/new_project.kts
```diff
@@ -4,7 +4,15 @@ object NewProject {
 
     private const val ARGUMENT_DELIMITER = "="
     private const val DOT_SEPARATOR = "."
+    private const val HELP_MESSAGE = """
+Run kscript new_project.kts to create a new project with the following arguments:
+    package-name=   New package name (i.e. com.example.package)
+    app-name=       New app name (i.e. MyApp, "My App")
+
+Example: kscript new_project.kts package-name=co.myproject.example app-name="My Project"
+    """
```
Wadeewee, line 13: Let's the format raw string to make it organized but you need to update the code a bit:
- In line `71` We need to call `.trimIndent()`-> `val argumentError = "ERROR: Invalid argument name: \n${HELP_MESSAGE.trimIndent()}"`
- And line `75` -> `message = HELP_MESSAGE.trimIndent()`
```suggestion
    private const val HELP_MESSAGE =
        """
            Run kscript new_project.kts to create a new project with the following arguments:
                package-name=   New package name (i.e. com.example.package)
                app-name=       New app name (i.e. MyApp, "My App")
            
            Example: kscript new_project.kts package-name=co.myproject.example app-name="My Project"
        """
```

Wadeewee (commented):
Rest LGTM 🚀

scripts/new_project.kts
chornerman: Or should we use `private val` and call` trimIndent()` when declaring the variable so we don't have to trim indent on each usage 🤔 

```
    private val HELP_MESSAGE =
        """
            Run kscript new_project.kts to create a new project with the following arguments:
                package-name=   New package name (i.e. com.example.package)
                app-name=       New app name (i.e. MyApp, "My App")
            
            Example: kscript new_project.kts package-name=co.myproject.example app-name="My Project"
        """.trimIndent()
```

scripts/new_project.kts
```diff
@@ -4,7 +4,15 @@ object NewProject {
 
     private const val ARGUMENT_DELIMITER = "="
     private const val DOT_SEPARATOR = "."
+    private const val HELP_MESSAGE = """
+Run kscript new_project.kts to create a new project with the following arguments:
+    package-name=   New package name (i.e. com.example.package)
+    app-name=       New app name (i.e. MyApp, "My App")
```
chornerman, line 10: I think we should also make this line the same as in the README file 🙏🏼 
```suggestion
    app-name=       New app name (i.e. MyApp, "My App", "my-app")
```

scripts/new_project.kts
Wadeewee: Good to me too 👍

scripts/new_project.kts
```diff
@@ -59,38 +67,60 @@ object NewProject {
...
             1 -> when {
-                args.first().startsWith(KEY_APP_NAME) -> showErrorMessage("ERROR: No package name has been provided")
-                args.first().startsWith(KEY_PACKAGE_NAME) -> showErrorMessage("ERROR: No app name has been provided")
-                else -> showErrorMessage(argumentError)
+                args.first() == KEY_HELP -> showMessage(
+                    message = HELP_MESSAGE,
+                    exitAfterMessage = true
+                )
+                args.first().startsWith(KEY_APP_NAME) -> showMessage(
+                    message = "ERROR: No package name has been provided",
```
chornerman, line 78: Should we extract this to constant? 🤔

scripts/new_project.kts
```diff
@@ -59,38 +67,60 @@ object NewProject {
...
+                    exitAfterMessage = true
+                )
+                args.first().startsWith(KEY_PACKAGE_NAME) -> showMessage(
+                    message = "ERROR: No app name has been provided",
```
chornerman, line 82: Same here

scripts/new_project.kts
```diff
@@ -59,38 +67,60 @@ object NewProject {
...
                 val (key, value) = arg.split(ARGUMENT_DELIMITER)
                 when (key) {
                     KEY_APP_NAME -> validateAppName(value)
                     KEY_PACKAGE_NAME -> validatePackageName(value)
-                    else -> showErrorMessage(argumentError)
+                    else -> showMessage(message = argumentError, exitAfterMessage = true)
                 }
             }
-            else -> showErrorMessage("ERROR: Require app-name and package-name to initialize the new project")
+            else -> showMessage(
```
ryan-conway, line 98: Not sure if we should update the logic or the error message here 🤔 
When entering `kscript new_project.kts package-name="my.package" app-name="MyApp" --help`, the error says `Error: Require app-name and package-name...` even though all three arguments are valid 🤔

Wadeewee: @ryan-conway Good point 👍 , I think we just show the `HELP_MESSAGE` in case of `else`, Same for line `85` and `95`

chornerman: I think we also miss handle this case

Screen Shot 2565-08-04 at 11 33 52

ryan-conway: Look like we shouldn't base the logic on argument count anymore, and instead should check to see if all the required arguments are there or if an invalid argument exists 🤔 How about something like this?
```kotlin
        var hasPackage = false
        var hasName = false
        args.forEach { arg ->
            val (key, value) = arg.split(ARGUMENT_DELIMITER)
            when (key) {
                KEY_APP_NAME -> {
                    validateAppName(value)
                    hasName = true
                }
                KEY_PACKAGE_NAME -> {
                    validatePackageName(value)
                    hasPackage = true
                }
                KEY_HELP -> showMessage(
                    message = HELP_MESSAGE,
                    exitAfterMessage = true
                )
                else -> showMessage(
                    message = "ERROR: Invalid argument name: $key $HELP_MESSAGE",
                    exitAfterMessage = true
                )
            }
        }
        when {
            !hasName -> showMessage(
                message = "ERROR: No app name has been provided",
                exitAfterMessage = true
            )
            !hasPackage -> showMessage(
                message = "ERROR: No package name has been provided",
                exitAfterMessage = true
            )
        }
```

chornerman: What do you think if we display a help message with a generic argument error on every error case? I think it should be enough, for instance

```
        val argumentError = "ERROR: Invalid argument: $HELP_MESSAGE"
        if (args.size == 1 && args.first() == KEY_HELP) {
            showMessage(message = HELP_MESSAGE, exitAfterMessage = true)
        } else if (args.size == 2) {
            args.forEach { arg ->
                try {
                    val (key, value) = arg.split(ARGUMENT_DELIMITER)
                    when (key) {
                        KEY_APP_NAME -> validateAppName(value)
                        KEY_PACKAGE_NAME -> validatePackageName(value)
                        else -> showMessage(message = argumentError, exitAfterMessage = true)
                    }
                } catch (exception: Exception) {
                    showMessage(message = argumentError, exitAfterMessage = true)
                }
            }
        } else {
            showMessage(message = argumentError, exitAfterMessage = true)
        }
```

@kaungkhantsoe @Wadeewee @ryan-conway @Tuubz

ryan-conway: @chornerman I really don't think the arg size check is the right call, because if we run `kscript new_project.kts package-name="my.package" package-name="my.package.again"` there's still two valid args even though `app-name` wasn't set 🤔

Wadeewee: Yeah, I agreed, @chornerman. As I recommended in this comment https://github.com/nimblehq/android-templates/pull/260#discussion_r934337402 to make it simple, but I think you need to update `logic` a bit. Let's use `when` instead of `if/else`

README.md
```diff
@@ -18,7 +18,7 @@ Our Android template: **[template](https://github.com/nimblehq/android-templates
 3. Run `kscript new_project.kts` to create a new project with the following arguments:
   ```   
     package-name=                      New package name (i.e. com.example.package)
-    app-name=                          New app name (i.e. MyApp, "My App")
+    app-name=                          New app name (i.e. MyApp, "My App", "my-app")
```
lydiasama, line 21: Some suggestions from Grammarly:
```
    package-name=                      New package name (i.e., com.example.package)
    app-name=                          New app name (i.e., MyApp, "My App", "my-app")
```

scripts/new_project.kts
```diff
@@ -4,7 +4,15 @@ object NewProject {
 
     private const val ARGUMENT_DELIMITER = "="
     private const val DOT_SEPARATOR = "."
+    private val HELP_MESSAGE = """
+Run kscript new_project.kts to create a new project with the following arguments:
+    package-name=   New package name (i.e. com.example.package)
+    app-name=       New app name (i.e. MyApp, "My App", "my-app")
```
lydiasama, line 10:
```suggestion
    package-name=   New package name (i.e., com.example.package)
    app-name=       New app name (i.e., MyApp, "My App", "my-app")
```

lydiasama (commented):
Some small addition suggestions.

scripts/new_project.kts
chornerman: @ryan-conway From your suggestion, I think if we run something like `kscript new_project.kts --help` it will crash since we can't split `--help`. How about if we combine our idea together 😄 

```
        var hasPackage = false
        var hasName = false
        val argumentError = "ERROR: Invalid argument\n$HELP_MESSAGE"
        if (args.size == 1 && args.first() == KEY_HELP) {
            showMessage(message = HELP_MESSAGE, exitAfterMessage = true)
        } else if (args.size == 2) {
            var hasPackage = false
            var hasName = false
            args.forEach { arg ->
                try {
                    val (key, value) = arg.split(ARGUMENT_DELIMITER)
                    when (key) {
                        KEY_APP_NAME -> {
                            validateAppName(value)
                            hasName = true
                        }
                        KEY_PACKAGE_NAME -> {
                            validatePackageName(value)
                            hasPackage = true
                        }
                        else -> showMessage(message = argumentError, exitAfterMessage = true)
                    }
                } catch (exception: Exception) {
                    showMessage(message = argumentError, exitAfterMessage = true)
                }
            }
            if (!hasName || !hasPackage) showMessage(message = argumentError, exitAfterMessage = true)
        } else {
            showMessage(message = argumentError, exitAfterMessage = true)
        }
```

ryan-conway: @chornerman Works for me 😄

scripts/new_project.kts
```diff
@@ -4,7 +4,15 @@ object NewProject {
 
     private const val ARGUMENT_DELIMITER = "="
     private const val DOT_SEPARATOR = "."
+    private val HELP_MESSAGE = """
```
chornerman, line 7: I think the name should be `helpMessage` and we should move it to near the other vals

kaungkhantsoe:
@ryan-conway @chornerman @Wadeewee Updated 🚀.

scripts/new_project.kts
lydiasama: @chornerman @ryan-conway I think we can simply check if `args.size > 1` then check further for the argument key. If we use `==2`, then in the future if we add more argument, we need to update the condition too.
```
var hasValidPackage = false
        var hasValidName = false
        val argumentError = "ERROR: Invalid argument\n$HELP_MESSAGE"
        when {
            args.size == 1 && args.first() == KEY_HELP -> {
                showMessage(message = HELP_MESSAGE, exitAfterMessage = true)
            }
            args.size > 1 -> {
                args.forEach { arg ->
                    try {
                        val (key, value) = arg.split(ARGUMENT_DELIMITER)
                        when (key) {
                            KEY_APP_NAME -> {
                                validateAppName(value)
                                hasValidName = true
                            }
                            KEY_PACKAGE_NAME -> {
                                validatePackageName(value)
                                hasValidPackage = true
                            }
                            else -> showMessage(message = argumentError, exitAfterMessage = true)
                        }
                    } catch (exception: Exception) {
                        showMessage(message = argumentError, exitAfterMessage = true)
                    }
                }
                if (!hasValidName || !hasValidPackage) showMessage(message = argumentError, exitAfterMessage = true)
            }
            else -> {
                showMessage(message = argumentError, exitAfterMessage = true)
            }
        }
```

scripts/new_project.kts
```diff
@@ -59,38 +68,69 @@ object NewProject {
     }
 
     private fun handleArguments(args: Array<String>) {
-        val argumentError = "ERROR: Invalid argument name: Ensure define arguments => app-name={\"MyProject\"} or {\"My Project\"} package-name={com.sample.myproject}"
-        when (args.size) {
-            1 -> when {
-                args.first().startsWith(KEY_APP_NAME) -> showErrorMessage("ERROR: No package name has been provided")
-                args.first().startsWith(KEY_PACKAGE_NAME) -> showErrorMessage("ERROR: No app name has been provided")
-                else -> showErrorMessage(argumentError)
+        val argumentError = "ERROR: Invalid argument name \n$HELP_MESSAGE"
```
ryan-conway, line 71: We'll need to update this line too since we changed `HELP_MESSAGE` to `helpMessage` 😅

scripts/new_project.kts
```diff
@@ -59,38 +68,69 @@ object NewProject {
...
+        when {
+            args.size == 1 -> when {
+                args.first().isHelp() -> showMessage(
+                    message = HELP_MESSAGE,
```
ryan-conway, line 75: This too 😄

scripts/new_project.kts
```diff
@@ -59,38 +68,69 @@ object NewProject {
...
                 when (key) {
                     KEY_APP_NAME -> validateAppName(value)
                     KEY_PACKAGE_NAME -> validatePackageName(value)
-                    else -> showErrorMessage(argumentError)
+                    else -> showMessage(message = argumentError, exitAfterMessage = true)
                 }
             }
-            else -> showErrorMessage("ERROR: Require app-name and package-name to initialize the new project")
+            else -> showMessage(
+                message = "ERROR: Wrong format to initialize the new project \n$HELP_MESSAGE",
```
ryan-conway, line 100: Last one here 😄

scripts/new_project.kts
```diff
@@ -16,6 +17,14 @@ object NewProject {
     private const val PATTERN_APP = "^([A-Z][a-zA-Z0-9\\s]*)|([a-z][a-z0-9-]*)$"
     private const val PATTERN_PACKAGE = "^[a-z]+(\\.[a-z][a-z0-9]*)+$"
 
+    private val helpMessage = """
+Run kscript new_project.kts to create a new project with the following arguments:
+    package-name=   New package name (i.e., com.example.package)
+    app-name=       New app name (i.e., MyApp, "My App", "my-app")
+
+Example: kscript new_project.kts package-name=co.myproject.example app-name="My Project"
+    """.trimIndent()
```
ryan-conway, line 26: I think @Wadeewee's comment about readability here meant to indent the actual message so it isn't aligned with the left most column, that's why we added the `.trimIndent()` call to it 😅 
```suggestion
    private val helpMessage = """
        Run kscript new_project.kts to create a new project with the following arguments:
            package-name=   New package name (i.e., com.example.package)
            app-name=       New app name (i.e., MyApp, "My App", "my-app")
        
        Example: kscript new_project.kts package-name=co.myproject.example app-name="My Project"
    """.trimIndent()
```

ryan-conway (commented):
Couple more concerns here 😅

scripts/new_project.kts
Tuubz: I've adapted @ryan-conway's suggestion a bit.

It would definitely be better if we can move away from checking the argument count:
```
        var hasAppName = false
        var hasPackageName = false
        args.forEach { arg ->
            if (arg == KEY_HELP) {
                showMessage(
                    message = HELP_MESSAGE,
                    exitAfterMessage = true
                )
            } else if (arg.contains("$KEY_APP_NAME$ARGUMENT_DELIMITER")) {
                val (key, value) = arg.split(ARGUMENT_DELIMITER)
                validateAppName(value)
                hasAppName = true
            } else if (arg.contains("$KEY_PACKAGE_NAME$ARGUMENT_DELIMITER")) {
                val (key, value) = arg.split(ARGUMENT_DELIMITER)
                validatePackageName(value)
                hasPackageName = true
            } else {
                showMessage(
                    message = "ERROR: Invalid argument name: $arg \n$HELP_MESSAGE",
                    exitAfterMessage = true
                )
            }
        }
        when {
            !hasAppName -> showMessage(
                message = "ERROR: No app name has been provided \n$HELP_MESSAGE",
                exitAfterMessage = true
            )
            !hasPackageName -> showMessage(
                message = "ERROR: No package name has been provided \n$HELP_MESSAGE",
                exitAfterMessage = true
            )
        }
    }
```

Hopefully, I didn't miss any cases 🙏

ryan-conway: @Tuubz Shall we use `arg.startsWith()` instead of `arg.contains()`? Just to prevent inputs like `some-app-name=` 🤔

chornerman: @Tuubz @ryan-conway Look neat! How about if we also show the `helpMessage` when displaying a validate app name and package name error? 😄

chornerman: One more thing, should we also add the `{my-project}` pattern to the validate app name error message to match the help message? 🤔 Something like
`ERROR: Invalid App Name: $value (needs to follow standard pattern {MyProject} or {My Project} or {my-project})`

ryan-conway (commented):
The rest LGTM aside from [this](https://github.com/nimblehq/android-templates/pull/260#discussion_r938442265) and [this](https://github.com/nimblehq/android-templates/pull/260#discussion_r938447865) 😄

scripts/new_project.kts
```diff
@@ -16,6 +17,14 @@ object NewProject {
...
+    private val helpMessage = """
+        Run kscript new_project.kts to create a new project with the following arguments:
+            package-name=   New package name (i.e., com.example.package)
+            app-name=       New app name (i.e., MyApp, "My App", "my-app")
+        
+        Example: kscript new_project.kts package-name=co.myproject.example app-name="My Project"
+    """.trimIndent()
+
     private val modules = listOf("app", "data", "domain")
     private val fileSeparator = File.separator
```
chornerman, line 29: (Optional) Add a blank line here so there's one blank line between each variable 😄

scripts/new_project.kts
```diff
@@ -59,38 +68,69 @@ object NewProject {
...
+                hasAppName = true
+            } else if (arg.startsWith("$KEY_PACKAGE_NAME$ARGUMENT_DELIMITER")) {
+                val (key, value) = arg.split(ARGUMENT_DELIMITER)
+                validatePackageName(value)
+                hasPackageName = true
+            } else {
+                showMessage(
+                    message = "ERROR: Invalid argument name: $arg \n$helpMessage",
+                    exitAfterMessage = true
+                )
```
lydiasama, line 91: Shall we use `when` instead for more readable?
```
            when {
                 arg == KEY_HELP -> {
                     showMessage(
                         message = helpMessage,
                         exitAfterMessage = true
                     )
                 }
                 arg.startsWith("$KEY_APP_NAME$ARGUMENT_DELIMITER") -> {
                     val (key, value) = arg.split(ARGUMENT_DELIMITER)
                     validateAppName(value)
                     hasAppName = true
                 }
                 arg.startsWith("$KEY_PACKAGE_NAME$ARGUMENT_DELIMITER") -> {
                     val (key, value) = arg.split(ARGUMENT_DELIMITER)
                     validatePackageName(value)
                     hasPackageName = true
                 } 
                 else -> {
                     showMessage(
                         message = "ERROR: Invalid argument name: $arg \n$helpMessage",
                         exitAfterMessage = true
                     )
                 }
             }
```

Tuubz (commented):
Rest LGTM 🔥

scripts/new_project.kts
```diff
@@ -232,13 +272,13 @@ object NewProject {
...
-
-    private fun showErrorMessage(message: String, exitCode: Int = 0) {
+    private fun showMessage(
+        message: String,
+        exitAfterMessage: Boolean = false,
+        exitValue: Int = 0
+    ) {
         println("\n${message}\n")
-        System.exit(exitCode)
+        if (exitAfterMessage) System.exit(0)
```
ryan-conway, line 281: Pretty sure we should pass in the `exitValue` here 😉 
```suggestion
        if (exitAfterMessage) System.exit(exitValue)
```

Wadeewee (approved):

scripts/new_project.kts
Tuubz: @kaungkhantsoe Last one 💪 💪 💪 💪 🇲🇲

chornerman (approved):
Besides Ryan's comment LGTM 🎉

Tuubz (approved):
Any last remarks? @lydiasama @ryan-conway  🚀

ryan-conway (approved):
LGTM now 👍🏻

Issue #242, opened 2022-06-17
[Sample] Create sample folder for Jetpack Compose


Tuubz:
**Description**

Create a sample project for Jetpack Compose inside `sample-compose`

1. Generate `sample-compose` from `new_project.kts`
2. Add sample code related to Jetpack Compose

**Notes**: 
- Only take the minimum which is required to showcase Jetpack Compose, keep it simple
- `sample-compose` must have the same folder structure as `sample-xml`

**Acceptance Criteria**
- Add sample code for Jetpack Compose to `sample-compose`:
   - Home screen:
       - Contains title bar at the top
       - Contains a recycler view showing API data
       - Contains a progress bar when loading API data
   - Second screen:
      - Contains title bar at the top
     - Contains information of the clicked recycler view item from home screen
     
- Package name must be `co.nimblehq.sample.compose`

PR #272, opened 2022-08-08
[#242] [Part 1] Create sample folder for Jetpack Compose

chornerman:
#242 

## What happened 👀

We have to create a sample project which integrates the Jetpack Compose

## Insight 📝

- [x] Generate a new project using `new_project.kts`
  - App name: `sample-compose`
  - Package name: `co.nimblehq.sample.compose`

- [x] Replace template references with Compose sample references
- [x] Remove unused code and dependencies
- [x] Restructure project

## Proof Of Work 📹

Screenshot_20220817_160053

PR #274, opened 2022-08-09
[#242] [Part 2] Create sample folder for Jetpack Compose

chornerman:
#242 

## What happened 👀

We have to create a sample project which integrates the Jetpack Compose (inside the Fragment and still using the traditional navigation library 📝 #279 )

## Insight 📝

- [x] Add sample code for Jetpack Compose to `sample-compose`:
- Home screen:
  - Contains a title bar at the top
  - Contains a list showing API data
  - Contains a progress bar when loading API data
- Second screen:
  - Contains a title bar at the top
  - Contains information of the clicked list item from the Home screen
- [x] Remove unused code and refactor code
- [x] Update Detekt config file to support Jetpack Compose convention
  - [FunctionNaming](https://detekt.dev/docs/rules/naming#functionnaming) issue: add `ignoreAnnotated: [ 'Composable' ]`
  - [TopLevelPropertyNaming](https://detekt.dev/docs/rules/naming#toplevelpropertynaming) issue: update `propertyPattern` to `'[A-Z][A-Za-z0-9]*'`

## Proof Of Work 📹

https://user-images.githubusercontent.com/13492460/185603539-764a9c34-3291-4643-b835-bd977a382c1f.mp4


sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/base/BaseComposeFragment.kt
```diff
@@ -1,62 +0,0 @@
...
-import androidx.lifecycle.Lifecycle
-import androidx.lifecycle.lifecycleScope
-import androidx.lifecycle.repeatOnLifecycle
-import co.nimblehq.sample.compose.ui.common.Toaster
-import kotlinx.coroutines.flow.Flow
-import kotlinx.coroutines.flow.collect
-import kotlinx.coroutines.launch
-import javax.inject.Inject
-
-abstract class BaseComposeFragment : Fragment(), BaseComposeFragmentCallbacks {
```
Tuubz, line 20: I think the plan is to keep this class, right?

sample-compose/app/src/main/res/navigation/nav_graph_main.xml
```diff
@@ -1,17 +0,0 @@
...
-    <fragment
-        android:id="@+id/homeFragment"
-        android:name="co.nimblehq.sample.compose.ui.screens.xml.HomeFragment"
-        tools:layout="@layout/fragment_home" />
-
-    <fragment
-        android:id="@+id/homeComposeFragment"
-        android:name="co.nimblehq.sample.compose.ui.screens.compose.HomeComposeFragment" />
-
-</navigation>
```
Tuubz, line 17: Sooo compose doesn't make use of navigation component? 😯

Tuubz (commented):
Hmmm, I think there might have been a miscommunication here. Ideally, we should keep all the classes that `template` is using, but adapt to make use of `Compose`. My expectation was to update the classes, instead or removing them 🙇

sample-compose/app/src/main/res/navigation/nav_graph_main.xml
ryan-conway: @Tuubz [it does](https://developer.android.com/jetpack/compose/navigation), but it goes through Kotlin instead of XML 😉

Tuubz: Cool, but also quite unexpected/confusing, since we weren't using this before 💭 

Just to make sure, we can still combine `fragment navigation with xml` with `compose`, right?

I'm thinking if we should stick with the old approach first, so it's aligned with the `template`. It could be quite confusing when a developer compares `template` with `sample-compose`  🤔 

cc @nimblehq/android-chapter

Tuubz: Resolved: https://github.com/nimblehq/android-templates/discussions/279

sample-compose/app/src/main/java/co/nimblehq/sample/compose/navigation/Navigation.kt
```diff
@@ -0,0 +1,38 @@
...
+
+    const val SECOND_ARGUMENT_ID = "id"
+}
+
+@Composable
+fun Navigation() {
+    val navController = rememberNavController()
+    NavHost(navController = navController, startDestination = HOME_ROUTE) {
+        composable(route = HOME_ROUTE) {
+            HomeScreen(navController = navController)
```
lydiasama, line 25: Only one argument with the same name seems to be in excess to me. I think we can omit the argument name in this case:
```suggestion
            HomeScreen(navController)
```

sample-compose/app/src/main/java/co/nimblehq/sample/compose/navigation/Navigation.kt
```diff
@@ -0,0 +1,38 @@
...
+        }
+        composable(
+            route = SECOND_ROUTE + "/{${SECOND_ARGUMENT_ID}}",
```
lydiasama, line 28: Why do we need to add `{}` for this variable?
```suggestion
            route = SECOND_ROUTE + "/{$SECOND_ARGUMENT_ID}",
```

sample-compose/app/build.gradle.kts
```diff
@@ -123,34 +123,26 @@ dependencies {
...
-    implementation("androidx.navigation:navigation-runtime-ktx:${Versions.ANDROIDX_NAVIGATION_VERSION}")
-    implementation("androidx.navigation:navigation-ui-ktx:${Versions.ANDROIDX_NAVIGATION_VERSION}")
+    implementation("androidx.activity:activity-compose:${Versions.COMPOSE_ACTIVITY_VERSION}")
+    implementation("androidx.navigation:navigation-compose:${Versions.COMPOSE_NAVIGATION_VERSION}")
+    implementation("androidx.lifecycle:lifecycle-viewmodel-compose:${Versions.COMPOSE_VIEWMODEL_VERSION}")
+    implementation("androidx.hilt:hilt-navigation-compose:${Versions.HILT_COMPOSE_NAVIGATION_VERSION}")
 
     implementation("com.google.dagger:hilt-android:${Versions.HILT_VERSION}")
 
     implementation("com.jakewharton.timber:timber:${Versions.TIMBER_LOG_VERSION}")
```
lydiasama, line 139: Do we use this Timber log in any places of this project? 🤔

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeViewModel.kt
```diff
@@ -14,25 +15,27 @@ import javax.inject.Inject
 @HiltViewModel
 class HomeViewModel @Inject constructor(
     private val useCase: UseCase,
-    dispatchers: DispatchersProvider
+    val toaster: Toaster,
```
lydiasama, line 18: I don't see this variable being used in this class. 🤔

sample-compose/buildSrc/src/main/java/Versions.kt
```diff
@@ -1,7 +1,7 @@
 object Versions {
     const val BUILD_GRADLE_VERSION = "4.2.1"
 
-    const val ANDROID_COMPILE_SDK_VERSION = 30
+    const val ANDROID_COMPILE_SDK_VERSION = 31
```
lydiasama, line 4: @Tuubz @chornerman  Shall we align the version with the main template too? 🤔 Or we can update it separately on each sample?

ryan-conway:
@chornerman Let's update the `detekt-config.yml` with the [rules for Compose](https://detekt.dev/docs/introduction/compose/), then we can get rid of some suppressions 😄

sample-compose/buildSrc/src/main/java/Versions.kt
ryan-conway: @lydiasama if we align with the main template, does that mean we'll need to update both samples too each time we change something? 🤔

lydiasama: Yeah @ryan-conway  That's the idea 👍🏻  I think it's good to clarify this time to be aligned. 🙌🏻

lydiasama:
Conflict 🦀  @chornerman

lydiasama:
> @chornerman Let's update the `detekt-config.yml` with the [rules for Compose](https://detekt.dev/docs/introduction/compose/), then we can get rid of some suppressions 😄

Let's create this issue to update the main template too. 👍🏻

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeScreen.kt
```diff
@@ -0,0 +1,36 @@
...
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.LocalContext
+import androidx.hilt.navigation.compose.hiltViewModel
+import androidx.navigation.NavController
+import co.nimblehq.sample.compose.navigation.NavigationRoute
+import co.nimblehq.sample.compose.ui.userReadableMessage
+
+@Composable
+fun HomeScreen(navController: NavController) {
+    val viewModel = hiltViewModel<HomeViewModel>()
```
ryan-conway, line 18: Shouldn't we pass in `viewModel` as an argument? This is quite coupled with Hilt 👀

ryan-conway:
@lydiasama Maybe we should leave the template `detekt-rules` as is and just update the Compose sample 🤔 Not sure if it's okay to ignore an XML layout project with functions like `fun PrintLog()` and the other rules, what do you think? CC: @chornerman @Tuubz

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeViewModel.kt
ryan-conway: @lydiasama Looks like it's being accessed directly from `HomeScreen` 🤔 Do you think we should make it private and have a public method for `toast()` instead? 🤔

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeScreen.kt
```diff
@@ -0,0 +1,36 @@
...
+            uiModels = uiModels,
+            onItemClick = { id -> navController.navigate("${NavigationRoute.SECOND_ROUTE}/$id") }
+        )
+        if (showLoading) {
+            CircularProgressIndicator(
+                modifier = Modifier.align(alignment = Alignment.Center)
+            )
+        }
+    }
+}
```
ryan-conway, line 36: Let's see if we can add a preview too 😄

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/Item.kt
```diff
@@ -0,0 +1,28 @@
...
+import androidx.compose.foundation.layout.*
+import androidx.compose.material.Surface
+import androidx.compose.material.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import co.nimblehq.sample.compose.model.UiModel
+import co.nimblehq.sample.compose.ui.theme.Dimension
+
+@Composable
+fun Item(
```
ryan-conway, line 13: Preview for this would be nice too, and all the other composables 😄

ryan-conway (commented):
First round 💪🏻

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeViewModel.kt
lydiasama: @ryan-conway I agree 👍🏻 . Moreover, it seems to not use the toaster here inside of this class too. So I think we can get rid of it.

ryan-conway: That might depend on [how we do navigation](https://github.com/nimblehq/android-templates/discussions/279) for this project tho 😉 
If we stick with fragments, then we can just inject the `Toaster` into the container fragment like usual 😄
But if we go with the new navigation method, we'll need to keep it in the `ViewModel` to stop it being too coupled with the `HomeScreen` composable 🤔

Wadeewee:
> @chornerman Let's update the `detekt-config.yml` with the [rules for Compose](https://detekt.dev/docs/introduction/compose/), then we can get rid of some suppressions 😄

Good catch @ryan-conway 👍 It would be great if we update `detekt-config.yml`, I pretty sure at least we can get rid of `@Suppress("FunctionNaming")` on the `Compose` function.

chornerman:
@ryan-conway @lydiasama @Tuubz I just updated this PR according to #279. Please help take a look again and leave comments if any. 🙏🏼

sample-compose/app/build.gradle.kts
```diff
@@ -125,6 +125,7 @@ dependencies {
 
     implementation("androidx.activity:activity-ktx:${Versions.ANDROIDX_ACTIVITY_KTX_VERSION}")
     implementation("androidx.appcompat:appcompat:${Versions.ANDROIDX_SUPPORT_VERSION}")
+    implementation("androidx.constraintlayout:constraintlayout:${Versions.CONSTRAINT_LAYOUT_VERSION}")
```
ryan-conway, line 128: Do we still need this library? We can probably replace the `ConstraintLayout` in `activity_main.xml` with a `LinearLayout` or something, what do you think? 😄

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/base/BaseComposeFragment.kt
```diff
@@ -35,7 +36,7 @@ abstract class BaseComposeFragment : Fragment(), BaseComposeFragmentCallbacks {
         savedInstanceState: Bundle?
     ): View? {
         return ComposeView(requireContext()).apply {
-            setContent { composeScreen.invoke() }
+            setContent { Theme { composeScreen.invoke() } }
```
ryan-conway, line 39: Quick question, are we gonna leave the theme as `Theme` and not something like `ComposeSampleTheme`? Not sure what our convention will be for this 😄 Altho I suppose convention is still being written so no biggie 😄

ryan-conway (commented):
Navigation looks fine, just a couple more comments 😄

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/ItemList.kt
```diff
@@ -0,0 +1,25 @@
...
+) {
+    LazyColumn {
+        itemsIndexed(items = uiModels) { index, uiModel ->
+            Item(
+                uiModel = uiModel,
+                onClick = onItemClick
+            )
+
+            val lastItemIndex = uiModels.size - 1
+            if (index != lastItemIndex) Divider()
```
minhnimble, line 22: I think we can provide the index computation directly into the if check for simplicity. What do you think?
```suggestion
            if (index != uiModels.size - 1) Divider()
```

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeComposeScreen.kt
```diff
@@ -0,0 +1,30 @@
...
+import androidx.compose.material.CircularProgressIndicator
+import androidx.compose.material.Scaffold
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import co.nimblehq.sample.compose.model.UiModel
+import co.nimblehq.sample.compose.ui.screens.AppBar
+
+@Composable
+fun HomeComposeScreen(
```
Wadeewee, line 14: I think we can naming as `HomeScreen` same as `HomeFragment` and `HomeViewModel`

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/ItemList.kt
Wadeewee: Also, I think we can use just `if (index != uiModels.lastIndex) Divider()` So we get rid of using `- 1` too 🙏

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/Item.kt
```diff
@@ -0,0 +1,26 @@
...
+    Surface(
+        modifier = Modifier
+            .fillMaxWidth()
+            .clickable(onClick = { onClick(uiModel) })
+    ) {
+        Row(modifier = Modifier.padding(Dimension.SpacingNormal)) {
+            Text(text = uiModel.id.toString())
+        }
+    }
+}
```
Wadeewee, line 26: @chornerman I don't think we need a `Surface` here, Let's add the preview then we can see how was it look like. Here's some idea:
```suggestion
    Row(
        modifier = Modifier
            .fillMaxWidth()
            .padding(Dimension.SpacingNormal)
            .clickable(onClick = { onClick(uiModel) })
    ) {
        Text(text = uiModel.id.toString())
    }
}

@Preview(showBackground = true)
@Composable
fun ItemPreview() {
    val uiModel = UiModel(id = 1)
    Theme {
        Item(
            uiModel = uiModel,
            onClick = { uiModel.id }
        )
    }
}
```

Wadeewee: Oh, my bad 🙏 I got it now why you need to use Surface, We need to make the background
cover all item when clicked 🚀

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/AppBar.kt
```diff
@@ -0,0 +1,14 @@
+package co.nimblehq.sample.compose.ui.screens
+
+import androidx.compose.material.Text
+import androidx.compose.material.TopAppBar
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.res.stringResource
+import co.nimblehq.sample.compose.R
+
+@Composable
+fun AppBar() {
```
Wadeewee, line 10: Shall we create a parameter for `title`? Currently, we use the sample title on both `Home` and `Second` screen

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/second/SecondFragment.kt
```diff
@@ -0,0 +1,19 @@
+package co.nimblehq.sample.compose.ui.screens.second
+
+import androidx.compose.runtime.Composable
+import androidx.navigation.fragment.navArgs
+import co.nimblehq.sample.compose.ui.base.BaseComposeFragment
+import dagger.hilt.android.AndroidEntryPoint
+
+@AndroidEntryPoint
+class SecondFragment : BaseComposeFragment() {
```
Wadeewee, line 9: Shall we rename it to `DetailFragment`, same as `DetailScreen`

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/Item.kt
Wadeewee, line 26: @chornerman I think we can get rid of `Surface` and use just `Row` is enough. Let's create a `preview` then we can see how was it looks like. Here's some idea
```suggestion
    Row(modifier = Modifier
        .fillMaxWidth()
        .clickable(onClick = { onClick(uiModel) })
    ) {
        Text(
            modifier = Modifier.padding(Dimension.SpacingNormal),
            text = uiModel.id.toString()
        )
    }
}

@Preview(showBackground = true)
@Composable
fun ItemPreview() {
    val uiModel = UiModel(id = 1)
    Theme {
        Item(
            uiModel = uiModel,
            onClick = { uiModel.id }
        )
    }
}
```

ryan-conway: @Wadeewee Isn't it good practice to build composables on a `Surface` so that it has appropriate content colours? 🤔 
[Reference](https://developer.android.com/jetpack/compose/themes/material#content-color)

sample-compose/app/src/main/java/co/nimblehq/sample/compose/navigation/Navigation.kt
chornerman: Resolved: https://github.com/nimblehq/android-templates/discussions/279

sample-compose/app/src/main/java/co/nimblehq/sample/compose/navigation/Navigation.kt
chornerman: Resolved: https://github.com/nimblehq/android-templates/discussions/279

sample-compose/app/build.gradle.kts
chornerman: Resolved: https://github.com/nimblehq/android-templates/discussions/279

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeViewModel.kt
chornerman: Resolved: https://github.com/nimblehq/android-templates/discussions/279

sample-compose/buildSrc/src/main/java/Versions.kt
chornerman: Resolved: https://github.com/nimblehq/android-templates/discussions/279

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeScreen.kt
chornerman: Resolved: https://github.com/nimblehq/android-templates/discussions/279

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/Item.kt
Wadeewee: @ryan-conway It seems like on the current implementation, we don't use `Surface` to handle the color anymore, but we're using it to handle `clickable ` and `padding` instead, Also, I didn't see any requirement of `color` for `background` of the `main screen` or `content` anymore. That's why I don't think we need it here 🙇‍♀️ 

And here're the result that I've got  with using `Surface` and without `Surface`
Surface
NonSurface

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeComposeScreen.kt
chornerman: Seem like @Tuubz don't agree on this 😄 https://github.com/nimblehq/android-templates/pull/272#discussion_r947660564

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/second/SecondFragment.kt
chornerman: Conclusion: Let's use SecondFragment and SecondScreen for now to align with XML. We can consider renaming them later

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeComposeScreen.kt
chornerman: Conclusion: I will leave TODO comments for now and we will consider renaming them when we [completely opt-out Fragment structure](https://github.com/nimblehq/android-templates/discussions/279#discussioncomment-3412071)

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/AppBar.kt
chornerman: Conclusion: We will use "Home" as the title for the Home screen and "Second" as the title for the Second screen

sample-compose/app/src/main/java/co/nimblehq/sample/compose/extension/ViewModelExt.kt
```diff
@@ -3,23 +3,19 @@ package co.nimblehq.sample.compose.extension
...
-import androidx.fragment.app.*
+import androidx.fragment.app.Fragment
+import androidx.fragment.app.viewModels
 import androidx.lifecycle.*
 
 /**
  * PLEASE READ THIS BEFORE IMPLEMENT:
- * Right now, there is no easy way to mock/ fake the viewModel inside the Fragment when applying
- * the 'by viewModels()' Kotlin property delegate from the activity-ktx/ fragment-ktx artifact.
+ * Right now, there is no easy way to mock / fake the viewModel inside the Fragment when applying
```
lydiasama, line 12:
```suggestion
 * Right now, there is no easy way to mock/fake the viewModel inside the Fragment when applying
```
I think the correct grammar is to omit the space before and after the `/`. Confirmed by Grammarly.

sample-compose/app/src/main/java/co/nimblehq/sample/compose/extension/ViewModelExt.kt
```diff
@@ -3,23 +3,19 @@ package co.nimblehq.sample.compose.extension
...
+ * the 'by viewModels()' Kotlin property delegate from the activity-ktx / fragment-ktx artifact.
```
lydiasama, line 13:
```suggestion
 * the 'by viewModels()' Kotlin property delegate from the activity-ktx/fragment-ktx artifact.
```
Ditto here.

sample-compose/app/src/main/java/co/nimblehq/sample/compose/model/UiModel.kt
```diff
@@ -1,10 +1,13 @@
...
 import co.nimblehq.sample.compose.domain.model.Model
+import kotlinx.parcelize.Parcelize
 
+@Parcelize
 data class UiModel(
     val id: Int
-)
+) : Parcelable
 
 private fun Model.toUiModel() = UiModel(id = id ?: -1)
```
lydiasama, line 12: The default `-1` could lead the text shows the `-1` string if I'm not wrong. 🤔  Why don't we declare the `UiModel.id` to be String and set the empty string as the default since we only use it to display?

sample-compose/app/src/main/res/values/strings.xml
```diff
@@ -2,4 +2,7 @@
 <resources>
     <string name="app_name">Sample Compose</string>
     <string name="error_generic">Unexpected error</string>
+    <string name="id_title">ID: %1$d</string>
+    <string name="home_title">Home</string>
+    <string name="second_title">Second</string>
```
Wadeewee, line 7: Let's rename them a bit to make it clearer. Also, add one more space in each section of screen that could be nice 🚀 
```suggestion
    <string name="error_generic">Unexpected error</string>
    
    <string name="home_title_bar">Home</string>
    
    <string name="second_title_bar">Second</string>
    <string name="second_id_title">ID: %1$d</string>
```

Wadeewee (commented):
Rest LGTM 🚀

sample-compose/app/src/main/java/co/nimblehq/sample/compose/navigation/Navigation.kt
lydiasama: @chornerman I'm not sure what you mean with the `Resolved: #279` here? Did you mean to resolve this once open the PR for #279 issue, or have you resolved it already in this PR?

sample-compose/app/src/main/java/co/nimblehq/sample/compose/model/UiModel.kt
Wadeewee: Good point @lydiasama 👍 I agree but I'm not sure should we follow the `template` or not for this 🤔

lydiasama: @Wadeewee I don't think we need to keep aligned with the template in this case. Because in the template, we don't use this value as a sample. 🤔 Also, some decision depends on the project independently. ([See the discussion here](https://github.com/nimblehq/android-templates/pull/274#discussion_r948629123))

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/theme/Color.kt
```diff
@@ -2,7 +2,6 @@ package co.nimblehq.sample.compose.ui.theme
 
 import androidx.compose.ui.graphics.Color
 
-@Suppress("MagicNumber")
```
ryan-conway, line 5: Still need to [update](https://detekt.dev/docs/introduction/compose/) `detekt-rules.yml` tho 😉

chornerman: Thank you, on the way 😄

sample-compose/app/src/main/java/co/nimblehq/sample/compose/navigation/Navigation.kt
chornerman: @lydiasama I mean this comment is outdated because I updated the PR according to the #279 discussion 😄

lydiasama: I see. So I think it sounds to be more clear if we say: `Resolve in #279`. 🤔 The `Resolved` meant we have resolved this issue already to me. 🙏🏻

Wadeewee (approved):

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/theme/Color.kt
```diff
@@ -1,9 +1,7 @@
+@file:Suppress("MagicNumber")
```
ryan-conway, line 1: Why not update `MagicNumber` in `detekt-rules.yml` instead of suppressing this? 🤔 
```
  MagicNumber:
    ignorePropertyDeclaration: true
```

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/theme/Dimension.kt
```diff
@@ -1,8 +1,7 @@
+@file:Suppress("MagicNumber")
```
ryan-conway, line 1: Same concern here 😅

ryan-conway (commented):
👀

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/theme/Color.kt
chornerman: @ryan-conway After discussing with @Wadeewee, we worry that setting this rule will also affect other non-composable functions. For example, we will not get any warning if we use a magic number in an area unrelated to compose. Moreover, there should be only two files we have to use the magic numbers: Color and Dimension. What do you think?

ryan-conway: That makes sense, but we should reconsider in case we end need to use magic number somewhere else. Sounds good to me! 👍🏻

ryan-conway (approved):

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/theme/Color.kt
chornerman: Agree, thank you

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/theme/Dimension.kt
```diff
@@ -1,8 +1,7 @@
...
+
 package co.nimblehq.sample.compose.ui.theme
 
 import androidx.compose.ui.unit.dp
 
-@Suppress("MagicNumber")
-object Dimension {
-    val SpacingNormal = 16.dp
-}
+val SpacingNormal = 16.dp
```
lydiasama, line 7: Shall we make it to be a compile-time constant by using `const val`?
since this dimension will never change. 🙏🏻

lydiasama (commented):
One last question. The rest LGTM.

minhnimble (approved):

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/theme/Dimension.kt
chornerman: @lydiasama I think it's not possible since the dimen variable has type DP, which is not a primitive type 🙏🏼

Screen Shot 2565-08-22 at 11 33 54

lydiasama: Oh, it's true. 😂

lydiasama (approved):

PR #278, opened 2022-08-16
[#273] Fix incorrect step for running tests in `new_project.kts`

### LINKED ISSUE #273: Fix incorrect step for running tests in `new_project.kts`
## Issue

When running `testStagingDebugUnitTest`, it only executes the tests for `:app` module

## Expected

When running `jacocoTestReport`, it executes the tests for `:app`, `:domain` & `:data` module

## Steps to reproduce

1. Run `new_project.kts`
2. Observe logs

Tuubz:
Closes #273

## What happened 👀

`testStagingDebugUnitTest` only executes tests for `app` module, so execute command for each module separately

## Proof Of Work 📹

Please check "Verify generating new project from template" logs
https://github.com/nimblehq/android-templates/runs/7893007827?check_suite_focus=true

### DIFF
--- a/scripts/new_project.kts
+++ b/scripts/new_project.kts
@@ -248,7 +248,9 @@ object NewProject {
             "$projectPath${fileSeparator}gradlew",
             "-p",
             "$projectPath",
-            "jacocoTestReport"
+            ":app:testStagingDebugUnitTest",
+            ":data:testDebugUnitTest",
+            ":domain:test"
         )
         showMessage("=> 🚀 Done! The project is ready for development")
     }
--- a/scripts/new_project.kts
+++ b/scripts/new_project.kts
@@ -248,7 +248,7 @@ object NewProject {
             "$projectPath${fileSeparator}gradlew",
             "-p",
             "$projectPath",
-            "testStagingDebugUnitTest"
+            "jacocoTestReport"
         )
         showMessage("=> 🚀 Done! The project is ready for development")
     }


chornerman:
@Tuubz Conflict 🦀

chornerman (approved):
@Tuubz Do we have any POW for this PR? 🙏🏼

ryan-conway (approved):

scripts/new_project.kts
```diff
@@ -248,7 +248,7 @@ object NewProject {
             "$projectPath${fileSeparator}gradlew",
             "-p",
             "$projectPath",
-            "testStagingDebugUnitTest"
+            "jacocoTestReport"
```
sleepylee, line 251: I have mixed feeling of this solution 🤔 what if we don't use jacoco anymore, given that there is Kover on the corner waiting for the evaluation 🤔

Tuubz: `testStagingDebugUnitTest` is only running tests for `app` module, hence the update

If we end up using Kover, we can always adapt it again later 🚀

sleepylee: have you tried just `test` or `testStagingDebug`? I think jacoco must run a similar gradle command under the hood, so probably you can sneak into the gradle log to see what has been called. Running jacocoTestReport is slower and you are not going to read the report at the time you init the project right?

Tuubz: @sleepylee You're right, `jacocoTestReport` should be avoided, as we won't be reading the report 👍 

Unfortunately, I can't make it work with other commands:
- `test` : Executes tests for both `debug` & `release` variant too, which should be avoided
- `testStagingDebug` : Only runs tests for `app` module
- `testDebugUnitTest` : Only runs tests for `data` module

Maybe something is wrong with our set-up 💭 

Screen Shot 2565-08-18 at 12 32 18

sleepylee: @Tuubz seems like it's possibly because we don't have `staging` flavor in `data`, so perhaps we can execute 2 commands:
```
testStagingDebug -> this ensure any module that has Staging flavor will run unit test
testDebugUnitTest -> any module that does not have extra flavor, must run the debug build type
```

pros: 
 - you don't need to think about this again when moving to Kover or any other coverage report tools, or even if we remove it entirely.
cons:
 - 🤔

Tuubz: @sleepylee thanks for the feedback, I'll go with your suggestion.

I think we'll end up needing to execute 3 commands (`app`, `data` & `domain`)

sleepylee: 🤔  @Tuubz correct me if I'm wrong but doesn't `testDebugUnitTest` execute `:data` AND `:domain`?

Tuubz: @sleepylee No, it only runs for `:data` 😞

scripts/new_project.kts
```diff
@@ -248,7 +248,21 @@ object NewProject {
             "$projectPath${fileSeparator}gradlew",
             "-p",
             "$projectPath",
-            "testStagingDebugUnitTest"
+            ":app:testStagingDebugUnitTest"
```
sleepylee, line 251: then possibly you can use `&` instead of 3 executeCommand?

```suggestion
            ":app:testStagingDebugUnitTest & :data:testDebugUnitTest & :domain:test"
```

Tuubz: @sleepylee Thanks for the suggestion, looks a lot better now

Thank you 🎸

sleepylee (approved):

Issue #213, opened 2022-05-05
Apply to use Flow and kotlinx.Result instead of suspend fun in data layer and custom UseCaseResult


luongvo:
## Issue

Currently, we're using `suspend fun` to transfer the data from `Repository` through `UseCase` to `ViewModel`, then using try/catch to wrap the result which is `Success` or `Fail` in `UseCase` layer. We should migrate this to the new approach by applying [Flow](https://developer.android.com/kotlin/flow).

https://github.com/nimblehq/android-templates/blob/403bd3fa68bfc2c235ff7ccef89973a38b8c341e/CoroutineTemplate/domain/src/main/java/co/nimblehq/coroutine/domain/usecase/GetUsersUseCase.kt#L9-L16

## Solution

- Apply to use [Flow](https://developer.android.com/kotlin/flow) to transfer to transfer to data from `Repository` through `UseCase` to `ViewModel`.
- Here is a quick POC for everyone to checkout how it would be 👉  https://github.com/nimblehq/android-templates/commit/a1b599e9b61a235bc8e4af6cd472179114c915fe. This approach also help us avoid duplicating a new try/catch block in both `UseCase` and `Repository` when implementing https://github.com/nimblehq/android-templates/issues/212 as we can't access `retrofit objects` such as [HttpException](https://square.github.io/retrofit/2.x/retrofit/retrofit2/HttpException.html) in UseCase layer.

## Who Benefits?

Developers

I think this is **essential** because it will help us simplify the error mapping in `Repository` and other components too.

## What's Next?

- [x] Apply to use [Flow](https://developer.android.com/kotlin/flow) instead of `suspend fun`.
- [ ] Apply to use built-in [Result](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-result/) instead of custom [UseCaseResult](https://github.com/nimblehq/android-templates/blob/develop/CoroutineTemplate/domain/src/main/java/co/nimblehq/coroutine/domain/usecase/UseCaseResult.kt) to wrap `Success` or `Fail` result.
- [x] Update all ViewModel logics to adapt new structure.

ryan-conway:
About using `kotlinx.Result` instead of our own class, do we ever have a use case where we would want to emit another state (for example, `Loading`) before emitting `Success` or `Failure`? 🤔  For example, something like this:
```kotlin
fun execute(): Flow<UseCaseResult<List<Any>>> = flow {
    emit(UseCaseResult.Loading())
[...]

Issue #212, opened 2022-05-05
Add error mapping logic to handle custom errors from data layer


luongvo:
## Why

Currently, `CoroutineTemplate` does not have a logic to handle custom errors from `data` module, such as [HttpException](https://square.github.io/retrofit/2.x/retrofit/retrofit2/HttpException.html). For example, we need to know when is a validation error from API response happen to display a custom validation error on the UI 💡 .

We need to build an error mapping logic in `data` or `domain` layer to catch and pass the custom errors into the `app` module layer for other higher business implementations, instead of throwing a generic `Exception` to ViewModel.

https://github.com/nimblehq/android-templates/blob/403bd3fa68bfc2c235ff7ccef89973a38b8c341e/CoroutineTemplate/domain/src/main/java/co/nimblehq/coroutine/domain/usecase/GetUsersUseCase.kt#L9-L16

## Who Benefits?

Developers

I think this is **essential** because custom error handling is a required logic that almost apps need it.

luongvo:
@manh-t This is a related + small task too, we will need to catch and map custom error in Repository layer 👋

Tuubz:
@luongvo @manh-t Voted 👍 

Do you have any references for the solution? This could be useful for others to better understand the issue

Tuubz:
@luongvo 

Please correct me if I'm wrong, the error mapping for this issue is like you explained [here](https://github.com/nimblehq/android-templates/issues/213#issuecomment-1135375842), right? `flowErrorMapping()` 🙏

luongvo:
After discussing internally with @manh-t and @hoangnguyen92dn, I will help @manh-t to work on this ticket as it's quite related to the previous task #213 💪

luongvo:
We need to sync this to our sample projects in part 2.

PR #291, opened 2022-08-29
[#213] [1/2 - Template] Apply to use Flow + get rid of custom UseCaseResult

luongvo:
Resolves https://github.com/nimblehq/android-templates/issues/213

## What happened 👀

- Apply to use [Flow](https://developer.android.com/kotlin/flow) to transfer the data from `Repository` through `UseCase` to `ViewModel`.
- Get rid of duplicating a new **try/catch** block in both `UseCase` and `Repository` when implementing https://github.com/nimblehq/android-templates/issues/212 as we can't access `retrofit objects` such as [HttpException](https://square.github.io/retrofit/2.x/retrofit/retrofit2/HttpException.html) in UseCase layer.
- Get rid of using a custom `UseCaseResult` to transfer data and exception from Usecase to ViewModel.
- Update all ViewModel logics and samples to adapt new structure.

## Insight 📝

- I don't think we should apply to use a custom `UseCaseResult` or even built-in [Result](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-result/), we can rely on kotlin.Flow to catch the error or collect data flow properly. 🤓 

## Proof Of Work 📹

The `template` app should work properly after migrating.

### DIFF
--- a/template/app/src/main/java/co/nimblehq/template/ui/screens/compose/HomeComposeViewModel.kt
+++ b/template/app/src/main/java/co/nimblehq/template/ui/screens/compose/HomeComposeViewModel.kt
@@ -1,14 +1,12 @@
 package co.nimblehq.template.ui.screens.compose
 
 import co.nimblehq.template.domain.usecase.UseCase
-import co.nimblehq.template.domain.usecase.UseCaseResult
 import co.nimblehq.template.model.UiModel
 import co.nimblehq.template.model.toUiModels
 import co.nimblehq.template.ui.base.BaseViewModel
 import co.nimblehq.template.util.DispatchersProvider
 import dagger.hilt.android.lifecycle.HiltViewModel
-import kotlinx.coroutines.flow.MutableStateFlow
-import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.*
 import javax.inject.Inject
 
 @HiltViewModel
@@ -23,16 +21,17 @@ class HomeComposeViewModel @Inject constructor(
 
     init {
         execute {
-            when (val result = useCase.execute()) {
-                is UseCaseResult.Success -> {
-                    val uiModels = result.data.toUiModels()
-                    _uiModels.emit(uiModels)
-                }
-                is UseCaseResult.Error -> {
-                    val errorMessage = result.exception.message.orEmpty()
+            showLoading()
+            useCase.execute()
+                .catch {
+                    val errorMessage = it.message.orEmpty()
                     _error.emit(errorMessage)
                 }
-            }
+                .collect { result ->
+                    val uiModels = result.toUiModels()
+                    _uiModels.emit(uiModels)
+                }
+            hideLoading()
         }
     }
 }
--- a/template/data/src/main/java/co/nimblehq/template/data/repository/RepositoryImpl.kt
+++ b/template/data/src/main/java/co/nimblehq/template/data/repository/RepositoryImpl.kt
@@ -4,10 +4,19 @@ import co.nimblehq.template.data.response.toModels
 import co.nimblehq.template.data.service.ApiService
 import co.nimblehq.template.domain.model.Model
 import co.nimblehq.template.domain.repository.Repository
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.flow
 
 class RepositoryImpl constructor(
     private val apiService: ApiService
 ) : Repository {
 
-    override suspend fun getModels(): List<Model> = apiService.getResponses().toModels()
+    override fun getModels(): Flow<List<Model>> = flow {
+        try {
+            val result = apiService.getResponses().toModels()
+            emit(result)
+        } catch (e: Exception) {
+            throw e
+        }
+    }
 }
--- a/template/data/src/test/java/co/nimblehq/template/data/repository/RepositoryTest.kt
+++ b/template/data/src/test/java/co/nimblehq/template/data/repository/RepositoryTest.kt
@@ -9,6 +9,7 @@ import io.kotest.matchers.shouldBe
 import io.mockk.coEvery
 import io.mockk.mockk
 import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.flow.*
 import kotlinx.coroutines.test.runBlockingTest
 import org.junit.Before
 import org.junit.Test
@@ -32,13 +33,18 @@ class RepositoryTest {
         val expected = listOf(response)
         coEvery { mockService.getResponses() } returns expected
 
-        repository.getModels() shouldBe expected.toModels()
+        repository.getModels().collect {
+            it shouldBe expected.toModels()
+        }
     }
 
     @Test
     fun `When request failed, it returns error`() = runBlockingTest {
-        coEvery { mockService.getResponses() } throws Throwable()
+        val expected = Throwable()
+        coEvery { mockService.getResponses() } throws expected
 
-        shouldThrow<Throwable> { repository.getModels() }
+        repository.getModels().catch {
+            it shouldBe expected
+        }.collect()
     }
 }
--- a/template/domain/src/main/java/co/nimblehq/template/domain/usecase/UseCase.kt
+++ b/template/domain/src/main/java/co/nimblehq/template/domain/usecase/UseCase.kt
@@ -2,16 +2,12 @@ package co.nimblehq.template.domain.usecase
 
 import co.nimblehq.template.domain.model.Model
 import co.nimblehq.template.domain.repository.Repository
+import kotlinx.coroutines.flow.Flow
 import javax.inject.Inject
 
 class UseCase @Inject constructor(private val repository: Repository) {
 
-    suspend fun execute(): UseCaseResult<List<Model>> {
-        return try {
-            val response = repository.getModels()
-            UseCaseResult.Success(response)
-        } catch (e: IllegalStateException) {
-            UseCaseResult.Error(e)
-        }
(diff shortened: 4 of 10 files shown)


template/app/src/main/java/co/nimblehq/template/ui/screens/compose/HomeComposeViewModel.kt
```diff
@@ -23,16 +21,17 @@ class HomeComposeViewModel @Inject constructor(
...
     init {
         execute {
-            when (val result = useCase.execute()) {
-                is UseCaseResult.Success -> {
-                    val uiModels = result.data.toUiModels()
-                    _uiModels.emit(uiModels)
-                }
-                is UseCaseResult.Error -> {
-                    val errorMessage = result.exception.message.orEmpty()
+            showLoading()
```
Wadeewee, line 24: @luongvo Let's revert these changes. I saw you already updated in [this PR](https://github.com/nimblehq/android-templates/pull/292/files#diff-eb44200d9d6b03ba7a180a699c76534b141ad10297fb217d311a8fa4733a5a9d) So for this PR we can focus only the changes in `template.

template/data/src/main/java/co/nimblehq/template/data/repository/RepositoryImpl.kt
```diff
@@ -4,10 +4,19 @@ import co.nimblehq.template.data.response.toModels
...
     private val apiService: ApiService
 ) : Repository {
 
-    override suspend fun getModels(): List<Model> = apiService.getResponses().toModels()
+    override fun getModels(): Flow<List<Model>> = flow {
+        try {
+            val result = apiService.getResponses().toModels()
+            emit(result)
+        } catch (e: Exception) {
+            throw e
```
Wadeewee, line 19: No biggie, but better to name it more clearly.
```suggestion
        } catch (exception: Exception) {
            throw exception
```

template/domain/src/main/java/co/nimblehq/template/domain/usecase/UseCase.kt
```diff
@@ -2,16 +2,12 @@ package co.nimblehq.template.domain.usecase
...
 
-    suspend fun execute(): UseCaseResult<List<Model>> {
-        return try {
-            val response = repository.getModels()
-            UseCaseResult.Success(response)
-        } catch (e: IllegalStateException) {
-            UseCaseResult.Error(e)
-        }
+    fun execute(): Flow<List<Model>> {
+        return repository.getModels()
```
Wadeewee, line 11: @luongvo Hmm, Do we still need `UseCase` if we don't handle any of the results(`success/error`) in `UseCase`? I worried that it would have a lot of logic inside `VM` instead 🤔

template/app/src/main/java/co/nimblehq/template/ui/screens/compose/HomeComposeViewModel.kt
luongvo: @Wadeewee Yes, but it's in the template 😂

template/domain/src/main/java/co/nimblehq/template/domain/usecase/UseCase.kt
luongvo: @Wadeewee From what we can see from VM implementation, we have `catch` and `collect` to subscribe for success or error case already. We can leave the usecase back for some data preparation (request) or data consolidation (response) actions. 🤔

template/data/src/main/java/co/nimblehq/template/data/repository/RepositoryImpl.kt
ryan-conway: @luongvo Quick question, why did we wrap this in a try catch if we immediately throw the caught exception anyway? 🤔

template/data/src/test/java/co/nimblehq/template/data/repository/RepositoryTest.kt
```diff
@@ -9,6 +9,7 @@ import io.kotest.matchers.shouldBe
 import io.mockk.coEvery
```
Wadeewee, line 9: Please remove `import io.kotest.assertions.throwables.shouldThrow`

template/data/src/main/java/co/nimblehq/template/data/repository/RepositoryImpl.kt
luongvo: @ryan-conway if we throw it, we need to catch it at usecase or VM level.
- Catching at usecase level: it's hard for us to check the exception's body such as `HttpException` which can not access from **domain** module. Please checkout more insights here https://github.com/nimblehq/android-templates/issues/213#issuecomment-1135375842. 👋 


- Catching at VM level: we need to have 1 more try/catch to wrap the Flow collector. It's a bad idea I think

```
try{
              useCase.execute()
                .catch {
                   // catch error
                }
                .collect { 
                    // get value
                }
}catch (e: Exception){
   // catch error
}
```

ryan-conway: @luongvo Ah, so this is just preparation for the custom error mapping, right? 😄

luongvo: @ryan-conway exactly, we will have something like this

image

template/domain/src/main/java/co/nimblehq/template/domain/usecase/UseCase.kt
Wadeewee: > From what we can see from VM implementation, we have catch and collect to subscribe for success or error case already. It's very clear.

I understand this point: ☝️ @luongvo, but my concern is when handling the chain calls because the previous way with `UseCaseResult` in `UseCase`, we can inject another `UseCase,` So whenever it's `success/error` and we need to call something from another UseCase to get the result, Let's say we want to call some `API` when the error occurs So, by this way we can execute the next `UseCase` in current UseCase instead of filter in `VM` 🙏  It seems like using `UseCaseResult` it's more flexible, IMO🤔

luongvo: > So, by this way we can execute the next UseCase in current UseCase instead of filter in VM 🙏 It seems like using UseCaseResult it's more flexible, IMO🤔

@Wadeewee I think you're missing and remember how `flatMap` benefits us when working with RxJava. So switching to Flow, we have something called `flatMapConcat` to map multiple flows in a chain. Please checkout more sample here https://github.com/nimblehq/android-templates/issues/213#issuecomment-1135375842

Do you think which one is better, the left or the right side? 🤓 

image

Wadeewee: > Do you think which one is better, the left or the right side? 🤓

@luongvo If we handle the `Chain` call in `VM` like your example.  Surely, the right side without `UseCaseResult` is better, but Actually, we can handle it in `UseCase` instead of `VM,` So in `VM,` we will handle it only 
- if success -> emit result value to view or do something but not `execute another usecase.`
- If error -> emit error to view or do something but not `execute another usecase.` 

Currently, our template doesn't have complex logic, So I think it's good to go without `UseCaseResult,` but TBH, I'm not sure for the `Client` project that might have complex logic.

chornerman:
@luongvo I think there are some duplicate words in the description 😄 
Screen Shot 2565-09-01 at 12 38 39

ryan-conway (commented):
LGTM aside from the comments [here](https://github.com/nimblehq/android-templates/pull/291/files#r959131964) and [here](https://github.com/nimblehq/android-templates/pull/291/files#r959141388) 👍🏻

template/data/src/main/java/co/nimblehq/template/data/repository/RepositoryImpl.kt
sleepylee: better add TODO

chornerman (commented):
LGTM aside from Ryan's comment 😄

template/domain/src/main/java/co/nimblehq/template/domain/usecase/UseCase.kt
luongvo: @Wadeewee A usecase should do a single job only and it should not contain a chain logic there. We should keep the use case calls chain in VM layer instead.

template/data/src/test/java/co/nimblehq/template/data/repository/RepositoryTest.kt
luongvo: @Wadeewee Fixed in https://github.com/nimblehq/android-templates/pull/292/commits/0d2563c9390e5a2bdfc1ae122af7d966f7093594

template/data/src/main/java/co/nimblehq/template/data/repository/RepositoryImpl.kt
luongvo: @Wadeewee Fixed in https://github.com/nimblehq/android-templates/pull/292/commits/0d2563c9390e5a2bdfc1ae122af7d966f7093594

chornerman (approved):
LGTM 🚀

lydiasama:
Do we need to update the sample projects after merging this to template? @ryan-conway @luongvo

lydiasama (approved):

ryan-conway:
@lydiasama Yes, that will be done in #292  😄

ryan-conway (approved):
👌🏻

Wadeewee (approved):

PR #292, opened 2022-08-29
[#213] [2/2 - Sample] Apply to use Flow for Compose and XML samples

luongvo:
Resolves https://github.com/nimblehq/android-templates/issues/213

## What happened 👀

- Apply to use [Flow](https://developer.android.com/kotlin/flow) to transfer the data from `Repository` through `UseCase` to `ViewModel`.
- Get rid of duplicating a new **try/catch** block in both `UseCase` and `Repository` when implementing https://github.com/nimblehq/android-templates/issues/212 as we can't access `retrofit objects` such as [HttpException](https://square.github.io/retrofit/2.x/retrofit/retrofit2/HttpException.html) in UseCase layer.
- Get rid of using a custom `UseCaseResult` to transfer data and exception from Usecase to ViewModel.
- Update all ViewModel logics and samples to adapt new structure.

## Insight 📝

- I don't think we should apply to use a custom `UseCaseResult` or even built-in [Result](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-result/), we can rely on kotlin.Flow to catch the error or collect data flow properly. 🤓 

## Proof Of Work 📹

The sample apps should work properly after migrating.

Issue #163, opened 2022-03-22
Update Gradle to version 7+ 


Tuubz:
## Why

Change the Gradle version as following:
- Android Gradle Plugin version: `4.2.1` > `7.1.2` or higher
- Gradle version: `6.7.1` > `7.4.1` or higher

## Who Benefits?

Developers will benefit from faster building times and more: https://gradle.org/whats-new/gradle-7/

I think this is **essential**, as it will bring a good experiences for the developers themselves

sleepylee:
@Tuubz will there be any conflict or restriction with the android build tools, or Android studio version, java/kotlin version with this upgrade?

Tuubz:
@sleepylee I've quickly tried upgrading the Gradle version as mentioned in the pull request description.

I can build app successfully after making the following changes:

- Remove `versionCode` & `versionName` from `build.gradle.kts (:data:)` > [Read more](https://stackoverflow.com/questions/67803478/error-unresolved-reference-versioncode-in-build-gradle-kts)
- Upgrading to Java v11 is required > Most of us are still using Java v1.8 if I'm not mistaken

What do you think? 🙏

manh-t:
@Tuubz upgrading to Java 11 may cause the build to fail if we use a 3rd-party lib that is using Java 1.8 ([Jacoco for example](https://github.com/nimblehq/android-templates/pull/137/files#diff-6055d9b70d081d3d5d5e04ed372749f3663844b75ff1d9f91dd98a4e81a23508R115)), so we may need to recheck a bit, overall it's good 👍

Tuubz:
@manh-t Good thinking, I'll approve it for now but won't assign this to any sprint yet (until we finish #169) 👍

luongvo:
@Tuubz @hoangnguyen92dn We should consider upgrading Kotlin and Kotlin coroutines test versions in this task (or a separate task) as well.

![image](https://user-images.githubusercontent.com/16315358/163355510-ed76d0ef-15b5-4ccf-8754-d1156d9ceb68.png)

hoangnguyen92dn:
@luongvo 👌 Sounds good to me 👍

PR #311, opened 2022-09-15
[#212] Add flowTransform to map and handle custom errors as Flow in data layer

luongvo:
https://github.com/nimblehq/android-templates/issues/212

## What happened 👀

- Add `flowTransform` operator to map and handle custom errors logic (as `exceptions`) in `data` or `domain` layer to catch and pass the custom errors into the `app` module layer for other higher business implementations, instead of throwing a generic `Exception` to ViewModel.

## Insight 📝

- Build a custom `flowTransform {}` which basically inspired by `flow {}` operator.

## Proof Of Work 📹

Show us the implementation: screenshots, GIFs, etc.

### DIFF
--- a/template/data/src/main/java/co/nimblehq/template/data/extensions/ResponseMapping.kt
+++ b/template/data/src/main/java/co/nimblehq/template/data/extensions/ResponseMapping.kt
@@ -1,0 +1,53 @@
+package co.nimblehq.template.data.extensions
+
+import co.nimblehq.template.data.response.ErrorResponse
+import co.nimblehq.template.data.service.ApiException
+import co.nimblehq.template.data.service.NoConnectivityException
+import co.nimblehq.template.data.service.providers.MoshiBuilderProvider
+import com.squareup.moshi.JsonDataException
+import kotlinx.coroutines.flow.FlowCollector
+import kotlinx.coroutines.flow.flow
+import retrofit2.HttpException
+import java.io.IOException
+import java.io.InterruptedIOException
+import java.net.UnknownHostException
+import kotlin.experimental.ExperimentalTypeInference
+
+@Suppress("TooGenericExceptionCaught")
+@OptIn(ExperimentalTypeInference::class)
+fun <T> flowTransform(@BuilderInference block: suspend FlowCollector<T>.() -> T) = flow {
+    val result = try {
+        block()
+    } catch (e: Exception) {
+        throw e.mapError()
+    }
+    emit(result)
+}
+
+private fun Throwable.mapError(): Throwable {
+    return when (this) {
+        is UnknownHostException, is InterruptedIOException -> NoConnectivityException
+        is HttpException -> {
+            val errorResponse = parseErrorResponse(this)
+            ApiException(errorResponse, response())
+        }
+        else -> this
+    }
+}
+
+private fun parseErrorResponse(httpException: HttpException): ErrorResponse? {
+    val jsonString = httpException.response()?.errorBody()?.string()
+    return parseErrorResponse(jsonString)
+}
+
+private fun parseErrorResponse(jsonString: String?): ErrorResponse? {
+    return try {
+        val moshi = MoshiBuilderProvider.moshiBuilder.build()
+        val adapter = moshi.adapter(ErrorResponse::class.java)
+        adapter.fromJson(jsonString.orEmpty())
+    } catch (exception: IOException) {
+        null
+    } catch (exception: JsonDataException) {
+        null
+    }
+}
--- a/template/data/src/main/java/co/nimblehq/template/data/repository/RepositoryImpl.kt
+++ b/template/data/src/main/java/co/nimblehq/template/data/repository/RepositoryImpl.kt
@@ -1,23 +1,17 @@
 package co.nimblehq.template.data.repository
 
+import co.nimblehq.template.data.extensions.flowTransform
 import co.nimblehq.template.data.response.toModels
 import co.nimblehq.template.data.service.ApiService
 import co.nimblehq.template.domain.model.Model
 import co.nimblehq.template.domain.repository.Repository
 import kotlinx.coroutines.flow.Flow
-import kotlinx.coroutines.flow.flow
 
 class RepositoryImpl constructor(
     private val apiService: ApiService
 ) : Repository {
 
-    override fun getModels(): Flow<List<Model>> = flow {
-        try {
-            val result = apiService.getResponses().toModels()
-            emit(result)
-        } catch (exception: Exception) {
-            // TODO do error mapping here
-            throw exception
-        }
+    override fun getModels(): Flow<List<Model>> = flowTransform {
+        apiService.getResponses().toModels()
     }
 }
--- a/template/data/src/main/java/co/nimblehq/template/data/service/Exceptions.kt
+++ b/template/data/src/main/java/co/nimblehq/template/data/service/Exceptions.kt
@@ -1,0 +1,12 @@
+package co.nimblehq.template.data.service
+
+import co.nimblehq.template.data.response.ErrorResponse
+import retrofit2.HttpException
+import retrofit2.Response
+
+object NoConnectivityException : RuntimeException()
+
+data class ApiException(
+    val error: ErrorResponse?,
+    val response: Response<*>?
+) : HttpException(response)
--- a/template/data/src/main/java/co/nimblehq/template/data/response/ErrorResponse.kt
+++ b/template/data/src/main/java/co/nimblehq/template/data/response/ErrorResponse.kt
@@ -1,0 +1,8 @@
+package co.nimblehq.template.data.response
+
+import com.squareup.moshi.Json
+
+data class ErrorResponse(
+    @Json(name = "message")
+    val message: String
+)


template/data/src/main/java/co/nimblehq/template/data/service/Exceptions.kt
```diff
@@ -0,0 +1,12 @@
+package co.nimblehq.template.data.service
```
ryan-conway, line 1: Any reason why we didn't declare these exception classes at the domain level? 🤔 
Or do we assume that `UseCase` doesn't care what is being thrown and will let the `ViewModel` handle it? 😄

template/data/src/main/java/co/nimblehq/template/data/extensions/ResponseMapping.kt
```diff
@@ -0,0 +1,53 @@
...
+import kotlinx.coroutines.flow.flow
+import retrofit2.HttpException
+import java.io.IOException
+import java.io.InterruptedIOException
+import java.net.UnknownHostException
+import kotlin.experimental.ExperimentalTypeInference
+
+@Suppress("TooGenericExceptionCaught")
+@OptIn(ExperimentalTypeInference::class)
+fun <T> flowTransform(@BuilderInference block: suspend FlowCollector<T>.() -> T) = flow {
```
ryan-conway, line 18: Is the `@BuilderInference` annotation really necessary here? The method appears to work even without it 🤔 Not sure if I've missed anything 😄

template/data/src/main/java/co/nimblehq/template/data/extensions/ResponseMapping.kt
```diff
@@ -0,0 +1,53 @@
...
+        block()
+    } catch (e: Exception) {
+        throw e.mapError()
+    }
+    emit(result)
+}
+
+private fun Throwable.mapError(): Throwable {
+    return when (this) {
+        is UnknownHostException, is InterruptedIOException -> NoConnectivityException
```
ryan-conway, line 29: [Optional] This seems more readable IMO 😄 
```suggestion
        is UnknownHostException, 
        is InterruptedIOException -> NoConnectivityException
```

template/app/src/main/java/co/nimblehq/template/ui/ErrorMapping.kt
```diff
@@ -2,7 +2,11 @@ package co.nimblehq.template.ui
 
 import android.content.Context
 import co.nimblehq.template.R
+import co.nimblehq.template.data.service.ApiException
 
 fun Throwable.userReadableMessage(context: Context): String {
-    return context.getString(R.string.error_generic)
+    return when (this) {
+        is ApiException -> error?.message
+        else -> message
```
ryan-conway, line 10: [Optional] Shall we also add a case for `NoConnectivityException`? 😄 
```suggestion
    return when (this) {
        is ApiException -> error?.message
        is NoConnectivityException -> context.getString(R.string.error_no_connectivity)
        else -> message
```

ryan-conway (commented):
Will there also be a follow up PR to update the samples as well? 😄 

The rest LGTM 👍🏻

template/app/src/main/java/co/nimblehq/template/ui/ErrorMapping.kt
luongvo: @ryan-conway I would like to keep the pure concept of the template. 😄

luongvo:
> Will there also be a follow up PR to update the samples as well? 😄

@ryan-conway Yes, after finalizing the mapping logic in this PR. I updated the PR title 👌

template/data/src/main/java/co/nimblehq/template/data/extensions/ResponseMapping.kt
ryan-conway: Ah, so we don't have to declare the type explicitly 👍🏻 
So the reason this worked on my end is because it was still declared as 
`override fun getModels(): Flow<List<Model>> = flowTransform {` 😅

luongvo: @ryan-conway Updated the repository ✅

template/data/src/main/java/co/nimblehq/template/data/service/Exceptions.kt
luongvo: @ryan-conway Hmm true. Currently, the `ApiException` depends on `ErrorResponse` and `HttpException` from the `data` module. It can not be moved into `domain` module for usecase to access 💭 

I would like to look for more ideas @nimblehq/android-chapter, should we move it to `domain` for usecase to access? If yes, I think we will need
- Create a new Error model in `domain` and map `ErrorResponse` to `Error` model.
- Get rid of the base `HttpException`, the `ApiException` will contain Error model & other other fields such as response's `code` or `message`.

AVI5HEK: @luongvo since `:data` depends on `:domain`, I think moving the exceptions to `:domain` is fine. They can be used by both `:app` and `:data` components.

ryan-conway: @luongvo I agree, `domain` should have its own error models independent of any library or framework. For example, we may have something like this in `domain` module
```kotlin
class ApiException(
    val httpCode: Int,
    val apiMessage: String?
): Exception()
```
And the mapping will remain in `data` module
```kotlin
private fun Throwable.mapError(): Throwable {
    return when (this) {
        is UnknownHostException, 
        is InterruptedIOException -> NoConnectivityException
        is HttpException -> ApiException(code(), getErrorBody(this))
        else -> this
    }
}
```

template/data/src/main/java/co/nimblehq/template/data/extensions/ResponseMapping.kt
luongvo: @ryan-conway Java 1.8 does not support it yet. I reverted the Repository back to declare the type explicitly with `Flow<List<Model>>`. We can upgrade the Java to 11 and update this later.

template/data/src/main/java/co/nimblehq/template/data/repository/RepositoryImpl.kt
```diff
@@ -1,23 +1,17 @@
...
 
-    override fun getModels(): Flow<List<Model>> = flow {
-        try {
-            val result = apiService.getResponses().toModels()
-            emit(result)
-        } catch (exception: Exception) {
-            // TODO do error mapping here
-            throw exception
-        }
+    override fun getModels(): Flow<List<Model>> = flowTransform {
```
lydiasama, line 14: From [this discussion](https://github.com/nimblehq/android-templates/pull/311/files#r971788418), why don't you omit the explicit type here since you have the `@BuilderInference` annotation for this?
```suggestion
    override fun getModels() = flowTransform {
```

luongvo: @lydiasama As I said here https://github.com/nimblehq/android-templates/pull/311/files#r971836394, it requires Java to be updated to 11, and it requires the CI to use Java 11 also https://github.com/nimblehq/android-templates/blob/develop/.github/workflows/review_pull_request.yml#L13-L16.

Not sure if we really need to update to use Java 11 in this PR. However, I think it's good to keep the support for the next Java version with `@BuilderInference`.

ryan-conway: @luongvo Updating Java version seems out of scope for this and could introduce a whole bunch of issues ([for example](https://github.com/nimblehq/android-templates/issues/163#issuecomment-1081553940)). Let's stick with this solution as is for now 😉

template/data/src/main/java/co/nimblehq/template/data/service/Exceptions.kt
luongvo: @AVI5HEK @ryan-conway Fixed in https://github.com/nimblehq/android-templates/pull/311/commits/fd17e4a8d0fdf0a966bb6ff94168f4203917d2f3

ryan-conway (approved):
👌🏻

template/data/src/main/java/co/nimblehq/template/data/extensions/ResponseMapping.kt
```diff
@@ -0,0 +1,60 @@
...
+        else -> this
+    }
+}
+
+private fun parseErrorResponse(response: Response<*>?): ErrorResponse? {
+    val jsonString = response?.errorBody()?.string()
+    return parseErrorResponse(jsonString)
+}
+
+private fun parseErrorResponse(jsonString: String?): ErrorResponse? {
```
ryan-conway, line 50: Just realised, why did we separate these if this is the only place it is used? 🤔 
```suggestion
private fun parseErrorResponse(response: Response<*>?): ErrorResponse? {
    val jsonString = response?.errorBody()?.string()
```

luongvo: @ryan-conway I think separating the logic into smaller step is a good thing. As my experience, we will need `private fun parseErrorResponse(jsonString: String?): ErrorResponse?` to parse the raw error response string in `Authentication` layer. But yeah, let's do it later when needed. Let me update 👍

lydiasama (approved):

ryan-conway (approved):

lydiasama:
@luongvo I don't think we should link the issue now because when we merge this PR, it will close the issue immediately. But it seems you still have [Part2](#292), right? How about linking it when it gets merged? cc @ryan-conway

ryan-conway:
@lydiasama agreed. @luongvo If you prefer I can unlink it before merging, then link it back in after merging 😄

luongvo:
> @lydiasama agreed. @luongvo If you prefer I can unlink it before merging, then link it back in after merging 😄

@ryan-conway yes please

luongvo:
@ryan-conway please note that we need to have more reviews and merge the base branch first

ryan-conway:
@luongvo Yep, waiting on at least one more review from @nimblehq/android-chapter 😉

luongvo:
@hoangnguyen92dn @manh-t This needs your review as well, besides others in @nimblehq/android-chapter :wave:

template/data/src/main/java/co/nimblehq/template/data/extensions/ResponseMapping.kt
```diff
@@ -0,0 +1,56 @@
...
+import kotlinx.coroutines.flow.flow
+import retrofit2.HttpException
+import retrofit2.Response
+import java.io.IOException
+import java.io.InterruptedIOException
+import java.net.UnknownHostException
+import kotlin.experimental.ExperimentalTypeInference
+
+@Suppress("TooGenericExceptionCaught")
+@OptIn(ExperimentalTypeInference::class)
```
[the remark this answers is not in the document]
luongvo: @chornerman We [won't need to define it in Gradle level from Kotlin 1.7.0](https://kotlinlang.org/docs/whatsnew17.html#stable-opt-in-requirements). As it's just a warning, I think we can ignore and wait for the Kotlin version upgrading soon 🚀 

More detail about `OptIn` annotation here https://dev.to/vtsen/how-to-add-opt-in-compiler-argument-in-buildgradle-jb3

template/data/src/main/java/co/nimblehq/template/data/extensions/ResponseMapping.kt
```diff
@@ -0,0 +1,56 @@
...
+fun <T> flowTransform(@BuilderInference block: suspend FlowCollector<T>.() -> T) = flow {
+    val result = try {
+        block()
+    } catch (e: Exception) {
+        throw e.mapError()
```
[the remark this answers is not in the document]
luongvo: @chornerman Fixed in https://github.com/nimblehq/android-templates/pull/311/commits/01da5550d6fe018c5869acad8f542bff0a6e1acf

template/data/src/main/java/co/nimblehq/template/data/extensions/ResponseMapping.kt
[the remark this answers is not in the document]
chornerman: I see, thanks for sharing this @luongvo 🙏🏼

chornerman (approved):
lgtm 🚀

PR #313, opened 2022-10-03
[Chore] Update versions

Tuubz:
## What happened 👀

- Bumped version from `3.12.0` to `3.13.0` for sample projects & script
- Set version to `1.0.0` for template

### DIFF
--- a/template/buildSrc/src/main/java/Versions.kt
+++ b/template/buildSrc/src/main/java/Versions.kt
@@ -6,7 +6,7 @@ object Versions {
     const val ANDROID_TARGET_SDK_VERSION = 30
 
     const val ANDROID_VERSION_CODE = 1
-    const val ANDROID_VERSION_NAME = "3.12.0"
+    const val ANDROID_VERSION_NAME = "1.0.0"
 
     // Dependencies (Alphabet sorted)
     const val ANDROID_COMMON_KTX_VERSION = "0.1.1"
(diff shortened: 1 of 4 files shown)


lydiasama (approved):

template/buildSrc/src/main/java/Versions.kt
```diff
@@ -6,7 +6,7 @@ object Versions {
     const val ANDROID_TARGET_SDK_VERSION = 30
 
     const val ANDROID_VERSION_CODE = 1
-    const val ANDROID_VERSION_NAME = "3.12.0"
+    const val ANDROID_VERSION_NAME = "3.13.0"
```
luongvo, line 9: Since we have the version for template scripting, do we still need to bump the `template` project & `samples` manually for each new iteration? Does it bring any sense? 🤔

Tuubz: I'll change this to `1.0.0`, the rest we'll keep for now. In a future pull request, we'll rename `templateScriptVersion` to `templateVersion` so it can be reused for the sample projects too 🚀 

Reference: https://nimble-co.slack.com/archives/G41UCTGHL/p1664770189337109

Tuubz: @luongvo Updated 🤖

kaungkhantsoe (approved):

chornerman (approved):

luongvo (approved):

PR #315, opened 2022-10-03
[#212] [2/2 - Sample] Add flowTransform to map and handle custom errors as Flow in data layer

luongvo:
https://github.com/nimblehq/android-templates/issues/212

## What happened 👀

- Bring logic updates in [this PR](https://github.com/nimblehq/android-templates/pull/311) from the `template` project into `sample` projects.

## Insight 📝

Add `flowTransform` to map and handle custom errors as Flow in data layer for:
- Compose sample.
- XML sample.

## Proof Of Work 📹

N/A

### DIFF
--- a/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/ErrorMapping.kt
+++ b/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/ErrorMapping.kt
@@ -2,7 +2,11 @@ package co.nimblehq.sample.compose.ui
 
 import android.content.Context
 import co.nimblehq.sample.compose.R
+import co.nimblehq.sample.compose.domain.exceptions.ApiException
 
 fun Throwable.userReadableMessage(context: Context): String {
-    return context.getString(R.string.error_generic)
+    return when (this) {
+        is ApiException -> error?.message
+        else -> message
+    } ?: context.getString(R.string.error_generic)
 }
--- a/sample-compose/data/src/main/java/co/nimblehq/sample/compose/data/extensions/ResponseMapping.kt
+++ b/sample-compose/data/src/main/java/co/nimblehq/sample/compose/data/extensions/ResponseMapping.kt
@@ -1,0 +1,56 @@
+package co.nimblehq.sample.compose.data.extensions
+
+import co.nimblehq.sample.compose.data.response.ErrorResponse
+import co.nimblehq.sample.compose.data.response.toModel
+import co.nimblehq.sample.compose.data.service.providers.MoshiBuilderProvider
+import co.nimblehq.sample.compose.domain.exceptions.ApiException
+import co.nimblehq.sample.compose.domain.exceptions.NoConnectivityException
+import com.squareup.moshi.JsonDataException
+import kotlinx.coroutines.flow.FlowCollector
+import kotlinx.coroutines.flow.flow
+import retrofit2.HttpException
+import retrofit2.Response
+import java.io.IOException
+import java.io.InterruptedIOException
+import java.net.UnknownHostException
+import kotlin.experimental.ExperimentalTypeInference
+
+@Suppress("TooGenericExceptionCaught")
+@OptIn(ExperimentalTypeInference::class)
+fun <T> flowTransform(@BuilderInference block: suspend FlowCollector<T>.() -> T) = flow {
+    val result = try {
+        block()
+    } catch (exception: Exception) {
+        throw exception.mapError()
+    }
+    emit(result)
+}
+
+private fun Throwable.mapError(): Throwable {
+    return when (this) {
+        is UnknownHostException,
+        is InterruptedIOException -> NoConnectivityException
+        is HttpException -> {
+            val errorResponse = parseErrorResponse(response())
+            ApiException(
+                errorResponse?.toModel(),
+                code(),
+                message()
+            )
+        }
+        else -> this
+    }
+}
+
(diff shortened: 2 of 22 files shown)


Tuubz (approved):

manh-t:
LGTM 💯

manh-t (approved):

Tuubz:
@nimblehq/android-chapter One more review needed here, por favor 🙏 🇪🇸

sample-compose/domain/src/main/java/co/nimblehq/sample/compose/domain/exceptions/Exceptions.kt
```diff
@@ -0,0 +1,11 @@
+package co.nimblehq.sample.compose.domain.exceptions
+
+import co.nimblehq.sample.compose.domain.model.Error
+
+object NoConnectivityException : RuntimeException()
+
+data class ApiException(
+    val error: Error?,
```
lydiasama, line 8: I have a question. 🙋🏻  While the Error model has only `message: String`, why don't we declare the error property in `ApiException` as just String?
```suggestion
    val errorMessage: String,
```

sample-compose/data/src/main/java/co/nimblehq/sample/compose/data/response/ErrorResponse.kt
```diff
@@ -0,0 +1,11 @@
...
+
+import co.nimblehq.sample.compose.domain.model.Error
+import com.squareup.moshi.Json
+
+data class ErrorResponse(
+    @Json(name = "message")
+    val message: String
+)
+
+internal fun ErrorResponse.toModel() = Error(message = message)
```
lydiasama, line 11: Then we can get rid of this `toModel` extension.

sample-compose/data/src/main/java/co/nimblehq/sample/compose/data/extensions/ResponseMapping.kt
```diff
@@ -0,0 +1,56 @@
...
+}
+
+private fun Throwable.mapError(): Throwable {
+    return when (this) {
+        is UnknownHostException,
+        is InterruptedIOException -> NoConnectivityException
+        is HttpException -> {
+            val errorResponse = parseErrorResponse(response())
+            ApiException(
+                errorResponse?.toModel(),
```
lydiasama, line 36: Then it will become to this:
```suggestion
                errorResponse?.message,
```

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/ErrorMapping.kt
```diff
@@ -2,7 +2,11 @@ package co.nimblehq.sample.compose.ui
 
 import android.content.Context
 import co.nimblehq.sample.compose.R
+import co.nimblehq.sample.compose.domain.exceptions.ApiException
 
 fun Throwable.userReadableMessage(context: Context): String {
-    return context.getString(R.string.error_generic)
+    return when (this) {
+        is ApiException -> error?.message
```
doannimble, line 9: @luongvo Should we add the default error message in case of null? 🤔

lydiasama, line 9: And this will be:
```suggestion
        is ApiException -> errorMessage
```

kaungkhantsoe (approved):

luongvo:
@lydiasama Friendly reminder, this PR is the second part to apply what we did with the new approach from the PR #311 to sample projects. It should be the same to the template, any new logic would be considered carefully. :tipping_hand_man:

lydiasama:
@luongvo So we should let this pass, right? Then you will get a 100% review passed. 😅

lydiasama (approved):

luongvo:
> @luongvo So we should let this pass, right? Then you will get a 100% review passed. sweat_smile

I think so @lydiasama :pray: . I will review your reviews separately 👋

lydiasama:
@luongvo Next time, Let's use the word "Please approve this PR" instead of review. 😂

sample-compose/domain/src/main/java/co/nimblehq/sample/compose/domain/exceptions/Exceptions.kt
luongvo: @lydiasama As it's the template, the `Error` model could be more complex in real projects so we would like to keep the bare logic for future implementation,

sample-compose/data/src/main/java/co/nimblehq/sample/compose/data/response/ErrorResponse.kt
luongvo: @lydiasama explained above ☝️

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/ErrorMapping.kt
luongvo: @doannimble Yes, already, could you please re-check that logic again as we're falling back the `error_generic` after a `when` block.

https://github.com/nimblehq/android-templates/blob/58c443a96406175f99cd7abcb2997f3efc4d9f5c/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/ErrorMapping.kt#L7-L12

PR #321, opened 2022-10-11
[#262] Update use case to use operator invoke function instead of old…

### LINKED ISSUE #262: Update execute function for use case
## Why

In Kotlin we can [override the invoke method](https://riptutorial.com/kotlin/example/31101/overriding-invoke-method-to-build-dsl). 

With this approach, we can eliminate `.execute()` and use the invoke function instead

---

Ex:

If we add this to our use case:
```
interface DoSomethingUseCase {
    suspend operator fun invoke(id: String)
}
```

Then we can call the use case without `.execute()`:
```
DoSomethingUseCase(id)
```

Reference: 
- https://nimble-co.slack.com/archives/G41UCTGHL/p1657876809735139
- https://github.com/nimblehq/android-templates/discussions/257

## Who Benefits?

I think this is **essential** for us developers because it will bring a good experience for us.

kaungkhantsoe:
Closes #262 

## What happened 👀

In Kotlin we can [override the invoke method](https://riptutorial.com/kotlin/example/31101/overriding-invoke-method-to-build-dsl). 

With this approach, we can eliminate `.execute()` and use the invoke function instead

---

Ex:

If we add this to our use case:
```
interface DoSomethingUseCase {
    suspend operator fun invoke(id: String)
}
```

Then we can call the use case without `.execute()`:
```
DoSomethingUseCase(id)
```

## Insight 📝

- Refactor `UseCase` classes to use `operator fun invoke()` instead of `fun execute()`
- Refactor classes which calls `execute()`

## Proof Of Work 📹

https://user-images.githubusercontent.com/32578035/195009412-a9a840f6-bcb8-4378-a53c-cb29e23087f7.mp4

### DIFF
--- a/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeComposeViewModel.kt
+++ b/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeComposeViewModel.kt
@@ -24,7 +24,7 @@ class HomeComposeViewModel @Inject constructor(
     init {
         execute {
             showLoading()
-            useCase.execute()
+            useCase()
                 .catch {
                     val errorMessage = it.message.orEmpty()
                     _error.emit(errorMessage)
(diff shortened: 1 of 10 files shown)


Tuubz (approved):
💪 💪 💪

sleepylee:
you'll need to update the Test as well because inside there, we are still calling `execute()`

Tuubz:
@sleepylee Hmmm, which files are you referring to? From what I see the tests have been updated already 🙏

sleepylee:
ahh nvm - I compared with a wrong version 🙏 sorry for the confusion @Tuubz @kaungkhantsoe it should be good now 💯

sleepylee (approved):

Issue #191, opened 2022-04-21
Update Kotlin Version


Tuubz:
## Why

Change the Kotlin version:
- `kotlin-stdlib` : 1.5.21 > 1.7.0 or higher
- `kotlinx-coroutines-android` : 1.5.0 > 1.6.3 or higher
- `kotlinx-coroutines-test` : 1.5.0 > 1.6.3 or higher
- `kotlin-reflect` : 1.5.10 > 1.7.0 or higher

## Who Benefits?

Developers will benefit from new features and performance improvements: 
- https://github.com/JetBrains/kotlin/releases
- https://github.com/Kotlin/kotlinx.coroutines/releases
- https://kotlinlang.org/docs/whatsnew16.html

I think this is essential, as it will bring a good experience for the developers themselves

PR #322, opened 2022-10-13
[#191][Part2] Update Kotlin Version

doannimble:
Closes https://github.com/nimblehq/android-templates/issues/191

## What happened 👀

- Upgrade Kotlin version
- Upgrade Build gradle version
- Upgrade Android compiler/target version
- Upgrade Compose and Constraint layout version
- Upgrade Hilt and Test Runner version

## Insight 📝

- Upgrade Kotlin to 1.7.2, we also have to  upgrade Build Gradle (7+) and Android Compiler/Target Version to the latest (33)
- Add `exported` tag into activity manifest
- Update ViewModelExt to match with the latest ktx
- Add fragment ktx
- > Our app should be built and run normally after upgrading

## Proof Of Work 📹

image

Screenshot_1665669876


Tuubz:
@doannimble Thanks for the update 💪 

However, we can't merge this if the CI is failing, or we won't know if the changes broke anything.

I didn't add this issue to the current sprint, because there were issues with *Jacoco* when upgrading to Gradle 7 ☠️ 

If you can fix the CI, we can consider adding it to the current release. 

But next time, please communicate with the repository maintainer, so we can commit to [the release plan](https://github.com/orgs/nimblehq/projects/3)

luongvo:
@doannimble are you working on an unplanned issue this sprint? Cc @Tuubz

Tuubz:
@doannimble The code coverage has dropped to `0.66%` 😨 

I think we could try to switch to Kover first, before merging these changes. 

What do you think? If we agree, I'll create a ticket for this 🙏

Issue #261, opened 2022-08-01
Add permission request handler


Tuubz:
## Why

[RxPermissions](https://github.com/tbruyelle/RxPermissions) is used for RxJavaTemplate. 

In our new template, we currently don't have a permissions request handler, so let's add [Peko](https://github.com/deva666/Peko) to support `Coroutine & Flow`

Reference: https://nimble-co.slack.com/archives/G41UCTGHL/p1657594027286929

## Who Benefits?

I think this is **essential** for us developers because it will bring a good experience for us.

sleepylee:
maybe start with a prefix `[RFC]` to help improve its visibility? 💡

Tuubz:
@sleepylee I'm glad you raised this 👍 In my opinion, adding `[RFC]` is quite redundant, because every feature or chore needs to go through a voting process anyways.

What do you think? Please correct me if I'm wrong, but I don't see any added value of the prefix 💭

sleepylee:
it's your call, my point of view is from the backlog standpoint, as you introduced various different formats, I'm just suggesting a better way to manage their visibility better. But if you think you can manage it and so do the others, then you can skip it.
![image](https://user-images.githubusercontent.com/13435717/182080968-4589db5c-9204-4d82-804b-86ed3d36d42b.png)

sleepylee:
and to do or to skip, please clarify so the others can follow up our conversation, just a reminder in an opposing way to an emoji, @Tuubz 😇

Tuubz:
My bad. 

The idea for having `[xxx]` in front of the issue title was to reference the epic of that issue (for better visibility). But after discussing with Julien on [GitHub Projects Handbook](https://github.com/nimblehq/compass/issues/985), we agreed to add a new field for it in the project board instead:


![Screen Shot 2565-08-01 at 14 40 29](https://user-images.githubusercontent.com/18277915/182098095-76506901-8764-4ba9-92af-5edaaf1f05b7.png)


So I'll stop adding `[Sample]`, `[CoroutineTemplate]`, ... from now on 🙏

PR #324, opened 2022-10-19
[#261] Add permission request handler

ryan-conway:
https://github.com/nimblehq/android-templates/issues/261

## What happened 👀

- Added `Peko` dependency to template
- Updated XML and Compose samples
  - Added `requestPermissions` `StateFlow` and `onPermissionResult()` to `HomeViewModel` and `HomeComposeViewModel`
  - Added `requestPermissions()` method to `HomeFragment` and `HomeComposeFragment`

## Insight 📝

Since the template currently targets `SDK 30` instead of `SDK 31`, we must use an older version of Peko (Current stable is `2.2.0`, we are using `2.1.2`). This version makes use of Coroutines and LiveData, and the library must be called from an Activity or Fragment (We may call `Peko.requestPermissionsAsync()` if we have a Context in our ViewModel, but this could lead to leaks)

If we wish to use Flow with permission requests, we should use Peko `3.X` (latest is `3.0.0-ALPHA-01`). This would require updating target SDK version to `31`, as well as updating Kotlin and Coroutines version to `v1.6.21`.

The logic and code is the same for both XML and Compose, will update the `HomeComposeFragment` and `HomeComposeViewModel` once code for XML is finalized 👍🏻 

## Proof Of Work 📹

[untitled.webm](https://user-images.githubusercontent.com/53168251/196604648-ab9314c4-a5e8-4adc-abf8-81255c42e225.webm)

### DIFF
--- a/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeComposeFragment.kt
+++ b/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeComposeFragment.kt
@@ -1,6 +1,7 @@
 package co.nimblehq.sample.compose.ui.screens.home
 
-import android.Manifest
+import android.Manifest.permission.ACCESS_FINE_LOCATION
+import android.Manifest.permission.CAMERA
 import android.os.Bundle
 import android.view.View
 import androidx.compose.runtime.Composable
@@ -37,10 +38,7 @@ class HomeComposeFragment : BaseComposeFragment() {
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
-        requestPermissions(
-            Manifest.permission.ACCESS_FINE_LOCATION,
-            Manifest.permission.CAMERA
-        )
+        requestPermissions(ACCESS_FINE_LOCATION, CAMERA)
     }
 
     override fun bindViewModel() {
@@ -51,11 +49,11 @@ class HomeComposeFragment : BaseComposeFragment() {
     private fun requestPermissions(vararg permissions: String) {
         lifecycleScope.launch {
             val result = requestPermissionsAsync(*permissions)
-            onPermissionResult(result)
+            handlePermissionResult(result)
         }
     }
 
-    private fun onPermissionResult(permissionResult: PermissionResult) {
+    private fun handlePermissionResult(permissionResult: PermissionResult) {
         when (permissionResult) {
             is Granted -> Timber.d("${permissionResult.grantedPermissions} granted")
             is Denied.NeedsRationale -> Timber.d("${permissionResult.deniedPermissions} needs rationale")
--- a/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeComposeFragment.kt
+++ b/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeComposeFragment.kt
@@ -2,10 +2,13 @@ package co.nimblehq.sample.compose.ui.screens.home
 
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.collectAsState
+import androidx.lifecycle.lifecycleScope
 import co.nimblehq.sample.compose.extension.provideViewModels
 import co.nimblehq.sample.compose.ui.base.BaseComposeFragment
 import co.nimblehq.sample.compose.ui.screens.MainNavigator
+import com.markodevcic.peko.requestPermissionsAsync
 import dagger.hilt.android.AndroidEntryPoint
+import kotlinx.coroutines.launch
 import javax.inject.Inject
 
 @AndroidEntryPoint
@@ -28,5 +31,13 @@ class HomeComposeFragment : BaseComposeFragment() {
     override fun bindViewModel() {
         viewModel.error bindTo ::displayError
         viewModel.navigator bindTo navigator::navigate
+        viewModel.requestPermissions bindTo ::requestPermissions
+    }
+
+    private fun requestPermissions(vararg permissions: String) {
+        lifecycleScope.launch {
+            val result = requestPermissionsAsync(*permissions)
+            viewModel.onPermissionResult(result)
+        }
     }
 }
--- a/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeComposeFragment.kt
+++ b/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeComposeFragment.kt
@@ -1,14 +1,21 @@
 package co.nimblehq.sample.compose.ui.screens.home
 
+import android.Manifest
+import android.os.Bundle
+import android.view.View
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.collectAsState
 import androidx.lifecycle.lifecycleScope
 import co.nimblehq.sample.compose.extension.provideViewModels
 import co.nimblehq.sample.compose.ui.base.BaseComposeFragment
 import co.nimblehq.sample.compose.ui.screens.MainNavigator
+import com.markodevcic.peko.PermissionResult
+import com.markodevcic.peko.PermissionResult.Denied
+import com.markodevcic.peko.PermissionResult.Granted
 import com.markodevcic.peko.requestPermissionsAsync
 import dagger.hilt.android.AndroidEntryPoint
 import kotlinx.coroutines.launch
+import timber.log.Timber
 import javax.inject.Inject
 
 @AndroidEntryPoint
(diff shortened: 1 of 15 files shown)


template/app/src/main/AndroidManifest.xml
```diff
@@ -3,6 +3,8 @@
     package="co.nimblehq.template">
 
     <uses-permission android:name="android.permission.INTERNET" />
+    <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
+    <uses-permission android:name="android.permission.CAMERA" />
```
Tuubz, line 7: This feels like sample code, would you agree? 💭

template/app/src/main/java/co/nimblehq/template/ui/screens/xml/HomeViewModel.kt
```diff
@@ -31,6 +39,27 @@ class HomeViewModel @Inject constructor(
...
+        )
+    }
+
+    fun onPermissionResult(permissionResult: PermissionResult) {
+        when (permissionResult) {
+            is Granted -> Timber.d("${permissionResult.grantedPermissions} granted")
+            is Denied.JustDenied -> Timber.d("${permissionResult.deniedPermissions} denied")
+            is Denied.NeedsRationale -> Timber.d("${permissionResult.deniedPermissions} needs rationale")
+            is Denied.DeniedPermanently -> Timber.d("${permissionResult.deniedPermissions} denied for good")
+            is Cancelled -> Timber.d("request cancelled")
```
Tuubz, line 62: Same here, I'm wondering if we should only keep the dependency in `template` and move the rest into the `samples` 

What do you think? 🙏

Tuubz (commented):
Rest LGTM 💪

template/app/src/main/AndroidManifest.xml
ryan-conway: Good point 👍🏻 Shall we also update `HomeViewModel.requestPermissions()` to emit an empty array instead? Or remove that method entirely? 🤔

Tuubz: I'm fine to keep `HomeViewModel.requestPermissions()` as it is (for the `samples`) 💪

ryan-conway: If thats the case then running the project after generating could lead to a crash because those permissions weren't declared in the manifest 🤔

Tuubz: Can we do this?

- `template`: Only define the dependency, nothing else
- `samples`: Define dependency, declare permissions, integrate home-fragment & home-view model

Please let me know if I got the wrong idea here 🙇

ryan-conway: That could definitely work 👀 Will apply it like that 😄

template/app/src/main/AndroidManifest.xml
```diff
@@ -14,7 +14,7 @@
         android:supportsRtl="true"
         android:theme="@style/AppTheme">
 
-        <activity android:name=".ui.screens.MainActivity">
+        <activity android:name=".ui.screens.MainActivity" android:exported="true">
```
Tuubz, line 17: Is this change necessary for this pull request? 🙏

Tuubz (approved):
Rest LGTM 🎸

sample-xml/app/src/main/java/co/nimblehq/sample/xml/ui/screens/home/HomeViewModel.kt
```diff
@@ -32,10 +40,31 @@ class HomeViewModel @Inject constructor(
...
+
+    fun onPermissionResult(permissionResult: PermissionResult) {
+        when (permissionResult) {
+            is Granted -> Timber.d("${permissionResult.grantedPermissions} granted")
+            is Denied.JustDenied -> Timber.d("${permissionResult.deniedPermissions} denied")
+            is Denied.NeedsRationale -> Timber.d("${permissionResult.deniedPermissions} needs rationale")
+            is Denied.DeniedPermanently -> Timber.d("${permissionResult.deniedPermissions} denied for good")
+            is Cancelled -> Timber.d("request cancelled")
+        }
+    }
```
luongvo, line 69: I personally think permission request is a UI action, so the logic to request permission should be in the View layer 💭 What do you think? What's the benefit of keeping it in VM layer?

luongvo: `onPermissionResult` could be in VM layer 💭

lydiasama (approved):

sample-xml/app/src/main/java/co/nimblehq/sample/xml/ui/screens/home/HomeViewModel.kt
```diff
@@ -20,6 +24,10 @@ class HomeViewModel @Inject constructor(
     val uiModels: StateFlow<List<UiModel>>
         get() = _uiModels
 
+    private val _requestPermissions = MutableStateFlow<Array<out String>>(emptyArray())
+    val requestPermissions: StateFlow<Array<out String>>
+        get() = _requestPermissions
```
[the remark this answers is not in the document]
ryan-conway: Shouldn't the view be kept as simple as possible with all logic in the ViewModel? 🤔 
Suppose we have a button that needs to check `LOCATION` permission, so the view's code would look something like
```kotlin
button.setOnClickListener {
    if (hasPermission(LOCATION)) {
         doSomething()
    } else {
        requestPermission(LOCATION)
    }
}
```
instead of just 
```kotlin
button.setOnClickListener { viewModel.doSomething() }
```
Please correct me if I've misunderstood something 😅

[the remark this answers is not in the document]
Tuubz: What if we do this in `HomeFragment.kt?` So no code in `HomeViewmodel.kt` as @luongvo suggested

If the user already has permission, it will go to `Granted`, then we can still call `viewModel.doSomething()` if needed.

```
override fun setupView() {
      button.setOnClickListener { 
         requestPermissions(
               Manifest.permission.ACCESS_FINE_LOCATION,
               Manifest.permission.CAMERA
          ) 
     }
}

private fun requestPermissions(vararg permissions: String) {
    lifecycleScope.launch {
        val result = requestPermissionsAsync(*permissions)
        handlePermissionResult(result)
    }
}

private fun handlePermissionResult(permissionResult: PermissionResult) {
    when (permissionResult) {
        is Granted -> Timber.d("${permissionResult.grantedPermissions} granted")
        is Denied.JustDenied -> Timber.d("${permissionResult.deniedPermissions} denied")
        is Denied.NeedsRationale -> Timber.d("${permissionResult.deniedPermissions} needs rationale")
        is Denied.DeniedPermanently -> Timber.d("${permissionResult.deniedPermissions} denied for good")
        is Cancelled -> Timber.d("request cancelled")
    }
}

```

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeComposeViewModel.kt
```diff
@@ -33,10 +41,31 @@ class HomeComposeViewModel @Inject constructor(
...
+        )
+    }
+
+    fun onPermissionResult(permissionResult: PermissionResult) {
+        when (permissionResult) {
+            is Granted -> Timber.d("${permissionResult.grantedPermissions} granted")
+            is Denied.JustDenied -> Timber.d("${permissionResult.deniedPermissions} denied")
+            is Denied.NeedsRationale -> Timber.d("${permissionResult.deniedPermissions} needs rationale")
+            is Denied.DeniedPermanently -> Timber.d("${permissionResult.deniedPermissions} denied for good")
+            is Cancelled -> Timber.d("request cancelled")
```
Tuubz, line 68: @ryan-conway Any idea how to trigger `Cancelled` & `JustDenied`? I can trigger the rest correctly though

ryan-conway: According the the [readme for v2.1.2](https://github.com/deva666/Peko/tree/release/2.1.2)
>PermissionResult.Denied.JustDenied -> subclass of PermissionResult.Denied, returned when previous two cases are not the cause, for example if you forget to register the Permission in AndroidManifest
PermissionResult.Cancelled -> returned when Android System cancels the request, ie returned

Tuubz: Nice, thank you 🙏 

Also, when I tap outside the permission dialog, it results in `DeniedPermanently`.

Not what I was expecting... What do you think? 💭

ryan-conway: Looks like this is an [existing issue](https://github.com/deva666/Peko/issues/21) with the library 👀 
Not sure if this has been fixed in the latest version, but even if it was we can't use it anyway until we update our dependencies 🥲 
Given that, shall we still continue to use the library? Seems like a pretty big issue to have considering this is meant to be a good template to use 😅

Tuubz: Ughhhh, that sucks 😭 

I'm afraid we'll have to put this on hold for now and wait until we can upgrade to `SDK 31` (Probably in December)

Hopefully, the issue has been fixed by then.

If you agree, you can put this pull request as a draft 🪦

ryan-conway: Agreed, will draft it for now 🥲

Tuubz:
@luongvo Any last word on this pull request? 🤩

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeComposeFragment.kt
```diff
@@ -25,8 +35,31 @@ class HomeComposeFragment : BaseComposeFragment() {
             )
         }
 
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+        requestPermissions(
+            Manifest.permission.ACCESS_FINE_LOCATION,
+            Manifest.permission.CAMERA
```
chornerman, line 42: (Optional) Use static import
```suggestion
            ACCESS_FINE_LOCATION,
            CAMERA
```

chornerman (commented):
The rest LGTM

Tuubz:
@chornerman I've updated the pull request, since @ryan-conway will be off today 🚀

chornerman (approved):
LGTM 🚀

luongvo (approved):
LGTM

Issue #230, opened 2022-06-02
Update `publish_docs_to_wiki` workflow to use `publish_wiki` of github_action_workflow repository


manh-t:
## Why

We have a `github_action_workflow` repository to maintain the reusable workflows, including [publish_wiki.yml](https://github.com/nimblehq/github-actions-workflows/blob/develop/.github/workflows/publish_wiki.yml) workflow. It's better not to fragment the workflows and maintain it ourself.

Updating `publish_docs_to_wiki` to extend and use the actions in [publish_wiki.yml](https://github.com/nimblehq/github-actions-workflows/blob/develop/.github/workflows/publish_wiki.yml).

## Who Benefits?

Developers.

Tuubz:
@nimblehq/android-chapter Could I get more votes on this one, please? 🙏

Tuubz:
Please check out [this issue](https://github.com/nimblehq/compass/discussions/1019) too, we might end up using [this repository](https://github.com/nimblehq/publish-github-wiki-action) instead 🙏

manh-t:
@Tuubz yes, I'm aware of that issue. We might need to update this issue's description when the RFC is approved 🙏

hoangnguyen92dn:
@Tuubz @manh-t To conclude, which workflow we could use for now? `Updating the description when the RFC is approved` doesn't make sense imo.

Tuubz:
@manh-t 

If you agree, please update the description to make use of [this repository](https://github.com/nimblehq/publish-github-wiki-action) 🚀

PR #330, opened 2022-10-25
[#230] Update publish_docs_to_wiki workflow to use publish_wiki of github_action_workflow repository

manh-t:
Close https://github.com/nimblehq/android-templates/issues/230

## What happened 👀

The Nimble team has published a [publish-github-wiki-action](https://github.com/nimblehq/publish-github-wiki-action) to support our team NOT creating a publishing wiki workflow manually. We should use it as an advantage.

## Insight 📝

Now we only need to have a step to call [publish-github-wiki-action](https://github.com/nimblehq/publish-github-wiki-action)

## Proof Of Work 📹

image


.github/workflows/publish_docs_to_wiki.yml
```diff
@@ -7,35 +7,15 @@ on:
     branches:
```
Tuubz, line 7: @manh-t 

> This workflow will synchronize the .github/wiki folder with your project's Github Wiki.

If I understand correctly, we need to:
1.  Update `docs/**` with `.github/wiki/**` in `publish_docs_to_wiki.yml` 
2. Move everything inside `docs` to `.github/wiki` 
3. Remove `docs` folder

Reference: https://github.com/nimblehq/ios-templates/pull/299/files

What do you think? 🙏

manh-t: @Tuubz I'm aware of this, and I'm going to create an issue to move `docs` to `.github/wiki/**` instead of refactoring them in this PR. 🙆‍♂️

Tuubz: Oh, any reason we don't do it in this PR? Seems like an easy fix

If we merge this now, then we would be merging broken code 💭

manh-t: > If we merge this now, then we would be merging broken code

@Tuubz ah true that I forgot this. Okay I'll update it 😁

manh-t: @Tuubz updated 🙏

.github/workflows/publish_docs_to_wiki.yml
```diff
@@ -2,40 +2,18 @@ name: Publish docs to Wiki
 
 on:
   push:
-    paths:
-      - docs/**
```
Tuubz, line 6: @manh-t I could be wrong, but instead of removing this, should it be `- .github/wiki/**` instead?

Same as iOS has done: https://github.com/nimblehq/ios-templates/blob/master/.github/workflows/publish_docs_to_wiki.yml

Tuubz: @manh-t Without `paths:` the workflow will still work correctly, indeed 👍 

But from my understanding, keeping it will **optimize** our CI:
- If we remove `paths:`,  this workflow will be called every time we push to main
- If we keep `paths:`,  this workflow will only be called if something has been updated inside `.github/wiki/**`
Screenshot 2022-10-28 at 09 18 03

**Reference**: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-and-excluding-paths

manh-t: @Tuubz valid point 👍 Let me bring it back so we can limit the scope of running this workflow

manh-t: @Tuubz updated at 45d336e559aef785136f41efa8f0de84bc254dc9

Tuubz (approved):
🇻🇳 🏸 🥇

.github/workflows/publish_docs_to_wiki.yml
```diff
@@ -3,39 +3,19 @@ name: Publish docs to Wiki
 on:
   push:
     paths:
-      - docs/**
+      - .github/wiki/**
     branches:
       - main
```
luongvo, line 8: @manh-t I think we can manually update this branch to `chore/230-use-publish-github-wifi-action-in-publish-docs-to-wiki-workflow` and verify if it works properly on this branch, then it would be fine on the main branch 😄

luongvo (commented):
rest gltm

.github/workflows/publish_docs_to_wiki.yml
manh-t: @luongvo thank you 🤗 I will give it a try to create a PoW for this PR (it's better to know whether the workflow works correctly or not)

.github/workflows/publish_docs_to_wiki.yml
```diff
@@ -2,40 +2,18 @@ name: Publish docs to Wiki
...
-
-      - name: Push content to Wiki
-        run: |
-          rsync -av --delete docs/ tmp_wiki/ --exclude .git
-          cd tmp_wiki
-          git add .
-          git commit -m "Update Wiki content"
-          git push -f --set-upstream https://$USER_TOKEN@github.com/$OWNER/$REPOSITORY_NAME.wiki.git master
+      - name: Publish Github Wiki
+        uses: nimblehq/publish-github-wiki-action@v0.1.0-alpha-3
```
luongvo, line 15: should it be `nimblehq/github-actions-workflows/.github/workflows/publish_wiki.yml@0.1.0`?

luongvo: oh `nimblehq/publish-github-wiki-action@v0.1.0-alpha-3` works 👍

luongvo (approved):
LGTM

Tuubz:
@nimblehq/android-chapter Can anyone give one more review/approval? 💯

chornerman (approved):

Tuubz:
@manh-t We can't merge this pull request yet unfortunately 😢 

@bterone suggests we wait until `v1.0` (he's working on it now), because of [this unresolved PR](https://github.com/nimblehq/publish-github-wiki-action/pull/9)

manh-t:
> @manh-t We can't merge this pull request yet unfortunately 😢
> 
> @bterone suggests we wait until `v1.0` (he's working on it now), because of [this unresolved PR](https://github.com/nimblehq/publish-github-wiki-action/pull/9)

@Tuubz I see 😭 so I will take the next train then.

Issue #332, opened 2022-10-27
Rename the current template to XML


Tuubz:
## Why

The template currently consists of both `XML` & `Compose`.

For the sake of simplicity, we will have a separate template for XML and another one for Compose. 

We rename the current template, so we can still keep track on its git history

## Acceptance Criteria

- Rename `template` to `template-xml`
- Change package name to `co.nimblehq.template.xml`
- Update app name to `Template XML`
- Update the workflows for `template-xml`
- Update `new_project.kts` so it still works correctly

**Note**: Removing the code of Compose will be done in a different issue (#334)

luongvo:
> Create new folder template-xml copied from template
> Change package name to co.nimblehq.template.xml and adjust (AndroidManifest.xml, applicationId, ...)

@Tuubz @Wadeewee  I'm wondering if we follow this way, we will lose all git history from what we contributed into the current `template` folder. Should we rename `template` to the new `template-xml` instead? This way does not actually destroy the history & not require the next task #334 to remove what we just cloned (Compose).

Wadeewee:
Good point @luongvo 👍 So, If we have a plan to remove `template` and use `template-xml` instead I think renaming way it  could be better than copy and remove the previous one then we can keep git history 🤔

Tuubz:
@luongvo @Wadeewee I have updated the issue 🚀 

However, I still believe we should do it in two pull requests (clone first, then remove)

IMO, if we do it in two pull requests it will be clear to see what files of compose are being removed from template-compose.

(Ex: Base classes, dependencies, other gradle configurations, ...)

What do you think? 🙏

luongvo:
@Tuubz lgtm +1

Issue #334, opened 2022-10-27
Remove Compose code from XML template


Tuubz:
## Why

Now we have a separate template for both XML & Compose.

Remove and/or update the code accordingly, so there's no code regarding Compose left.

## Acceptance Criteria

Remove all code regarding Compose and update accordingly

Issue #333, opened 2022-10-27
Create a new template for Compose


Tuubz:
## Why

The template currently consists of both `XML` & `Compose`.

For the sake of simplicity, we will have a separate template for XML and another one for Compose. 

## Acceptance Criteria

- Create new folder `template-compose` copied from `template`
- Change package name to `co.nimblehq.template.compose`
- Update app name to `Template Compose`
- Update the workflows for `template-compose`

**Note**: Removing the code of XML will be done in a different issue (#335)

luongvo:
@Tuubz @chornerman I think we should do #334 in this task also. Clone the `template` into the new `template-compose` and remove unnecessary things from the XML method in just 1 PR. I don't think we should introduce new file added and then remove them in a separate PR.

Tuubz:
@chornerman @luongvo

 Let's not do #333 & #334 together in one pull request, otherwise, it will be difficult to see what changes have been made 🙏 
 
IMO, if we do it in two pull requests it will be clear to see what files of XML are being removed from `template-compose`.

What do you think? 🙏

luongvo:
@Tuubz lgtm 👍

Wadeewee:
@Tuubz, I think you need to update the `AC` a bit of the first bullet point

> - Create new folder template-compose copied from template

Tuubz:
@Wadeewee Good morning 👋 

What update do you suggest I should make? 🙏

Wadeewee:
> What update do you suggest I should make? 🙏

@Tuubz To work on this, Should we branch out from `chore/332-rename-the-current-template-to-xml` because we don't have `template` once we implemented [#332](https://github.com/nimblehq/android-templates/issues/332) 🤔

Wadeewee:
Oh, wait, question @Tuubz Why don't we rename `sample-compose` instead of creating a new one for `template-compose` 🤔

Tuubz:
@Wadeewee @chornerman  The idea originally was to do #333 first, but I'm fine to do it after #332 too 👍 

> Why don't we rename sample-compose instead of creating a new one for template-compose 🤔

Hmmm, `sample-compose` is for sample purposes, `template-compose` is used to generate a new project.

Please let me know if I misunderstood you question 🙇

PR #349, opened 2022-11-17
[#332] Rename the current template to XML

Wadeewee:
Closes https://github.com/nimblehq/android-templates/issues/332

## What happened 👀

- Rename template to `template-xml`
- Change package name to `co.nimblehq.template.xml`
- Update app name to `Template XML`
- Update the workflows for `template-xml`
- Update `new_project.kts` so it still works correctly
Note: Removing the code of Compose will be done in a different issue (https://github.com/nimblehq/android-templates/issues/334)

## Insight 📝

We need to update the path name to support `Kover` in `Dangerfile`file as well.

## Proof Of Work 📹

https://user-images.githubusercontent.com/28002315/202333731-78889179-627f-4a0d-9322-a0a18e404028.mov


Wadeewee:
@Tuubz, Should we update `README.md` on the root `android-template` at line `12` 🤔

Tuubz:
@Wadeewee Yes, sounds good 👍

.github/workflows/review_pull_request.yml
```diff
@@ -30,16 +30,16 @@ jobs:
...
-      - name: Run Android Lint on template
-        working-directory: ./template
+      - name: Run Android Lint on template-xml
+        working-directory: ./template-xml
         run: ./gradlew lint
 
-      - name: Run unit tests with Kover on template
-        working-directory: ./template
+      - name: Run unit tests with Kover on template-xml
+        working-directory: ./template-xml
```
lydiasama, line 42: @Wadeewee Wait, This workflow is in the big `android-template`, isn't it? 
If we update on this level, I am concerned that this workflow will not work with the `template-compose`. We probably need to separate workflow files for `template-xml` and `template-compose`. What do you think? cc @Tuubz 

Screen Shot 2565-11-17 at 10 14 47 AM

.github/workflows/run_detekt_and_unit_tests.yml
```diff
@@ -38,8 +38,8 @@ jobs:
           name: DetektReportsTemplate
           path: template/build/reports/detekt/
 
-      - name: Run unit tests with Kover on template
-        working-directory: ./template
+      - name: Run unit tests with Kover on template-xml
+        working-directory: ./template-xml
```
lydiasama, line 42: Ditto here.

Dangerfile
```diff
@@ -32,6 +32,6 @@ end
 # Show Danger test coverage report from Kover for template
 # Report coverage of modified files, warn if total project coverage is under 80%
 # or if any modified file's coverage is under 90%
-kover_file = "template/build/reports/kover/merged/xml/report.xml"
+kover_file = "template-xml/build/reports/kover/merged/xml/report.xml"
 markdown "## Kover report:"
-shroud.reportKover "Template Unit Tests", kover_file, 80, 95, false
+shroud.reportKover "Template - XML Unit Tests", kover_file, 80, 95, false
```
lydiasama, line 37: The same question as the workflow file. How about the `template-compose` in the future? Do we have the Kover for it too? If we change the path here, what will happen when we would like to do it for `template-compose`? 🤔

lydiasama (commented):
I have concerns about the top level shared file between the `template-xml` and `template-compose` 🙏🏻

.github/workflows/review_pull_request.yml
Tuubz: The step for `template-compose` will be run later, once @chornerman integrates #333

lydiasama: @Tuubz Hmmm 😅  Then finally the workflow will have only for `template-compose`, right? 😂 Imagine when someone updates the `template-xml`, the workflow will not run for it. 🤔  Is that your intention?

Tuubz: @lydiasama No, the workflow will have both `template-compose` & `template-xml` 

A new step will be added for `template-compose` in #333

Did I miss something? 🙇

lydiasama: @Tuubz Then I think it's time to rename this file to `xml_review_pull_request` and for #333 will not be update but create a new workflow `compose_review_pull_request`. Same to the rest workflows. What do you think?

Tuubz: Hmmm, any reason we can't keep it in one file? We used to do the same for `RxJava` & `Coroutine` template  🙏

lydiasama: @Tuubz Oh, I might be wrong. If we can create it in one file, it would be best. 👍🏻

luongvo:
@Tuubz @Wadeewee As discussed, we should base this PR on #333? Do you aware of this ticket's blocking?

image

scripts/new_project.kts
```diff
@@ -21,7 +21,7 @@ object NewProject {
 
     private const val TEMPLATE_APP_NAME = "Template"
     private const val TEMPLATE_APPLICATION_CLASS_NAME = "TemplateApplication"
-    private const val TEMPLATE_FOLDER_NAME = "template"
+    private const val TEMPLATE_FOLDER_NAME = "template-xml"
     private const val TEMPLATE_PACKAGE_NAME = "co.nimblehq.template"
```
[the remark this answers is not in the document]
lydiasama: Why don't we update the `TEMPLATE_APP_NAME` to `Template-xml` too? 🤔

template-xml/domain/src/main/java/co/nimblehq/template/xml/domain/repository/Repository.kt
```diff
@@ -1,6 +1,6 @@
-package co.nimblehq.template.domain.repository
+package co.nimblehq.template.xml.domain.repository
```
luongvo, line 1: I don't think we need to move the `data` and `domain` module to the new package `co.nimblehq.template.xml`. I think  `co.nimblehq.template` is enough. What do you think @Wadeewee @Tuubz?

luongvo:
@Wadeewee I think what you listed in the Insight section is already for What happened, so please don't try to repeat them to have something in the Insight section 🙏 

image

template-xml/domain/src/main/java/co/nimblehq/template/xml/domain/repository/Repository.kt
Tuubz: If the package for `data` & `domain` are not aligned with `app`, then we will have to make changes to the script, no? 💭

Tuubz:
> @Tuubz @Wadeewee As discussed, we should base this PR on #333? Do you aware of this ticket's blocking?
> 
> image

I think we can still proceed with this approach, however, @Wadeewee will have to wait until @chornerman has started with #333, before she can start on #334  🙏

Wadeewee:
> > @Tuubz @Wadeewee As discussed, we should base this PR on #333? Do you aware of this ticket's blocking?
> > image
> 
> I think we can still proceed with this approach, however, @Wadeewee will have to wait until @chornerman has started with #333, before she can start on #334 🙏

Oh, @Tuubz @luongvo 🙈  I just noticed it was blocked by [#333](uhttps://github.com/nimblehq/android-templates/issues/333rl) but why don't we sort the `issue` to avoid confusion of number 🙏

lydiasama:
>why don't we sort the issue to avoid confusion of number 🙏

@Wadeewee I don't think we can sort it at the project board. 😄  The order is generated automatically once we create the issue. And I don't think we could have a concrete plan when creating it. 😂

lydiasama:
@Tuubz Maybe moving the `blocking` column to be closer by the title could help?

Wadeewee:
> > why don't we sort the issue to avoid confusion of number 🙏
> 
> @Wadeewee I don't think we can sort it at the project board. 😄 The order is generated automatically once we create the issue. And I don't think we could have a concrete plan when creating it. 😂

@lydiasama Actually, I mean the `#333` should be `#332` 🙏

Dangerfile
Wadeewee: @lydiasama We have used the `template` as the default before but once we rename it, we need to rename the path for this as well `template` -> `template-xml`; if not, the `CI` will be `failed for this time`, I guess once we already implemented [#333](https://github.com/nimblehq/android-templates/issues/333) we might have some task to filter template between `template-compose` and `template-xml` before we start work on [#332](https://github.com/nimblehq/android-templates/issues/332) If you have a suggestion, please let me know cc @Tuubz 🙇‍♀️

template-xml/domain/src/main/java/co/nimblehq/template/xml/domain/repository/Repository.kt
Wadeewee: > I don't think we need to move the data and domain module to the new package co.nimblehq.template.xml

@luongvo Shouldn't we name the package to make it consistent with the `module's name`

Dangerfile
Tuubz: @lydiasama Same as the other workflows, we'll add one more step for `template-compose` in #333  🙏

Wadeewee: FYI: I will leave `TODO` to avoid confusion at this time 🚀

template-xml/domain/src/main/java/co/nimblehq/template/xml/domain/repository/Repository.kt
luongvo: @Wadeewee @Tuubz Alright, makes sense cuz we will have a different package for compose template too 🤔

.github/workflows/run_detekt_and_unit_tests.yml
```diff
@@ -38,8 +39,8 @@ jobs:
           name: DetektReportsTemplate
           path: template/build/reports/detekt/
```
luongvo, line 40: should we update the path for detekt and `Archive code coverage reports on template` step too?

template-xml/data/src/main/java/co/nimblehq/template/xml/storage/NormalSharedPreferences.kt
```diff
@@ -1,4 +1,4 @@
-package co.nimblehq.template.data.storage
```
luongvo, line 1: Why do we remove `data` in data package?

template-xml/domain/src/main/java/co/nimblehq/template/xml/domain/exceptions/Exceptions.kt
```diff
@@ -1,6 +1,6 @@
-package co.nimblehq.template.domain.exceptions
```
luongvo, line 1: same for `domain` package 🤔

luongvo (commented):
rest lgtm

template-xml/data/src/main/java/co/nimblehq/template/xml/storage/NormalSharedPreferences.kt
Wadeewee: @luongvo Copied wrong 🙈 Thank to spotting this 🙏

template-xml/domain/src/main/java/co/nimblehq/template/xml/domain/exceptions/Exceptions.kt
Wadeewee: @luongvo We still have `domain` here ☝️ `co.nimblehq.template.xml.domain`

.github/workflows/run_detekt_and_unit_tests.yml
```diff
@@ -28,18 +28,19 @@ jobs:
...
-          path: template/build/reports/detekt/
+          path: template-xml/build/reports/detekt/
 
-      - name: Run unit tests with Kover on template
-        working-directory: ./template
+      - name: Run unit tests with Kover on template-xml
+        working-directory: ./template-xml
         run: ./gradlew koverMergedHtmlReport
 
       - name: Archive code coverage reports on template
```
chornerman, line 46: I think we might have to update some more names as discussed 🙏🏼 
Screenshot 2565-11-18 at 11 48 23

chornerman (approved):
LGTM 🚀

AVI5HEK (approved):

Tuubz (commented):
@Wadeewee We missed one small thing regarding the code coverage 🦅 

In `build.gradle.kts` for `koverMerged`, we forgot to update the excluded packages:

```
 val excludedPackages = setOf(
      "com.bumptech.glide.*",
      "dagger.hilt.internal.*",
      "hilt_aggregated_deps.*",
      "co.nimblehq.template.databinding.*",
      "co.nimblehq.template.di.*"
  )
```

The code coverage should be `20.24%` instead of `18.07%`  🚀

doannimble (approved):
LGTM 🚀

PR #350, opened 2022-11-18
[#333] Create new template for Compose

chornerman:
https://github.com/nimblehq/android-templates/issues/333

## What happened 👀

Create a new template for Compose copied from a template for XML

## Insight 📝

- Create a new folder `template-compose` copied from `template-xml`
- Change package name to `co.nimblehq.template.compose`
- Update application class name to `TemplateComposeApplication`
- Update app name to `Template Compose`
- Update the workflows for `template-compose`
- Add `build/` path to `.gitignore` to prevent generated `build` folder from being pushed to git (apply for both templates and samples)
- Add a step to show the Danger test coverage report from Kover for `template-compose`
- Update `README` file


Tuubz:
@chornerman Good morning, can we do a rebase? 🌻

chornerman:
> @chornerman Good morning, can we do a rebase? 🌻

Done 🚀

.gitignore
```diff
@@ -1,5 +1,6 @@
 *.iml
 .idea
 .DS_Store
+build/
```
luongvo, line 4: There is no build folder at the root template. Not sure why do we need this?

image

luongvo (commented):
rest lgtm

sample-compose/build.gradle.kts
```diff
@@ -102,8 +102,8 @@ koverMerged {
         "com.bumptech.glide.*",
         "dagger.hilt.internal.*",
         "hilt_aggregated_deps.*",
-        "co.nimblehq.template.databinding.*",
-        "co.nimblehq.template.di.*"
+        "co.nimblehq.sample.compose.databinding.*",
+        "co.nimblehq.sample.compose.di.*"
```
Wadeewee, line 106:
```suggestion
        "co.nimblehq.template.compose.databinding.*",
        "co.nimblehq.template.compose.di.*"
```

Wadeewee: Oh, wait, this one is in [sample-compose/build.gradle.kts](https://github.com/nimblehq/android-templates/pull/350/files/99a718f36539580628e76d62a10f31147ced2e71#diff-d38b787b9906da28c79ae1c08c59ed7879dc1842807cb7ec8138cabb22c0a2a3) It seems like it's correctly but Do we need to update this 🤔  @Tuubz As we will have `template-xml/template-compose` So, Do we still need `sample-compose`

Tuubz: > Do we need to update this

The update is correct (although it is out of scope, that's OK)

> Do we still need sample-compose

Yes, `sample-compose` is for sample purposes, `template-compose` is used to generate a new project 🚀

template-compose/app/build.gradle.kts
```diff
@@ -0,0 +1,174 @@
...
+
+dependencies {
+    implementation(project(Module.DATA))
+    implementation(project(Module.DOMAIN))
+
+    implementation(fileTree(mapOf("dir" to "libs", "include" to listOf("*.jar"))))
+
+    implementation("androidx.activity:activity-ktx:${Versions.ANDROIDX_ACTIVITY_KTX_VERSION}")
+    implementation("androidx.appcompat:appcompat:${Versions.ANDROIDX_SUPPORT_VERSION}")
+    implementation("androidx.constraintlayout:constraintlayout:${Versions.CONSTRAINT_LAYOUT_VERSION}")
```
Wadeewee, line 127: We can remove these `libs` for `template-compose` as we don't require it anymore 🙏

Tuubz: Code removal should be done in #334 🙏

chornerman: @Wadeewee Done in #351 🙏🏼

sample-compose/build.gradle.kts
Wadeewee: > Yes, sample-compose is for sample purposes, template-compose is used to generate a new project 🚀

@Tuubz May I know why don't we add the example  of `compose` to `template-compose` instead of `sample-compose` So when developer generates the `template-compose`, then they will see the sample `compose`🤔

.gitignore
chornerman: @luongvo without `build/` the below auto-generated build folder will be added to git (this folder is generated every time we run the project)

Screenshot 2565-11-21 at 15 18 38

sample-compose/build.gradle.kts
Tuubz: If we keep the sample code in `template-compose`, then the developer will have to remove all the sample code manually. This is an annoying process, so the purpose of `template-compose` is to start a new project as clean and fast as possible 👍 

Please let me know if I misunderstood your question 🐱

Wadeewee: > This is an annoying process, so the purpose of template-compose is to start a new project as clean and as fast as possible

@Tuubz What about the `classes` in module `data` and `domain` e.g., `ApiService` , including `unitTest` inside there 🤔

Tuubz: The idea of `template` is to provide the minimum amount of code required to get started (quickly), but also highlight the project structure 🚀 

If we remove `ApiService.kt`, then the developer will need to add the setup manually (Ex: `ApiServiceProvider.kt`). It will require more time to add this ourselves, instead of simply renaming it.

For `UseCaseTest.kt`, I think it's acceptable to keep some basic test classes (think about how we generate a new project from Android Studio, they also include a few test classes by default)

Anyways, if you still feel like they should be removed, let's move this discussion to an RFC ✍️

.gitignore
luongvo: @chornerman Do we have `buildSrc` folder at the root template? `buildSrc` is included in each `template-xml` or `template-compose` folder, not at the root level. So why do we need to ignore `build/` at root?

luongvo: @chornerman This gitignore file will actually take care of all the files in `template-compose` folder. 🤓 

image

chornerman: Good point 😄 Let me update

sample-compose/.gitignore
```diff
@@ -14,6 +14,7 @@
 bin/
 gen/
 out/
+build/
```
luongvo, line 17: We already ignored the `build` folder in each template and sample project. This is duplicated action 🙏 

image

image

luongvo: @chornerman In this case, I think we should add the missing gitignore file for each `buildSrc` module instead. We should not put all git ignore config at the root file.

image

chornerman: I see, let me update 🚀

luongvo: @chornerman Yes, please, also for the old Rx template project

image

chornerman: @luongvo Done 🚀

luongvo (approved):
lgtm

Wadeewee (approved):
LGTM 🚀

Tuubz (approved):
💪 💪 💪

PR #139, opened 2021-10-29
[#83] Setup the unit testing part for Coroutine Template

manh-t:
https://github.com/nimblehq/android-templates/issues/83

## What happened 👀

We haven't written any unit testing part for the presentation layer. This PR is to set up all necessary pieces so we can easily write the unit test for Fragment/ Activity later.

## Insight 📝

Continue the work for CoroutineTemplate from #137 

## Proof Of Work 📹

https://user-images.githubusercontent.com/60863885/139396980-7a5fa194-e239-4c42-bc1c-18e67f56dadb.mov

PR #355, opened 2022-11-23
[#334] Remove Compose code from template-xml

Wadeewee:
https://github.com/nimblehq/android-templates/issues/334

## What happened 👀

Now we have a separate template for both XML & Compose.

Remove and/or update the code accordingly, so there's no code regarding Compose left.

## Insight 📝

- Remove the config option and some libs of `Compose` in `build.gradle.kts.`
- Remove `BaseComposeFragment` and `BaseComposeFragmentCallbacks` inside the `base` directory.
- Remove directory `compose` inside directory `screen.`
- Rename directory `xml` to `home` inside the directory screen.`

Please note: We have removed some fragment extensions in `ViewModelExt` as it was unused for this time. We can consider adding them later if needed.

## Proof Of Work 📹

Screenshot_20221123_100716


template-xml/app/src/main/java/co/nimblehq/template/xml/extension/ViewModelExt.kt
```diff
@@ -16,25 +14,13 @@ import androidx.lifecycle.viewmodel.CreationExtras
...
     noinline extrasProducer: (() -> CreationExtras)? = null,
     noinline factoryProducer: (() -> ViewModelProvider.Factory)? = null
 ): Lazy<VM> = OverridableLazy(viewModels(ownerProducer, extrasProducer, factoryProducer))
 
-@MainThread
-inline fun <reified VM : ViewModel> ComponentActivity.provideViewModels(
-    noinline extrasProducer: (() -> CreationExtras)? = null,
-    noinline factoryProducer: (() -> ViewModelProvider.Factory)? = null
-): Lazy<VM> = OverridableLazy(viewModels(extrasProducer, factoryProducer))
-
```
Tuubz, line 37: @manh-t What do you think of these changes? 

Just want to make sure these can be removed, as they are currently not being used 🙏 

Related to #137 & #139

Tuubz (approved):

Tuubz:
Test

template-xml/app/src/main/java/co/nimblehq/template/xml/extension/ViewModelExt.kt
Tuubz: Weird, this comment didn't trigger in the slack channel cc @manh-t

template-xml/app/src/test/java/co/nimblehq/template/xml/ui/screens/xml/HomeFragmentTest.kt
```diff
@@ -6,6 +6,8 @@ import co.nimblehq.template.xml.test.TestNavigatorModule.mockMainNavigator
 import co.nimblehq.template.xml.test.getPrivateProperty
 import co.nimblehq.template.xml.test.replace
 import co.nimblehq.template.xml.ui.BaseFragmentTest
+import co.nimblehq.template.xml.ui.screens.home.HomeFragment
+import co.nimblehq.template.xml.ui.screens.home.HomeViewModel
```
luongvo, line 10: Please move this `HomeFragmentTest` file to `package co.nimblehq.template.xml.ui.screens.home`

image

template-xml/app/src/main/java/co/nimblehq/template/xml/extension/ViewModelExt.kt
manh-t: @Tuubz @Wadeewee they are the handy functions for unit testing. We didn't have the test for Activity in our template, so these functions are useless. We can remove them IMO (we have a link above for reference if needed ) 🙏

luongvo: @Tuubz @manh-t I do think we should keep them as helper extensions except until we have another approach to do testing with Fragment and Activity. Proguard will clean unused code when doing code obfuscation. 

That's the same as having [common extensions lib](https://github.com/nimblehq/android-common-ktx/tree/develop/common-ktx/src/main/java/co/nimblehq/common/extensions) with no usage yet.

luongvo (commented):
rest lgtm

template-xml/app/src/main/java/co/nimblehq/template/xml/extension/ViewModelExt.kt
Wadeewee: @luongvo @Tuubz To make it clean when generating the script for a new project, I don't think we need to keep the unused function. We can add it later in the future if needed, Also @manh-t said 

> We didn't have the test for Activity in our template, so these functions are useless. We can remove them IMO (we have a link above for reference if needed ) 

So, I prefer to remove it for now 🙏

luongvo (approved):

chornerman (approved):
LGTM 🚀

Issue #303, opened 2022-09-09
Update Detekt rules for Jetpack Compose


luongvo:
## Why

We're still using the legacy Detekt config for Jetpack Compose project, that's why Detekt reports a lot of issues related to new [Jetpack Compose conventions](https://nimblehq.co/compass/development/android/user-interface/jetpack-compose/#conventions). Hence we have to skip them by adding a lot of [`Suppress` annotations as a workaround](https://github.com/nimblehq/jetpack-compose-crypto/blob/develop/app/src/main/java/co/nimblehq/compose/crypto/ui/screens/home/HomeScreen.kt#L34).

There is a [new Detekt rules for Compose](https://detekt.dev/docs/introduction/compose/) should be applied to get rid of those issues.

## Who Benefits?

Developers

lydiasama:
@luongvo Currently, when we create a new project from the script, it generate both `xml` and `compose` project. Does it mean we will not support detekt for the `xml` project after generating it? 🤔 
Also @chornerman  has updated the detekt rules for `sample-compose` project already [here](https://github.com/nimblehq/android-templates/pull/274). So if the consumer could check it in that sample repo.

BTW, do we need to vote on this kind of issue first? @ryan-conway

ryan-conway:
@lydiasama Yes we should vote 😆 Preferably on the [discussions page](https://github.com/nimblehq/android-templates/discussions), but lets migrate slowly 😉 

@luongvo It's true that our Detekt config needs to be updated for Jetpack Compose, but we need to make sure that the new rules we apply don't negatively affect the developer when they choose to use XML instead.
Take function naming for example. We could update the pattern to `[a-zA-Z][a-zA-Z0-9]*` from `[a-z][a-zA-Z0-9]*`, but this means that _all_ functions can be named in Pascal case which is probably not our desired outcome 😉 

Personally I think we should wait for #214 to be implemented and have separate Detekt configs for XML and Compose 🙏🏻

luongvo:
> Also @chornerman has updated the detekt rules for sample-compose project already https://github.com/nimblehq/android-templates/pull/274. So, the consumer could check it in that sample repo.

@lydiasama Yes I see that, however, we're not having full config from https://detekt.dev/docs/introduction/compose. That needs to be updated more 🚀

luongvo:
@ryan-conway Yes, actually we should have different config between legacy XML and Compose projects.

PR #362, opened 2022-12-08
[#303] Update detekt ruels for Jetpack Compose

Wadeewee:
https://github.com/nimblehq/android-templates/issues/303

## What happened 👀

We're still using the legacy Detekt config for Jetpack Compose project, that's why Detekt reports a lot of issues related to new [Jetpack Compose conventions](https://nimblehq.co/compass/development/android/user-interface/jetpack-compose/#conventions). Hence we have to skip them by adding a lot of [Suppress annotations as a workaround](https://github.com/nimblehq/jetpack-compose-crypto/blob/develop/app/src/main/java/co/nimblehq/compose/crypto/ui/screens/home/HomeScreen.kt#L34).

There is a [new Detekt rules for Compose](https://detekt.dev/docs/introduction/compose/) should be applied to get rid of those issues.

## Insight 📝

- Set `ignoreAnnotated: ['Composable']` in `tempate-compose`.
- Set `propertyPattern: '[A-Za-z][_A-Za-z0-9]*'` in both `tempate-compose` and `sample-compose`.
- Set ` ignorePropertyDeclaration: true`.
- Remove `@Suppress` for `MagicNumber` and  `FunctionNaming`.

## Proof Of Work 📹

No warning of Jetpack Compose from Detekt.


sample-compose/detekt-config.yml
```diff
@@ -189,7 +189,7 @@ naming:
   FunctionNaming:
     active: true
     functionPattern: '^([a-z$][a-zA-Z$0-9]*)|(`.*`)$'
-    ignoreAnnotated: [ 'Composable' ]
+    ignoreAnnotated: ['Composable']
```
luongvo, line 192: Wrong formatting?

```suggestion
    ignoreAnnotated: [ 'Composable' ]
```

sample-compose/detekt-config.yml
```diff
@@ -204,7 +204,7 @@ naming:
   TopLevelPropertyNaming:
     active: true
     constantPattern: '[A-Z][_A-Z0-9]*'
-    propertyPattern: '[A-Z][A-Za-z0-9]*'
+    propertyPattern: '[A-Za-z][_A-Za-z0-9]*'
```
luongvo, line 207: Saying the `HomeScreen` has a Compose constant style, this rule still reports issue

image

image

My suggestion is

```
    constantPattern: '[A-Z][_A-Za-z0-9]*'
    propertyPattern: '[A-Za-z][A-Za-z\d]*'
```

sample-compose/detekt-config.yml
```diff
@@ -189,7 +189,7 @@ naming:
   FunctionNaming:
     active: true
     functionPattern: '^([a-z$][a-zA-Z$0-9]*)|(`.*`)$'
```
luongvo, line 191: The migration doc says we should use this rule instead.

```suggestion
    functionPattern: '[a-zA-Z][a-zA-Z0-9]*'
```

luongvo (commented):
There are some deprecated Detekt attrs, please consider migrating them as well 🙏 

```
> Task :detekt
Property 'complexity>LongParameterList>threshold' is deprecated. Use `functionThreshold` and `constructorThreshold` instead.
Property 'exceptions>ExceptionRaisedInUnexpectedLocation>methodNames' should be a YAML array instead of a comma-separated String.
Property 'exceptions>ThrowingExceptionsWithoutMessageOrCause>exceptions' should be a YAML array instead of a comma-separated String.
Property 'naming>ForbiddenClassName>forbiddenName' should be a YAML array instead of a comma-separated String.
Property 'naming>MemberNameEqualsClassName>ignoreOverriddenFunction' is deprecated. Use `ignoreOverridden` instead.
Property 'potential-bugs>LateinitUsage>excludeAnnotatedProperties' is deprecated. Use `ignoreAnnotated` instead.
Property 'style>ForbiddenComment>values' should be a YAML array instead of a comma-separated String.
Property 'style>ForbiddenImport>imports' should be a YAML array instead of a comma-separated String.
Property 'style>MagicNumber>ignoreNumbers' should be a YAML array instead of a comma-separated String.
Property 'style>UseDataClass>excludeAnnotatedClasses' is deprecated. Use `ignoreAnnotated` instead.
```

image

sample-compose/detekt-config.yml
Wadeewee: Yeah, I agree @luongvo  but If we update this, I think we should update for `template-xml` and `sample-xml` as well 🚀

Wadeewee:
> There are some deprecated Detekt attrs, please consider migrating them as well 🙏
> 
> ```
> > Task :detekt
> Property 'complexity>LongParameterList>threshold' is deprecated. Use `functionThreshold` and `constructorThreshold` instead.
> Property 'exceptions>ExceptionRaisedInUnexpectedLocation>methodNames' should be a YAML array instead of a comma-separated String.
> Property 'exceptions>ThrowingExceptionsWithoutMessageOrCause>exceptions' should be a YAML array instead of a comma-separated String.
> Property 'naming>ForbiddenClassName>forbiddenName' should be a YAML array instead of a comma-separated String.
> Property 'naming>MemberNameEqualsClassName>ignoreOverriddenFunction' is deprecated. Use `ignoreOverridden` instead.
> Property 'potential-bugs>LateinitUsage>excludeAnnotatedProperties' is deprecated. Use `ignoreAnnotated` instead.
> Property 'style>ForbiddenComment>values' should be a YAML array instead of a comma-separated String.
> Property 'style>ForbiddenImport>imports' should be a YAML array instead of a comma-separated String.
> Property 'style>MagicNumber>ignoreNumbers' should be a YAML array instead of a comma-separated String.
> Property 'style>UseDataClass>excludeAnnotatedClasses' is deprecated. Use `ignoreAnnotated` instead.
> ```
> 
> image

Good point, @luongvo 👍 but I think it's better to create a new issue to fix because it was deprecated in all projects in `android-template`. What do you think? cc @Tuubz

luongvo:
> Good point, @luongvo 👍 but I think it's better to create a new issue to fix because it was deprecated in all projects in `android-template`. What do you think? cc @Tuubz If you guys ok then I will create an issue and include it in current sprint 🙏

lgtm, separate PR makes sense 👍

sample-compose/detekt-config.yml
luongvo: @Wadeewee I'm not sure. I think we should focus on updating Detekt rules for Compose in this PR only 🤔

sample-compose/detekt-config.yml
Wadeewee: Oh, sorry  @luongvo I think you're right about `constant style.` I just found from [API Guidelines](https://github.com/androidx/androidx/blob/androidx-main/compose/docs/compose-api-guidelines.md#api-guidelines-for-jetpack-compose)
Screen Shot 2565-12-09 at 11 48 10

 I think we should add more part about `Singletons, constants, sealed class, and enum class values` for the convention(`do/don't`) in compass as well. What do you think? cc @hoangnguyen92dn

template-compose/detekt-config.yml
```diff
@@ -202,8 +203,8 @@ naming:
     packagePattern: '^[a-z]+(\.[a-z][a-z0-9]*)*$'
   TopLevelPropertyNaming:
     active: true
-    constantPattern: '[A-Z][_A-Z0-9]*'
-    propertyPattern: '[a-z][A-Za-z\d]*'
+    constantPattern: '[A-Z][_A-Za-z0-9]*'
```
lydiasama, line 206: @Wadeewee Why don't we use this instead?
```suggestion
    constantPattern: '[A-Z][A-Za-z0-9]*'
```
Ref: https://detekt.dev/docs/introduction/compose/#toplevelpropertynaming

template-compose/detekt-config.yml
```diff
@@ -202,8 +203,8 @@ naming:
...
+    propertyPattern: '[A-Za-z][_A-Za-z0-9]*'
```
lydiasama, line 207: Why do we need to update this? Could you please suggest me as I didn't see it was suggested in the [document](https://detekt.dev/docs/introduction/compose/#toplevelpropertynaming). 🙏🏻

lydiasama (commented):
Let's add `UnusedPrivateMember` to follow the [document](https://detekt.dev/docs/introduction/compose/#unusedprivatemember) as well.

```
UnusedPrivateMember:
    active: true
    allowedNames: '(_|ignored|expected|serialVersionUID)'
    ignoreAnnotated: ['Preview']
```

template-compose/detekt-config.yml
Wadeewee: @lydiasama We need support in case It doesn't compose `constantPattern`. For example in `OkHttpClientModule.kt`:

```
private const val READ_TIME_OUT = 30L
```

That's why we need to use `[A-Z][_A-Za-z0-9]*` instead of `[A-Z][A-Za-z0-9]*`.

template-compose/detekt-config.yml
Wadeewee: @lydiasama As we have discussed on this [Thread](https://nimble-co.slack.com/archives/G41UCTGHL/p1668056608803069), We don't wrap compose for `Color.kt` or `Dimension.kt` in `Object` class. That means when we remove `Object`, then we will get the warning from `Detekt`:

Screen Shot 2565-12-12 at 17 35 51
Screen Shot 2565-12-12 at 17 36 16


Regarding this, we need to update the `pattern` for `propertyPattern`. For more [ref](https://detekt.dev/docs/rules/naming/#toplevelpropertynaming) 🙏

template-compose/detekt-config.yml
lydiasama: @Wadeewee Hmm, it's kind of weird. Since we used the recommended regex in DeeMoney Migration, and it works for all classes. 🤔  @Tuubz @ryan-conway @kaungkhantsoe Do you have any idea?

lydiasama: Oh, I was wrong. DeeMoney used this pattern too. 👍🏻

template-compose/detekt-config.yml
lydiasama: @Wadeewee BTW, we don't declare the `snake_case` for the properties, right? Shall we get rid of `_` in pattern?
```suggestion
    propertyPattern: '[A-Za-z][A-Za-z0-9]*'
```

kaungkhantsoe:
> > Let's add `UnusedPrivateMember` to follow the [document](https://detekt.dev/docs/introduction/compose/#unusedprivatemember) as well.
> > ```
> > UnusedPrivateMember:
> >     active: true
> >     allowedNames: '(_|ignored|expected|serialVersionUID)'
> >     ignoreAnnotated: ['Preview']
> > ```
> 
> @lydiasama For `ignoreAnnotated: ['Preview']`, I can't reproduce it. I don't get any warning from `Detekt`. That's why I think we don't need to use it, but if you can reproduce this case, please let me know🙏 I'm not sure if It's because the doc is not up to date or not 🤔

@Wadeewee I think it is because we no longer have `UnusedPrivateMember` in `detekt-config`. Normally when we define a private variable that is unused, we got the `detekt error`. Now we don't get the error anymore. Did you intentionally removed the rule? cc @lydiasama

template-compose/detekt-config.yml
Wadeewee: Oh yeah 😅  your right, @lydiasama 👍 I don't beware of `_` when I copied from the [doc](https://detekt.dev/docs/rules/naming/#toplevelpropertynaming)  Thank to spotting this 🙏

Wadeewee:
> > > Let's add `UnusedPrivateMember` to follow the [document](https://detekt.dev/docs/introduction/compose/#unusedprivatemember) as well.
> > > ```
> > > UnusedPrivateMember:
> > >     active: true
> > >     allowedNames: '(_|ignored|expected|serialVersionUID)'
> > >     ignoreAnnotated: ['Preview']
> > > ```
> > 
> > 
> > @lydiasama For `ignoreAnnotated: ['Preview']`, I can't reproduce it. I don't get any warning from `Detekt`. That's why I think we don't need to use it, but if you can reproduce this case, please let me know🙏 I'm not sure if It's because the doc is not up to date or not 🤔
> 
> @Wadeewee I think it is because we no longer have `UnusedPrivateMember` in `detekt-config`. Normally when we define a private variable that is unused, we got the `detekt error`. Now we don't get the error anymore. Did you intentionally remove the rule? cc @lydiasama

Yep, @kaungkhantsoe, we haven't applied `UnusedPrivateMember` in `detekt-config.` I'm still thinking should we add it, but if we decide to add it, then we need to add `ignoreAnnotated: [ 'Preview' ]` for sure. What do you think? cc @lydiasama

kaungkhantsoe:
> @Wadeewee BTW did you try producing the detekt error after generating a compose template? I tried removing some rules which can cause the error and run detekt but no error is shown. Am I the only one facing it? Eg. I removed `ignoreAnnotated: [ 'Composable' ]` from `FunctionNaming` and run detekt but no error is shown. Am I doing something wrong? 🙏
> 
>  Screen.Recording.2022-12-14.at.1.08.19.PM.mov

What about this one @Wadeewee ? `detekt` is not working on my side. Could it be because of `gradle version`? @ryan-conway did you also face this issue while you were working on IC?

Wadeewee:
> > @Wadeewee BTW did you try producing the detekt error after generating a compose template? I tried removing some rules which can cause the error and run detekt but no error is shown. Am I the only one facing it? Eg. I removed `ignoreAnnotated: [ 'Composable' ]` from `FunctionNaming` and run detekt but no error is shown. Am I doing something wrong? 🙏
> > Screen.Recording.2022-12-14.at.1.08.19.PM.mov
> 
> What about this one @Wadeewee ? `detekt` is not working on my side. Could it be because of `gradle version`? @ryan-conway did you also face this issue while you were working on IC?

It works properly on my side @kaungkhantsoe I can get the waring here:

https://user-images.githubusercontent.com/28002315/207540932-e88c0ede-cd03-4a18-bb38-0fce8eafcd33.mov

Could you please try with this steps:
1. Run generate script to create new project
2. Once successful then run `./gradlew clean`
3. run `./gradlew detekt`

lydiasama:
> > > > Let's add `UnusedPrivateMember` to follow the [document](https://detekt.dev/docs/introduction/compose/#unusedprivatemember) as well.
> > > > ```
> > > > UnusedPrivateMember:
> > > >     active: true
> > > >     allowedNames: '(_|ignored|expected|serialVersionUID)'
> > > >     ignoreAnnotated: ['Preview']
> > > > ```
> > > 
> > > 
> > > @lydiasama For `ignoreAnnotated: ['Preview']`, I can't reproduce it. I don't get any warning from `Detekt`. That's why I think we don't need to use it, but if you can reproduce this case, please let me know🙏 I'm not sure if It's because the doc is not up to date or not 🤔
> > 
> > 
> > @Wadeewee I think it is because we no longer have `UnusedPrivateMember` in `detekt-config`. Normally when we define a private variable that is unused, we got the `detekt error`. Now we don't get the error anymore. Did you intentionally remove the rule? cc @lydiasama
> 
> Yep, @kaungkhantsoe, we haven't applied `UnusedPrivateMember` in `detekt-config.` I'm still thinking should we add it, but if we decide to add it, then we need to add `ignoreAnnotated: [ 'Preview' ]` for sure. What do you think? cc @lydiasama

I think it would be great if we follow the document. 🙏🏻

Wadeewee:
> > > > > Let's add `UnusedPrivateMember` to follow the [document](https://detekt.dev/docs/introduction/compose/#unusedprivatemember) as well.
> > > > > ```
> > > > > UnusedPrivateMember:
> > > > >     active: true
> > > > >     allowedNames: '(_|ignored|expected|serialVersionUID)'
> > > > >     ignoreAnnotated: ['Preview']
> > > > > ```
> > > > 
> > > > 
> > > > @lydiasama For `ignoreAnnotated: ['Preview']`, I can't reproduce it. I don't get any warning from `Detekt`. That's why I think we don't need to use it, but if you can reproduce this case, please let me know🙏 I'm not sure if It's because the doc is not up to date or not 🤔
> > > 
> > > 
> > > @Wadeewee I think it is because we no longer have `UnusedPrivateMember` in `detekt-config`. Normally when we define a private variable that is unused, we got the `detekt error`. Now we don't get the error anymore. Did you intentionally remove the rule? cc @lydiasama
> > 
> > 
> > Yep, @kaungkhantsoe, we haven't applied `UnusedPrivateMember` in `detekt-config.` I'm still thinking should we add it, but if we decide to add it, then we need to add `ignoreAnnotated: [ 'Preview' ]` for sure. What do you think? cc @lydiasama
> 
> @Wadeewee I think it would be great if we follow the document. 🙏🏻 cc @kaungkhantsoe

FYI: @lydiasama @kaungkhantsoe, I think it's good to have UnusedPrivateMember. It will be useful when we have unused private properties or functions that don't relate to Compose a preview. So, Let me add only template-compose and sample-compose in this PR, but for template-xml and sample-xml I will include it without ignoreAnnotated: ['Preview'] in this https://github.com/nimblehq/android-templates/pull/371 🙏

template-compose/detekt-config.yml
```diff
@@ -189,6 +189,7 @@ naming:
   FunctionNaming:
     active: true
     functionPattern: '^([a-z$][a-zA-Z$0-9]*)|(`.*`)$'
+    ignoreAnnotated: [ 'Composable' ]
```
luongvo, line 192: I don't see this attr in sample project config 🤔

luongvo: @Wadeewee please adding missing config then I can approve this PR 💁‍♂️

Wadeewee: @luongvo It has been added already in some previous PR checks at line `192` in `sample-compose` 

Screen Shot 2565-12-16 at 09 06 33

luongvo (approved):
lgtm

template-compose/detekt-config.yml
```diff
@@ -286,7 +287,7 @@ style:
     active: true
     ignoreNumbers: '-1,0,1,2'
```
Tuubz, line 288: @Wadeewee Shall we also add: `ignoreAnnotated: [ 'Preview' ]` for `MagicNumber`? 🙏 

For example, if we have a composable with an `Int` parameter:

```
@Composable
fun AppBar(height: Int) {
    TopAppBar(
        modifier = Modifier.height(height)
    )
}
```

In some cases, we would want to pass a raw integer value to the preview:

```
@Preview(showBackground = true)
@Composable
private fun AppBarPreview() {
    AppBar(height = 200)
}
```

What do you think? We did the same for [dee-money migration](https://github.com/nimblehq/dee-money-android/pull/34/files)

Tuubz (approved):
@Wadeewee Rest LGTM 🚀 

Besides my small comment, there's one more thing:

 `sample-compose/detekt-config.yml` & `template-compose/detekt-config.yml` are not 100% the same, should we update it in this PR and make them aligned? 🙏

luongvo:
> sample-compose/detekt-config.yml & template-compose/detekt-config.yml are not 100% the same, should we update it in this PR and make them aligned? 🙏

@Tuubz They are the same, you can diff their content together by using https://www.diffchecker.com/compare/. 
Please check her reply [here](https://github.com/nimblehq/android-templates/pull/362#discussion_r1050307355).

Tuubz:
@luongvo Oh lol, now I see the difference. The content is the same, but there's a small difference in the indentation 🤪 

Screen Shot 2022-12-20 at 11 18 11

AVI5HEK (approved):

doannimble (approved):
LGTM 🚀

template-compose/detekt-config.yml
Wadeewee: @Tuubz It's good to handle this for `Preview` in `MagicNumber` but I think we should consider using `PreviewParameterProvider` instead So I think we can get rid of this 🤔 

Also, From your example Is it can be possible right -> `AppBar(height = 200)` 🤔  because it should be suffix with  `.dp` -> `AppBar(height = 200.dp)` What do you think?

luongvo: @Tuubz My Detekt does not report it with the current config. Am I missing something from your concern?

image

Wadeewee: @Tuubz Could you please give me more some example that can be happen for passing `int` to `Preview` 🙏

luongvo: @Tuubz I made another example and it still does not report the issue 🤔 cc @Wadeewee 

image

luongvo:
> @luongvo Oh lol, now I see the difference. The content is the same, but the indentation is not formatted correctly 🤪

@Tuubz No, the content of those 2 config files is totally the same, even the indentation. The issue is they don't totally follow the IDE formatting, so it will re-format when copying-pasting (I guess you tried to copy and paste to check the diff in AS).

@Tuubz @Wadeewee In the best way, should we re-format the config file again (via AS)?

image

template-compose/detekt-config.yml
Tuubz: Well, this is embarrassing, I can't reproduce it either 😂 🙈 

Sorry for wasting your time, let's resolve this

Tuubz:
@Wadeewee Can you do the formatting as @luongvo suggested in #369? 🙏

Issue #352, opened 2022-11-22
[Compose] Setup the nav graph and navigation logic


luongvo:
## Why

Our current Android template for Jetpack Compose is too inceptive to start a new project. To make our Compose template truly designed for Jetpack Compose and follow the right principles, we need to work on various updates to the project structure.

In this task, we will set up the nav graph and navigation logic in Jetpack Compose. This follows what we have achieved in:
- Jetpack Compose Crypto showcase ([this PR](https://github.com/nimblehq/jetpack-compose-crypto/pull/49)).
- Knowledge sharing session about [Navigation in Jetpack Compose](https://docs.google.com/presentation/d/1xJYi25WrYOdMxigM0LTCIVX5zLozccCFGvq_ulyeTd8/edit#slide=id.g1432be656bf_0_62).

## Who Benefits?

Developers.

luongvo:
I reopen because of 1 remaining PR https://github.com/nimblehq/android-templates/pull/372

Issue #354, opened 2022-11-22
[Compose] Setup UI and Integration tests


luongvo:
## Why

Testing, including UI and Integration tests with Jetpack Compose, also requires a different set of boilerplate to begin with. And currently, there is no UI & Integration test setup for Jetpack Compose.

Hence, we will need to work on the following:

- Set up the foundation for UI & Integrating tests. This includes all the libraries it involves and base rules that developers can start writing easily.
- Sample test suites in Compose sample project for demonstrating purpose.

## Who Benefits?

Developers

PR #367, opened 2022-12-09
[#352] [Sample] 1/2: Setup the nav graph and navigation logic

luongvo:
https://github.com/nimblehq/android-templates/issues/352

## What happened 👀

- Set up nav graph https://developer.android.com/jetpack/compose/navigation.
- Provide ViewModel directly in composable by [`hiltViewModel()`](https://developer.android.com/jetpack/androidx/releases/hilt#hilt-navigation-compose_version_100_2). This solves the last known issue https://github.com/nimblehq/jetpack-compose-crypto/pull/45#pullrequestreview-1125833203 👏 .
- [Stop passing `NavController` to composable](https://github.com/chrisbanes/tivi/pull/799/), pass the `navigator` callback with `AppDestination` event instead. This `navigator` callback helps us to send the navigation command from our composable & ViewModel to the NavGraph for navigating.
- Clean up a lot of redundant unused/deprecated components for XML.
- 🚧  We will apply all necessary changes back to the Compose template in the second part 2/2. 🚧 
- 🚧  We will [migrate to use the new permission requesting flow for Compose](https://github.com/nimblehq/android-templates/pull/367/files#diff-365b60bb6954c48dbea6bd42ad0a7a2df630a504614d328f344289f05ce4c6ba) in https://github.com/nimblehq/android-templates/issues/370. 🚧 

## Insight 📝

Following what we introduced in https://docs.google.com/presentation/d/1xJYi25WrYOdMxigM0LTCIVX5zLozccCFGvq_ulyeTd8/edit#slide=id.p, we will have `AppDestination` class. `AppDestination` is designed to provide the `route` and `arguments` config for each composable screen. It's also a navigation command which contains a built destination for navigating.

- For nav graph config 🔨 
```
composable(
    route = AppDestination.Home.route,
    arguments = AppDestination.Home.arguments
) {
    HomeScreen(
        navigator = { destination -> navController.navigate(destination) }
    )
}
```

- For navigation command sending 🚀 
```
override fun onMyCoinsItemClick(coin: CoinItemUiModel) {
    execute {
        _navigator.emit(AppDestination.CoinDetail.buildDestination(coin.id))
    }
}
```

- For navigation command callback executing 💁‍♂️  
```
private fun NavHostController.navigate(destination: AppDestination) {
    when (destination) {
        is AppDestination.Up -> popBackStack()
        else -> navigate(route = destination.destination)
    }
}
```

## Proof Of Work 📹

The existing navigation from the `Home` to the `Second` screen should work properly as it is.


sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeComposeFragment.kt
```diff
@@ -1,63 +1,62 @@
...
+//        }
+//    }
+//
+//    private fun handlePermissionResult(permissionResult: PermissionResult) {
+//        when (permissionResult) {
+//            is Granted -> Timber.d("${permissionResult.grantedPermissions} granted")
+//            is Denied.NeedsRationale -> Timber.d("${permissionResult.deniedPermissions} needs rationale")
+//            else -> Timber.d("Request cancelled, missing permissions in manifest or denied permanently")
+//        }
+//    }
```
luongvo, line 61: FYI: 🚧 We will migrate to use the new permission requesting flow for Compose in the third part 3/3. 🚧

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppDestination.kt
```diff
@@ -0,0 +1,36 @@
...
+
+    object Home : AppDestination("home")
+
+    /**
+     * We can define route as "coin/details" without "coinId" parameter because we're passing it as argument already.
+     * So either passing "coinId" via arguments or passing it via route.
+     *
+     * We keep passing "coinId" in both route and arguments for this destination to give example of navigation concept
+     * about how to build a destination with parameters.
+     */
```
Wadeewee, line 25: Please update the description a bit  -->  just replace  some words of `"coin/details"` and `"coinId"`

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppNavigation.kt
```diff
@@ -0,0 +1,51 @@
...
+        route = destination.route,
+        arguments = destination.arguments,
+        deepLinks = deepLinks,
+        content = content
+    )
+}
+
+private fun NavHostController.navigate(destination: AppDestination) {
+    when (destination) {
+        is AppDestination.Up -> popBackStack()
```
Wadeewee, line 48: @luongvo Should it be `navigateUp()` instead of `popBackStack()` 🤔 
```suggestion
        is AppDestination.Up -> navigateUp()
```

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppDestination.kt
luongvo: @Wadeewee I think we can remove this comment now 👍

luongvo: Fixed in https://github.com/nimblehq/android-templates/pull/367/commits/292e3157b647217bffd88bcba8b6874063dc981e

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppNavigation.kt
luongvo: Fixed in https://github.com/nimblehq/android-templates/pull/367/commits/292e3157b647217bffd88bcba8b6874063dc981e

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeComposeScreen.kt
```diff
@@ -1,29 +1,64 @@
...
+import co.nimblehq.sample.compose.ui.userReadableMessage
 
 // TODO: Rename to 'HomeScreen'
 @Composable
 fun HomeComposeScreen(
+    viewModel: HomeComposeViewModel = hiltViewModel(),
+    navigator: (destination: AppDestination) -> Unit
+) {
+    val context = LocalContext.current
+    LaunchedEffect(Unit) {
```
Wadeewee, line 28: @luongvo  Should we define the key as `viewModel.error`  instead of `Unit` because we observe  `viewModel.error` to show this?
```suggestion
    LaunchedEffect(viewModel.error) {
```

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeComposeScreen.kt
```diff
@@ -1,29 +1,64 @@
...
+        viewModel.error.collect { error ->
+            val message = error.userReadableMessage(context)
+            Toast.makeText(context, message, Toast.LENGTH_SHORT).show()
+        }
+    }
+    LaunchedEffect(viewModel) {
```
Wadeewee, line 34: Same here 
```suggestion
    LaunchedEffect(viewModel.navigator) {
```

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeComposeScreen.kt
```diff
@@ -1,29 +1,64 @@
...
+
+    HomeScreenContent(
+        uiModels = uiModels,
+        showLoading = showLoading,
+        onItemClick = viewModel::navigateToSecond
+    )
+}
+
+@Composable
+private fun HomeScreenContent(
```
lydiasama, line 49: Since you created a new `HomeScreenContent` here, I think it makes sense to rename this class to `HomeScreen` altogether. What do you think?

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeComposeScreen.kt
```diff
@@ -1,29 +1,64 @@
...
+import co.nimblehq.sample.compose.ui.AppDestination
 import co.nimblehq.sample.compose.ui.screens.AppBar
-import co.nimblehq.sample.compose.ui.theme.ComposeSampleTheme
+import co.nimblehq.sample.compose.ui.theme.ComposeTheme
+import co.nimblehq.sample.compose.ui.userReadableMessage
 
 // TODO: Rename to 'HomeScreen'
 @Composable
 fun HomeComposeScreen(
+    viewModel: HomeComposeViewModel = hiltViewModel(),
```
lydiasama, line 24: Shall we rename this ViewModel too?
```suggestion
    viewModel: HomeViewModel = hiltViewModel(),
```

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeComposeScreen.kt
```diff
@@ -1,29 +1,64 @@
...
+        showLoading = showLoading,
+        onItemClick = viewModel::navigateToSecond
+    )
+}
+
+@Composable
+private fun HomeScreenContent(
     uiModels: List<UiModel>,
-    showLoading: Boolean,
+    showLoading: IsLoading,
```
lydiasama, line 51: I have a question. Should we remember this `showLoading` flag to reduce the recomposition in case there is no changed value?
```
var isShowLoading by remember { mutableStateOf(true) }
LaunchedEffect(key1 = showLoading) {
        isShowLoading = showLoading
    }
```

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeComposeScreen.kt
luongvo: @Wadeewee makes sense 👍

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeComposeScreen.kt
luongvo: @lydiasama renaming this class file name to HomeScreen with over-half file changes will break git history (removed > added new file) and messy git diff in this PR. I renamed it in the next PR https://github.com/nimblehq/android-templates/pull/372/files#diff-190b9b9927a540b81a9f5d2f5014dbb5a57ae16dc60fbdde8d29580fd97a5654.

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeComposeScreen.kt
luongvo: Same, fixed in https://github.com/nimblehq/android-templates/pull/372

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeComposeScreen.kt
luongvo: @lydiasama `showLoading` is managed by ViewModel by a Flow so we don't need to use `remember` to keep its state because it's a state already by using `collectAsState()`. Check out more detail here https://developer.android.com/jetpack/compose/state#use-other-types-of-state-in-jetpack-compose 👋

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeComposeFragment.kt
luongvo: Fixed in https://github.com/nimblehq/android-templates/pull/367/commits/d709212d3883491148b3081e2f7a98d91c150df2

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeComposeScreen.kt
luongvo: Fixed in https://github.com/nimblehq/android-templates/pull/367/commits/d709212d3883491148b3081e2f7a98d91c150df2

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeComposeScreen.kt
luongvo: Fixed in https://github.com/nimblehq/android-templates/pull/367/commits/d709212d3883491148b3081e2f7a98d91c150df2

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeComposeScreen.kt
lydiasama: @luongvo I see. So if we access it by `collectAsState()`, it will handle as `state` and we don't wanna concern about the recomposition. 👍🏻  Thank you for your explanation.

lydiasama (approved):

thiennguyen0196 (approved):
LGTM, strong implementation with the nav graph set up for Jetpack Compose @luongvo 👏

Wadeewee (approved):
LGTM Now🚀

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeComposeScreen.kt
```diff
@@ -1,29 +1,64 @@
...
+import co.nimblehq.sample.compose.lib.IsLoading
 import co.nimblehq.sample.compose.model.UiModel
+import co.nimblehq.sample.compose.ui.AppDestination
 import co.nimblehq.sample.compose.ui.screens.AppBar
-import co.nimblehq.sample.compose.ui.theme.ComposeSampleTheme
+import co.nimblehq.sample.compose.ui.theme.ComposeTheme
+import co.nimblehq.sample.compose.ui.userReadableMessage
 
 // TODO: Rename to 'HomeScreen'
 @Composable
```
lydiasama, line 22: @luongvo I found some interesting pattern from Tivi project. They have another composable function to call the `private HomeScreen` for encapsulating the viewModel so when calling the `HomeScreen` from the `AppNavigation`, it will not see the `viewModel` property and cannot pass any value to it. What do you think?

Here it is what the function looks like:
```kotlin
@Composable
fun HomeScreen(
) {
    HomeScreen(
        viewModel = hiltViewModel(),
        navigator = navigator
    )
}
```

And we can make the `fun HomeScreen` at line:23 to be private.

luongvo: @lydiasama As discussed [here](https://nimble-co.slack.com/archives/G41UCTGHL/p1671596317788699), maybe we can investigate it later in https://github.com/nimblehq/android-templates/issues/354 when actually working in testing 🤔

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppDestination.kt
```diff
@@ -0,0 +1,29 @@
...
+
+    open val arguments: List<NamedNavArgument> = emptyList()
+
+    open var destination: String = route
+
+    //====================================================//
+
+    object Up : AppDestination()
+
+    object Home : AppDestination("home")
```
Tuubz, line 17: Would it be good to create `const` for these?

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppDestination.kt
```diff
@@ -0,0 +1,29 @@
...
+
+    object Second : AppDestination("second/{$KEY_ID}") {
+
+        override val arguments = listOf(
+            navArgument(KEY_ID) { type = NavType.StringType }
+        )
+
+        fun buildDestination(id: String) = apply {
+            destination = "second/$id"
+        }
```
Tuubz, line 27: Would this work too? 🙏 

```suggestion
        fun buildDestination(id: String) = apply {
            destination.replace(KEY_ID, id)
        }
```

Tuubz (commented):
@luongvo Cool! 🥇

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppDestination.kt
lydiasama: @Tuubz I'm curious why do we need to back to Java way? 🤔

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppDestination.kt
```diff
@@ -0,0 +1,29 @@
+package co.nimblehq.sample.compose.ui
+
+import androidx.navigation.*
+
+const val KEY_ID = "id"
```
lydiasama, line 5: @luongvo Shall we use `KeyId` -- PascalCase as it is the convention from the [Jetpack Compose guideline](https://android.googlesource.com/platform/frameworks/support/+/androidx-main/compose/docs/compose-api-guidelines.md#kotlin-style-singletons_constants_sealed-class-and-enum-class-values-do)?

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/second/SecondScreen.kt
```diff
@@ -1,27 +1,40 @@
...
+import co.nimblehq.sample.compose.ui.AppDestination
 import co.nimblehq.sample.compose.ui.screens.AppBar
-import co.nimblehq.sample.compose.ui.theme.ComposeSampleTheme
+import co.nimblehq.sample.compose.ui.theme.ComposeTheme
 
 @Composable
-fun SecondScreen(uiModel: UiModel) {
+fun SecondScreen(
+    viewModel: SecondViewModel = hiltViewModel(),
+    navigator: (destination: AppDestination) -> Unit,
```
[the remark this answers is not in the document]
Wadeewee: @Tuubz But we have `SecondViewMode` in `sample-xml`; it seems like we missed some process to emit the value to `VM` for `sample-compose`. I think we should keep it because after we finish setup the modules, then we will have more examples about `Compose` 🤔 cc @luongvo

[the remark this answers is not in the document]
Tuubz: Hmmm, good catch @Wadeewee  👍 

@luongvo Would it make sense to also collect the `id` from `VM` for `sample-compose`? 🙏

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppDestination.kt
Wadeewee: @lydiasama Let's keep it as `SnakeCase` for now. I will create `RFC` for voting, So after we have a finalized change to `PascalCase` then we can create an issue and update it 🚀 cc @luongvo @Tuubz

lydiasama:
@luongvo Would you mind updating the `collectAsState` to `collectAsStateWithLifecycle` in this PR? 😁

luongvo:
> @luongvo Would you mind updating the `collectAsState` to `collectAsStateWithLifecycle` in this PR? 😁

@lydiasama I think it's out of this PR's scope and needs to verify & vote on. Can we go through an RFC and contribute to improve it later cc @Tuubz?

sample-compose/buildSrc/src/main/java/Configurations.kt
```diff
@@ -1,7 +1,7 @@
 object Flavor {
     const val PRODUCTION = "production"
     const val STAGING = "staging"
-    const val DIMENSIONS = "stage"
+    const val DIMENSION_VERSION = "version"
```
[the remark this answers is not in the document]
luongvo: @Tuubz Actually, we need to provide a dimension name here instead of a list of dimensions so the config name should be `DIMENSION_[SOMETHING]`. Please check out more detail here https://developer.android.com/reference/tools/gradle-api/7.2/com/android/build/api/dsl/CommonExtension#flavorDimensions().

https://github.com/nimblehq/android-templates/blob/519a71181e06ad9fb5bdcab00c88cc1a1de254fe/sample-compose/app/build.gradle.kts#L60

sample-compose/app/build.gradle.kts
```diff
@@ -122,22 +114,17 @@ dependencies {
...
     implementation("androidx.compose.material:material:${Versions.COMPOSE_VERSION}")
-
-    implementation("androidx.fragment:fragment-ktx:${Versions.ANDROIDX_FRAGMENT_KTX_VERSION}")
-    implementation("androidx.navigation:navigation-fragment-ktx:${Versions.ANDROIDX_NAVIGATION_VERSION}")
-    implementation("androidx.navigation:navigation-runtime-ktx:${Versions.ANDROIDX_NAVIGATION_VERSION}")
-    implementation("androidx.navigation:navigation-ui-ktx:${Versions.ANDROIDX_NAVIGATION_VERSION}")
+    implementation("androidx.navigation:navigation-compose:${Versions.COMPOSE_NAVIGATION_VERSION}")
 
     implementation("com.google.dagger:hilt-android:${Versions.HILT_VERSION}")
+    implementation("androidx.hilt:hilt-navigation-compose:${Versions.HILT_NAVIGATION_COMPOSE_VERSION}")
```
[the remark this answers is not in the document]
luongvo: @Tuubz Actually, 
- I ordered it by version name.
- That makes more sense to have Hilt dependency before having Hilt Navigation or Compose, right?

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeComposeScreen.kt
```diff
@@ -1,29 +1,64 @@
...
+
+    HomeScreenContent(
+        uiModels = uiModels,
+        showLoading = showLoading,
+        onItemClick = viewModel::navigateToSecond
+    )
+}
+
+@Composable
+private fun HomeScreenContent(
```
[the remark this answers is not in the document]
luongvo: @Tuubz @Wadeewee Detekt suggests a higher value for [LongParamaterList](https://detekt.dev/docs/introduction/compose/#longparameterlist) issue, should we increase the config for `LongMethod` issue also:
- Normally, a Composable method is quite long.
- I saw a discussion about this config and they suggest using `60` https://github.com/detekt/detekt/issues/3251.
- I'm using `60` for my KMM IC too and it seems to be a suitable config.

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/second/SecondScreen.kt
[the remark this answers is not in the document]
luongvo: @Tuubz Can you clarify your suggestion a bit clearer? 🙏

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppDestination.kt
```diff
@@ -0,0 +1,29 @@
...
+
+    open var destination: String = route
+
+    //====================================================//
+
+    object Up : AppDestination()
+
+    object Home : AppDestination("home")
+
+    object Second : AppDestination("second/{$KEY_ID}") {
```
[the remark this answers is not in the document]
luongvo: @Tuubz it's not about injecting a value into a String (`"second/{$KEY_ID}"`, not `"second/${KEY_ID}"`). Actually, it's a config for navigation route https://developer.android.com/jetpack/compose/navigation#nav-with-args 😛

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppDestination.kt
luongvo: @Tuubz with `sealed class`, it's unnecessary, like what we do with enums 😛 Detekt will not report the issues ;)
```
enum class AppDestination(value: String) {
    HOME("home"),
    DETAIL("detail")
}
```

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppDestination.kt
luongvo: @lydiasama I think we should only use PascalCase for Jetpack Compose components. This issue still makes me feel confused sometimes in usage lol. 😂

lydiasama: @luongvo I feel weird when we declare the PascalCase for Jetpack Compose components but leave it as snake_case in other classes. IMO. I think every class should have the same convention.

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppDestination.kt
luongvo: @Tuubz But I think it's not safe. Say `destination  = route = "identifierScreen/id"`. The code above will cause the issue with a wrong path `"123entifierScreen/123"` by replacing `id` with `123`.

Tuubz: Truuuue 😂 

Alright then, lets' keep it 🚀

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeComposeScreen.kt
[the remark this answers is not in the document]
Tuubz: Yes, I'm fine if you adjust the detekt rules 🚀 

cc @Wadeewee

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppDestination.kt
```diff
@@ -0,0 +1,29 @@
...
+
+const val KEY_ID = "id"
+
+sealed class AppDestination(val route: String = "") {
+
+    open val arguments: List<NamedNavArgument> = emptyList()
+
+    open var destination: String = route
+
+    //====================================================//
```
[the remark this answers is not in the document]
luongvo: @Tuubz Fixed in https://github.com/nimblehq/android-templates/pull/367/commits/368e505d67bfa6beb8a6f7ae3d3a3cc8239a0a64

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppNavigation.kt
```diff
@@ -0,0 +1,51 @@
...
+) {
+    composable(
+        route = destination.route,
+        arguments = destination.arguments,
+        deepLinks = deepLinks,
+        content = content
+    )
+}
+
+private fun NavHostController.navigate(destination: AppDestination) {
```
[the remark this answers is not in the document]
luongvo: @Tuubz Fixed in https://github.com/nimblehq/android-templates/pull/367/commits/368e505d67bfa6beb8a6f7ae3d3a3cc8239a0a64

sample-compose/app/build.gradle.kts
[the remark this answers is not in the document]
Tuubz: Aha, I see. 

I wasn't aware we were ordering by version names, instead of dependency name  👍

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppDestination.kt
luongvo: @lydiasama Alright, it's in the `app` module and contributes to Compose UI logic so I made an update here https://github.com/nimblehq/android-templates/pull/367/commits/a0e3714a56943abb4611184bb1563c5399e7797f

lydiasama (approved):
Except for the detekt rule updating.

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeComposeScreen.kt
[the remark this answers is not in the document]
Wadeewee: Yep, I'm good too 🚀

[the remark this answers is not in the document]
luongvo: Fixed in https://github.com/nimblehq/android-templates/pull/367/commits/8a7923fe5f57cc3b8f11f29730c271263dd270fd

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/second/SecondScreen.kt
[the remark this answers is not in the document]
Tuubz: **Conclusion**: https://nimble-co.slack.com/archives/G41UCTGHL/p1672136329771999?thread_ts=1672109944.295459&cid=G41UCTGHL

Tuubz (approved):

Issue #159, opened 2022-02-17
Support DataStore in a replacement for SharedPreference


sleepylee:
## Why

As recommended by Google, it is more efficient and practical to use DataStore instead of SharedPreference. Perhaps it is time for us to consider having this either as a mandatory or an option to initialize a project.

Read more:
- https://android-developers.googleblog.com/2020/09/prefer-storing-data-with-jetpack.html

- https://developer.android.com/topic/libraries/architecture/datastore

## Who Benefits?

Template users.

👍 : make it a default choice of storage
👎  : make it an optional choice after SharedPreference.

lydiasama:
@sleepylee Do we need another RFC about saving the preference synchronously or asynchronously?

sleepylee:
@lydiasama I think it is depending on the use-case of a project and we can improvise later. If you look at the comparison chart, by default, DataStore provides async API - but if I'm not wrong, you can always `await` for an async call to reply instead of making a synchronous call.

![image](https://user-images.githubusercontent.com/13435717/154405244-0eebfff4-81be-4c46-be35-3429d9078f12.png)

Tuubz:
Anyone interested in taking responsibility for this ticket? Currently we're using `NormalSharedPreferences.kt`, `EncryptedSharedPreferences.kt` & `BaseSharedPreferences.kt`, so we'll have to make some changes in those classes. 

`v3.7.0 - Release date: 30/04/22`

Respect points guaranteed: 

![2771433-5751838547-10692](https://user-images.githubusercontent.com/18277915/159403059-c876d099-3f34-4283-996d-7ee65b81e15c.jpeg)

lydiasama:
@Tuubz Let me try this one. 😁

Tuubz:
@lydiasama Yes, please do 🤩 🤩 🤩

Tuubz:
@lydiasama I think we will have to postpone adding new features, until after we finish [separating sample code into separate folder](https://github.com/nimblehq/android-templates/issues/193). Otherwise, there could be some conflicts, what do you think? 🙏

lydiasama:
@Tuubz Sorry for the late reply. It's okay. 🙏

luongvo:
@ryan-conway @lydiasama This issue should not be closed yet 👀

PR #374, opened 2022-12-14
[#159][Part 3] Add DataStore preference to template-xml and template-compose

lydiasama:
#159 

## What happened 👀

Part 1: Integrate for sample-compose
Part 2: Integrate for sample-xml
Part 3: Integrate for both template-xml & template-compose

## Insight 📝

- We still keep the `EncryptedSharePreferences` for the sake of security preferences.
- In terms of storing the data object, I think we should use `Room database` instead of using `DataStore Proto` because Room supports partial updates. So, I think we should focus on simple `key-value` preferences like `DataStore Preference`.
- I tried to keep the code simple using just `Preferences` to make it not seems like the sample code. Unfortunately, we can't do it with `UserPreferences` model because there is `android's Preferences` class too that can make the conflict import. So I'd suggested to leave it as `UserPreferences` instead.

## Proof Of Work 📹

The application runs successfully.


template-compose/app/src/main/java/co/nimblehq/template/compose/di/modules/StorageModule.kt
```diff
@@ -17,18 +16,8 @@ abstract class StorageModule {
 
     companion object {
 
-        @Provides
-        fun provideDefaultSharedPreferences(@ApplicationContext context: Context): SharedPreferences =
-            PreferenceManager.getDefaultSharedPreferences(context)
-
         @Provides
         fun provideSecuredLocalStorage(@ApplicationContext context: Context) =
             EncryptedSharedPreferences(context)
```
Tuubz, line 21: Hmmmm, sucks that there is no encrypted version of data-store 🥲

template-compose/app/src/main/java/co/nimblehq/template/compose/di/modules/DataStoreModule.kt
```diff
@@ -0,0 +1,27 @@
...
+import androidx.datastore.preferences.core.*
+import androidx.datastore.preferences.preferencesDataStoreFile
+import dagger.Module
+import dagger.Provides
+import dagger.hilt.InstallIn
+import dagger.hilt.android.qualifiers.ApplicationContext
+import dagger.hilt.components.SingletonComponent
+import javax.inject.Singleton
+
+private const val USER_PREFERENCES = "user_preferences"
```
Tuubz, line 14: Shall we ty to keep the naming for template code as general as possible?

- `USER_PREFERENCES` -> `PREFERENCES`
- `UserPreferencesRepository` -> `PreferencesRepository`
- `UserPreferences` -> `Preferences`

template-compose/buildSrc/src/main/java/Versions.kt
```diff
@@ -12,6 +12,7 @@ object Versions {
     const val ANDROID_COMMON_KTX_VERSION = "0.1.1"
     const val ANDROID_CRYPTO_VERSION = "1.0.0"
     const val ANDROIDX_CORE_KTX_VERSION = "1.9.0"
+    const val ANDROIDX_DATASTORE_PREFERENCE_VERSION = "1.0.0"
```
Tuubz, line 15:
```suggestion
    const val ANDROIDX_DATASTORE_PREFERENCES_VERSION = "1.0.0"
```

template-compose/domain/src/main/java/co/nimblehq/template/compose/domain/model/UserPreferences.kt
```diff
@@ -0,0 +1,6 @@
+package co.nimblehq.template.compose.domain.model
+
+data class UserPreferences(
+    val username: String,
+    val isActive: Boolean
+)
```
Tuubz, line 6:
```suggestion
data class UserPreferences(
    val id: String
)
```

What about this? In attempt to not make it look like sample code 🙏

Tuubz (commented):
@lydiasama 🔥

template-compose/app/src/main/java/co/nimblehq/template/compose/di/modules/StorageModule.kt
lydiasama: @Tuubz Yeahhhh. I mentioned it in my blog too. 😁

template-compose/app/src/main/java/co/nimblehq/template/compose/di/modules/DataStoreModule.kt
lydiasama: @Tuubz Good idea. 👍🏻 I have no idea about this. 😂 Will update.

template-compose/domain/src/main/java/co/nimblehq/template/compose/domain/model/UserPreferences.kt
lydiasama: @Tuubz Since we have no plan to update it into the sample repo, I intended to demonstrate how to access another type except for the simple string. 🤔 But we could remove it too.

Tuubz: Hmmmm, any reason we would not bring it into the sample repo? 🙏 

cc @nimblehq/android-chapter

lydiasama: @Tuubz I just don't see the issue with doing it. 😂 Could you please help me create the issue so that I can do it further?

Tuubz: I see 👍 

Anyways, I won't create a separate issue for it, since we didn't do it for other stories either ([Example](https://github.com/nimblehq/android-templates/issues/352)).

However, I will start specifying it in the description for future issues for better clarity.

Would that be OK? 🙏

lydiasama: @Tuubz Thanks. That could help. 🙏🏻

lydiasama: @Tuubz I updated the PR description. I'll add it to sample repos at part 2.

template-compose/data/src/main/java/co/nimblehq/template/compose/data/repository/UserPreferencesRepositoryImpl.kt
```diff
@@ -0,0 +1,54 @@
...
+import co.nimblehq.template.compose.domain.model.UserPreferences
+import co.nimblehq.template.compose.domain.repository.UserPreferencesRepository
+import kotlinx.coroutines.flow.*
+import java.io.IOException
+import javax.inject.Inject
+
+class UserPreferencesRepositoryImpl @Inject constructor(private val dataStore: DataStore<Preferences>) :
+    UserPreferencesRepository {
+
+    private val TAG: String = "UserPreferencesRepo"
```
Wadeewee, line 15: Should we make it as `const val` and put it on the top level of class 🤔 
```suggestion
    private const val TAG: String = "UserPreferencesRepo"
```

template-compose/data/src/main/java/co/nimblehq/template/compose/data/repository/UserPreferencesRepositoryImpl.kt
```diff
@@ -0,0 +1,54 @@
...
+            }
+
+            if (usernamePreferences.isActive != currentUserPreferences.isActive) {
+                preferences[PreferencesKeys.IS_ACTIVE] = usernamePreferences.isActive
+            }
+        }
+    }
+
+    private fun mapUserPreferences(preferences: Preferences): UserPreferences {
+        val username = preferences[PreferencesKeys.USERNAME] ?: ""
```
Wadeewee, line 50:
```suggestion
        val username = preferences[PreferencesKeys.USERNAME].orEmpty()
```

Wadeewee, line 50:
```suggestion
        val username = preferences[PreferencesKeys.USERNAME].orEmpty()
```

template-compose/data/src/main/java/co/nimblehq/template/compose/data/repository/UserPreferencesRepositoryImpl.kt
```diff
@@ -0,0 +1,54 @@
...
+        val USERNAME = stringPreferencesKey("username")
+        val IS_ACTIVE = booleanPreferencesKey("is_active")
+    }
+
+    override fun getUsernamePreferences(): Flow<UserPreferences> = dataStore.data
+        .catch { exception ->
+            // dataStore.data throws an IOException when an error is encountered when reading data
+            if (exception is IOException) {
+                Log.e(TAG, "Error reading preferences.", exception)
+                emit(emptyPreferences())
```
Wadeewee, line 27: @lydiasama I don't have a similar with `DataStore`, So, May I know why we prefer to emit `emptyPreferences()` instead of `throw exception` when we got `IOException` 🙏

Wadeewee:
Question 🙋‍♀️ @Tuubz @lydiasama In this PR we just provide the classes of `DataStore` for `template-xml` and `template-compose` right? Do we have the plan to add a sample usecase(Integrate part in `:app`) in another issue 🙏

template-compose/app/src/main/java/co/nimblehq/template/compose/di/modules/DataStoreModule.kt
```diff
@@ -0,0 +1,27 @@
...
+
+@Module
+@InstallIn(SingletonComponent::class)
+object DataStoreModule {
+
+    @Provides
+    @Singleton
+    fun providePreferencesDataStore(@ApplicationContext appContext: Context): DataStore<Preferences> {
+        return PreferenceDataStoreFactory.create(
+            produceFile = { appContext.preferencesDataStoreFile(USER_PREFERENCES) }
```
kaungkhantsoe, line 24: Shall we also define `corruptionHandler` in case `migration` is needed?
```
            corruptionHandler = ReplaceFileCorruptionHandler(
                produceNewData = { emptyPreferences() }
            )
```

lydiasama: @kaungkhantsoe Let's simplify it on our template. We don't need to handle the migration case. 🙏🏻

lydiasama:
> Question 🙋‍♀️ @Tuubz @lydiasama In this PR we just provide the classes of `DataStore` for `template-xml` and `template-compose` right? Do we have the plan to add a sample usecase(Integrate part in `:app`) in another issue 🙏

Please check out the PR description. 🙏🏻

template-compose/data/src/main/java/co/nimblehq/template/compose/data/repository/UserPreferencesRepositoryImpl.kt
lydiasama: @Wadeewee Because this is a template, so I don't think we need to handle the exception. 🙏🏻

template-compose/data/src/main/java/co/nimblehq/template/compose/data/repository/UserPreferencesRepositoryImpl.kt
lydiasama: @Wadeewee I don't think we should. As It's just a TAG string and we use it only inside this class. 🤔  To raise it as a constant will waste memory. Can you share the benefit to do that for me?

template-compose/data/src/main/java/co/nimblehq/template/compose/data/repository/UserPreferencesRepositoryImpl.kt
lydiasama: @Wadeewee For more information
Screenshot 2566-01-03 at 2 42 18 PM

Ref: https://developer.android.com/codelabs/android-preferences-datastore#5

template-compose/data/src/main/java/co/nimblehq/template/compose/data/repository/UserPreferencesRepositoryImpl.kt
Wadeewee: @lydiasama Because you use `UpperCase` -> `TAG`, that's why I thought you intended to make it as a `const val`, but If not, can we use just a small case for `tag` 🤔  Or If you have any inside please let me know 🙏

Wadeewee:
> > Question 🙋‍♀️ @Tuubz @lydiasama In this PR we just provide the classes of `DataStore` for `template-xml` and `template-compose` right? Do we have the plan to add a sample usecase(Integrate part in `:app`) in another issue 🙏
> 
> @Wadeewee Please check out the PR description. 🙏🏻

Sorry @lydiasama 😢 I didn't get which part of the description you refer about this 👇 

> Do we have the plan to add a sample usecase(Integrate part in `:app`)

Cloud you please guide me a bit 🙏

template-compose/data/src/main/java/co/nimblehq/template/compose/data/repository/UserPreferencesRepositoryImpl.kt
lydiasama: @Wadeewee Yeah. That's why I updated it already to be in the `companion object`. Please check again. 😬

lydiasama:
> > > Question 🙋‍♀️ @Tuubz @lydiasama In this PR we just provide the classes of `DataStore` for `template-xml` and `template-compose` right? Do we have the plan to add a sample usecase(Integrate part in `:app`) in another issue 🙏
> > 
> > 
> > @Wadeewee Please check out the PR description. 🙏🏻
> 
> Sorry @lydiasama 😢 I didn't get which part of the description you refer about this 👇
> 
> > Do we have the plan to add a sample usecase(Integrate part in `:app`)
> 
> Could you please guide me a bit 🙏 Maybe It's because I didn't similar with this much 🙈 cc @Tuubz

@Wadeewee The sample code and use case should be inside of the `Sample` repos. Although we implement the use case here in `Template` repo, we don't use it anywhere, right? So I think it makes more sense to implement it at Sample.

@Tuubz Next time, please includes these info in the issue. 🙏🏻

Wadeewee:
> > > > Question 🙋‍♀️ @Tuubz @lydiasama In this PR we just provide the classes of `DataStore` for `template-xml` and `template-compose` right? Do we have the plan to add a sample usecase(Integrate part in `:app`) in another issue 🙏
> > > 
> > > 
> > > @Wadeewee Please check out the PR description. 🙏🏻
> > 
> > 
> > Sorry @lydiasama 😢 I didn't get which part of the description you refer about this 👇
> > > Do we have the plan to add a sample usecase(Integrate part in `:app`)
> > 
> > 
> > Could you please guide me a bit 🙏 Maybe It's because I didn't similar with this much 🙈 cc @Tuubz
> 
> @Wadeewee The sample code and use case should be inside of the `Sample` repos. Although we implement the use case here in `Template` repo, we don't use it anywhere, right? So I think it makes more sense to implement it at Sample.
> 
> @Tuubz Next time, please include this info in the issue. 🙏🏻

@lydiasama, I totally agree with you that the sample code and use case should be inside the `sample` 👍 but, back to this PR, Should we `draft` this PR or not? Because we don't have a `validate/verify` that the code we implemented `works` as we don't have an integrated part yet? What do you think? cc @Tuubz

Tuubz:
> > > > > Question 🙋‍♀️ @Tuubz @lydiasama In this PR we just provide the classes of `DataStore` for `template-xml` and `template-compose` right? Do we have the plan to add a sample usecase(Integrate part in `:app`) in another issue 🙏
> > > > 
> > > > 
> > > > @Wadeewee Please check out the PR description. 🙏🏻
> > > 
> > > 
> > > Sorry @lydiasama 😢 I didn't get which part of the description you refer about this 👇
> > > > Do we have the plan to add a sample usecase(Integrate part in `:app`)
> > > 
> > > 
> > > Could you please guide me a bit 🙏 Maybe It's because I didn't similar with this much 🙈 cc @Tuubz
> > 
> > 
> > @Wadeewee The sample code and use case should be inside of the `Sample` repos. Although we implement the use case here in `Template` repo, we don't use it anywhere, right? So I think it makes more sense to implement it at Sample.
> > @Tuubz Next time, please include this info in the issue. 🙏🏻
> 
> @lydiasama, I totally agree with you that the sample code and use case should be inside the `sample` 👍 but, back to this PR, Should we `draft` this PR or not? Because we don't have a `validate/verify` that the code we implemented `works` as we don't have an integrated part yet? What do you think? cc @Tuubz

Hmmm, it's true.

@lydiasama Would we somehow be able to provide a PoW that does prove the current implementation works? 🙏

template-compose/data/src/main/java/co/nimblehq/template/compose/data/repository/UserPreferencesRepositoryImpl.kt
```diff
@@ -0,0 +1,54 @@
...
+            }
+        }.map { preferences ->
+            mapUserPreferences(preferences)
+        }
+
+    override suspend fun updateUsernamePreferences(usernamePreferences: UserPreferences) {
+        dataStore.edit { preferences ->
+            val currentUserPreferences = mapUserPreferences(preferences)
+
+            if (usernamePreferences.username != currentUserPreferences.username) {
```
Tuubz: Is it necessary to check the current value? I think it would be faster to do it without 💭

lydiasama: @Tuubz What do you think? Which one is worthier, checking the value before the update or updating every time calling the function? 🤔

Tuubz: @lydiasama With the current approach, we will always need to fetch the values first, which counts as **one** action 1️⃣ 

If we then need to update the value too, that would count as **another** action. 2️⃣ 

Summary:
- **"checking the value before the update"** : we will always have _one_ action, and possibly have _two_ actions.
- **"updating every time"** : we can be sure it will always execute _one_ action

However, it could be that **updating** is more "heavy" compared to **"fetching"** 💭 (not sure)

Please let me know if I got the wrong idea here, what do you think? 😁 

cc @nimblehq/android-chapter

lydiasama: @Tuubz What if this case, it updates and gets an error with the same value? Do we need to handle the exception like `Cannot update preferences` like the different value? If we don't check the same value before, perhaps, we need to check when getting the exception to handle in case `Cannot update with new value to preferences` or `Cannot update with the same value -> can ignore this exception`. What do you think?

Tuubz: @lydiasama Oh, does datastore throw an error if it updates with the same value? 😯

lydiasama: @Tuubz Nooo, I meant if it update with the same error but get the error while updating for some reason.

Wadeewee: @Tuubz @lydiasama IMO, I think we should work on a `sample` project first and try to integrate, So we can debug more and prove the concept of your idea 🙏

template-xml/data/src/main/java/co/nimblehq/template/xml/data/repository/UserPreferencesRepositoryImpl.kt
```diff
@@ -0,0 +1,54 @@
...
+import co.nimblehq.template.xml.domain.model.UserPreferences
+import co.nimblehq.template.xml.domain.repository.UserPreferencesRepository
+import kotlinx.coroutines.flow.*
+import java.io.IOException
+import javax.inject.Inject
+
+class UserPreferencesRepositoryImpl @Inject constructor(private val dataStore: DataStore<Preferences>) :
+    UserPreferencesRepository {
+
+    private val TAG: String = "UserPreferencesRepo"
```
thiennguyen0196, line 15: [Optional] 
Normally when defining `TAG` I usually put it in the `companion object` block and I see that in `PreferencesRepositoryImpl` you declared it the same way, so I think it may be better if we keep it consistently 🤔 
```suggestion
       companion object {
        private const val TAG = "UserPreferencesRepo"
    }
```

thiennguyen0196 (commented):
Finish my first round of review, there are just 2 optional code styles that I think it may be necessary to keep it consistently 😄

template-compose/data/src/main/java/co/nimblehq/template/compose/data/repository/PreferencesRepositoryImpl.kt
```diff
@@ -0,0 +1,50 @@
...
+        }
+    }
+
+    private fun mapUserPreferences(preferences: Preferences): UserPreferences {
+        val id = preferences[PreferencesKeys.ID] ?: ""
+        return UserPreferences(id)
+    }
+
+    companion object {
+        private const val TAG_PREFERENCES: String = "PreferencesRepo"
```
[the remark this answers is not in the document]
thiennguyen0196: Or we can remove it and use `canonicalName`/`simpleName` like @Tuubz suggested [here](https://github.com/nimblehq/android-templates/pull/374#discussion_r1061251378)
```
Log.e(UserPreferencesRepositoryImpl::class.java.canonicalName, "Error reading preferences.", exception)
```

template-xml/data/src/main/java/co/nimblehq/template/xml/data/repository/UserPreferencesRepositoryImpl.kt
thiennguyen0196: Ditto here with the usage of `canonicalName`/`simpleName`
```
Log.e(UserPreferencesRepositoryImpl::class.java.canonicalName, "Error reading preferences.", exception)
```

template-compose/app/src/main/java/co/nimblehq/template/compose/di/modules/DataStoreModule.kt
kaungkhantsoe: We should add the `corruptionHandler`. I once read it in the article that it could lead to `IOException` if it got corrupted. @lydiasama 🙏

luongvo:
@lydiasama Do we have part 2 or only 2 parts of this task?

lydiasama:
Let me clean it up. Will create a new PR. 🙏🏻

Issue #370, opened 2022-12-13
[Compose] Migrate to the new permission request logic


luongvo:
## Why

According to the [Setup the nav graph and navigation logic](https://github.com/nimblehq/android-templates/issues/352) task, we need to migrate all deprecated Fragments into Composables. Hence, migrating legacy permission requests on Fragment to the new approach in Jetpack Compose is needed.

In the end, we will stop using Peko in Compose Template and switch to using https://google.github.io/accompanist/permissions/.

## Who Benefits?

Developers

Tuubz:
@luongvo We broke the rules of our wiki for this issue:

- Every issue needs to go through a voting process first (unless I missed a discussion/RFC somewhere?)
- Always inform the repository maintainer before adding any issue to the sprint, otherwise we risk overestimating

Since you already created a PR for this, let's keep it, but please keep it in mind for next time 🙏

Tuubz:
@luongvo We broke the rules of our wiki for this issue:

- Every issue needs to go through a voting process first (unless I missed a discussion/RFC somewhere?)
- Always inform the repository maintainer before adding any issue to the sprint, otherwise, we risk overestimating

Since you already created a PR for this, let's keep it, but please keep it in mind for next time 🙏

luongvo:
@Tuubz Ah yes, sorry for the issue 🙏 

Actually, I intended to do it in the same PR of https://github.com/nimblehq/android-templates/issues/352 when removing the deprecated stuff of the XML method. Then I thought, a separate PR and a separate ticket would be better for it.

PR #375, opened 2022-12-14
[#370] [Compose] Migrate to the new permission request logic

luongvo:
Solves https://github.com/nimblehq/android-templates/issues/370

## What happened 👀

- Migrating legacy permission requests on Fragment to the new approach in Jetpack Compose by removing `Peko` and switching to use https://google.github.io/accompanist/permissions/.

## Insight 📝

From what we have done in this PR https://github.com/nimblehq/android-templates/pull/324, we need to migrate all related `Peko` permission request stuff to [Accompanist permissions](https://google.github.io/accompanist/permissions) for Compose projects.

## Proof Of Work 📹

https://user-images.githubusercontent.com/16315358/207667660-4b3faebc-7cb9-4dc0-ab7e-441f879c028e.mp4

### DIFF
--- a/sample-compose/app/src/main/AndroidManifest.xml
+++ b/sample-compose/app/src/main/AndroidManifest.xml
@@ -3,8 +3,6 @@
     package="co.nimblehq.sample.compose">
 
     <uses-permission android:name="android.permission.INTERNET" />
-    <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
-    <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
     <uses-permission android:name="android.permission.CAMERA" />
 
     <application
--- a/sample-compose/app/src/main/AndroidManifest.xml
+++ b/sample-compose/app/src/main/AndroidManifest.xml
@@ -3,6 +3,7 @@
     package="co.nimblehq.sample.compose">
 
     <uses-permission android:name="android.permission.INTERNET" />
+    <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
     <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
     <uses-permission android:name="android.permission.CAMERA" />
 
--- a/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeScreen.kt
+++ b/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeScreen.kt
@@ -1,12 +1,7 @@
 package co.nimblehq.sample.compose.ui.screens.home
 
 import android.Manifest.permission.*
-import android.content.Context
-import android.content.pm.PackageManager
 import android.widget.Toast
-import androidx.activity.compose.ManagedActivityResultLauncher
-import androidx.activity.compose.rememberLauncherForActivityResult
-import androidx.activity.result.contract.ActivityResultContracts
 import androidx.compose.foundation.layout.*
 import androidx.compose.material.CircularProgressIndicator
 import androidx.compose.material.Scaffold
@@ -15,7 +10,6 @@ import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.tooling.preview.Preview
-import androidx.core.content.ContextCompat
 import androidx.hilt.navigation.compose.hiltViewModel
 import co.nimblehq.sample.compose.R
 import co.nimblehq.sample.compose.lib.IsLoading
@@ -24,6 +18,7 @@ import co.nimblehq.sample.compose.ui.AppDestination
 import co.nimblehq.sample.compose.ui.screens.AppBar
 import co.nimblehq.sample.compose.ui.theme.ComposeTheme
 import co.nimblehq.sample.compose.ui.userReadableMessage
+import com.google.accompanist.permissions.*
 import timber.log.Timber
 
 @Composable
@@ -52,35 +47,26 @@ fun HomeScreen(
     )
 }
 
+@OptIn(ExperimentalPermissionsApi::class)
 @Composable
 private fun HomeScreenContent(
     uiModels: List<UiModel>,
     showLoading: IsLoading,
     onItemClick: (UiModel) -> Unit
 ) {
-    val permissions = arrayOf(
-        ACCESS_COARSE_LOCATION,
-        ACCESS_FINE_LOCATION,
-        CAMERA
-    )
-    val launcherMultiplePermissions = rememberLauncherForActivityResult(
-        ActivityResultContracts.RequestMultiplePermissions()
-    ) { permissionsMap ->
-        val areGranted = permissionsMap.values.reduce { acc, next -> acc && next }
-        if (areGranted) {
-            Timber.d("${permissionsMap.keys} granted")
+    val cameraPermissionState = rememberPermissionState(CAMERA)
+    if (cameraPermissionState.status.isGranted) {
+        Timber.d("${cameraPermissionState.permission} granted")
+    } else {
+        if (cameraPermissionState.status.shouldShowRationale) {
+            Timber.d("${cameraPermissionState.permission} needs rationale")
         } else {
             Timber.d("Request cancelled, missing permissions in manifest or denied permanently")
         }
-    }
 
-    val context = LocalContext.current
-    LaunchedEffect(Unit) {
-        checkAndRequestLocationPermissions(
-            context,
-            permissions,
-            launcherMultiplePermissions
-        )
+        LaunchedEffect(Unit) {
+            cameraPermissionState.launchPermissionRequest()
+        }
     }
 
     Scaffold(topBar = {
@@ -100,26 +86,6 @@ private fun HomeScreenContent(
                 )
             }
         }
-    }
-}
-
-fun checkAndRequestLocationPermissions(
-    context: Context,
-    permissions: Array<String>,
-    launcher: ManagedActivityResultLauncher<Array<String>, Map<String, Boolean>>
-) {
-    if (
-        permissions.all {
-            ContextCompat.checkSelfPermission(
-                context,
-                it
-            ) == PackageManager.PERMISSION_GRANTED
-        }
-    ) {
-        // Permissions are already granted
-    } else {
-        // Request permissions
-        launcher.launch(permissions)
     }
 }
 
--- a/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeScreen.kt
+++ b/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeScreen.kt
@@ -39,7 +39,7 @@ fun HomeScreen(
     val showLoading: IsLoading by viewModel.showLoading.collectAsState()
     val uiModels: List<UiModel> by viewModel.uiModels.collectAsState()
 
-    CheckCameraPermission()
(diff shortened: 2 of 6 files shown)


Tuubz:
@luongvo We broke the rules of our wiki for this issue:

- Every issue needs to go through a voting process first (unless I missed a discussion/RFC somewhere?)
- Always inform the repository maintainer before adding any issue to the sprint, otherwise, we risk overestimating

Since you already created a PR for this, let's keep it, but please keep it in mind for next time 🙏

Tuubz:
@luongvo One more thing, I think we're using the wrong keyword in the PR description (please check [here](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword )) 🚀

luongvo:
> @luongvo One more thing, I think we're using the wrong keyword in the PR description (please check [here](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)) 🚀

Updated ✅

luongvo:
> @luongvo We broke the rules of our wiki for this issue:
> 
> * Every issue needs to go through a voting process first (unless I missed a discussion/RFC somewhere?)
> * Always inform the repository maintainer before adding any issue to the sprint, otherwise, we risk overestimating
> 
> Since you already created a PR for this, let's keep it, but please keep it in mind for next time 🙏

@Tuubz I replied in the ticket instead https://github.com/nimblehq/android-templates/issues/370#issuecomment-1352497064 👌

sample-compose/app/src/main/AndroidManifest.xml
```diff
@@ -3,7 +3,6 @@
     package="co.nimblehq.sample.compose">
 
     <uses-permission android:name="android.permission.INTERNET" />
-    <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
     <uses-permission android:name="android.permission.CAMERA" />
```
Tuubz, line 6: Should we request for both`CAMERA` & `LOCATION` permissions, so it's the same with `sample-xml`? 🙏

luongvo: @Tuubz As this is a sample, yessss

luongvo: @Tuubz I have re-thought this issue, according to this documentation https://google.github.io/accompanist/permissions/#rememberpermissionstate-and-remembermultiplepermissionsstate-apis, we can easily do `multiple permission request` by using `rememberMultiplePermissionsState`. However, applying `rememberMultiplePermissionsState` requires replacing the current sample with `single permission request`. So I think we just need to make a sample for one of them. I don't think we have enough effort to make many samples for all cases and maintain them later 🙏 

IMO, I will ignore this man.

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeScreen.kt
```diff
@@ -44,12 +47,28 @@ fun HomeScreen(
...
+        if (cameraPermissionState.status.shouldShowRationale) {
+            Timber.d("${cameraPermissionState.permission} needs rationale")
+        } else {
+            Timber.d("Request cancelled, missing permissions in manifest or denied permanently")
+        }
+
+        LaunchedEffect(Unit) {
+            cameraPermissionState.launchPermissionRequest()
+        }
+    }
```
hoangnguyen92dn, line 70: As the sample, could we show on Toast instead of Timber? 🤔

hoangnguyen92dn (approved):
LGTM 🚀

doannimble (approved):
Good to go 🏃‍♂️

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeScreen.kt
luongvo: @hoangnguyen92dn I think Timber logging is good enough. Adding Toast will require us to take care of more things from Toast implementation, it could be an easy step but out of scope 🙏

hoangnguyen92dn: @luongvo Yeah. My idea is that simply launch the application then play around to see the result rather than take a look at the log. Anw, it's not a big deal. We could skip it for now.

sample-compose/app/src/main/AndroidManifest.xml
Tuubz: > However, applying rememberMultiplePermissionsState requires replacing the current sample with single permission request

Could you please explain a bit more on this? 🙇  

I thought we just had to simply change to:

`val cameraPermissionState = rememberMultiplePermissionsState(listOf(CAMERA, LOCATION))`

luongvo: @Tuubz Do you want to have 2 samples `rememberPermissionState` + `rememberMultiplePermissionsState ` or only `rememberMultiplePermissionsState`?

Tuubz (approved):
💪 🇻🇳

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeScreen.kt
```diff
@@ -44,12 +47,28 @@ fun HomeScreen(
...
+    if (cameraPermissionState.status.isGranted) {
+        Timber.d("${cameraPermissionState.permission} granted")
+    } else {
+        if (cameraPermissionState.status.shouldShowRationale) {
+            Timber.d("${cameraPermissionState.permission} needs rationale")
+        } else {
+            Timber.d("Request cancelled, missing permissions in manifest or denied permanently")
+        }
+
+        LaunchedEffect(Unit) {
```
Wadeewee, line 67: Let's define the key instead of `Unit`
```suggestion
        LaunchedEffect(cameraPermissionState) {
```

Wadeewee, line 67: @luongvo May I know why we need to call `LaunchedEffect` inside `else` block here 🤔

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeScreen.kt
Wadeewee, line 70: @luongvo I think we should move the section of handing the `state` to `HomeScreen` instead to keep `HomeScreenContent` as `Stateless`, What do you think? 🙏 
```suggestion

```

Wadeewee (commented):
I'm done my review for LGTM 🚀

Tuubz:
@luongvo Any chance we can resolve the last remaining comments by end of today? (I'm off tomorrow)

If not, no biggie, then I'll create the release on 3th of January instead 🙏

luongvo:
> @luongvo Any chance we can resolve the last remaining comments by end of today? (I'm off tomorrow)
> 
> If not, no biggie, then I'll create the release on 3th of January instead 🙏

@Tuubz okay, let me switch the tasks in the meantime, it's in my plan today 👌

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeScreen.kt
luongvo: @Wadeewee You're right. However, the permission check composable needs to be separate from the `HomeScreenContent` and even `HomeScreen` composable to avoid re-composition.

Fixed in https://github.com/nimblehq/android-templates/pull/375/commits/decb5e2d24caf91d6a9ac742d4ea717c5b24f17c

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeScreen.kt
luongvo: @Wadeewee Nope, we only need to execute it 1 time each state changes. It will be re-executed when `cameraPermissionState` changes.

luongvo: @Wadeewee we need to request to show the `permission request dialog` for the user to approve the required permission in case of `shouldShowRationale` or `denied` 🤓

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeScreen.kt
luongvo: @hoangnguyen92dn Fixed in https://github.com/nimblehq/android-templates/pull/375/commits/decb5e2d24caf91d6a9ac742d4ea717c5b24f17c

luongvo:
@hoangnguyen92dn @Wadeewee @Tuubz I have a critical update for this permission request flow `Move permission check into new composable to avoid re-composition by HomeScreenContent`, please check the latest commit 🙏

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeScreen.kt
Wadeewee: Ah, I see because It's state, So, That makes sense to use `Unit` 👍

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeScreen.kt
```diff
@@ -37,13 +39,42 @@ fun HomeScreen(
     val showLoading: IsLoading by viewModel.showLoading.collectAsState()
     val uiModels: List<UiModel> by viewModel.uiModels.collectAsState()
 
+    CheckCameraPermission()
```
Tuubz, line 42: I could be wrong, but should composable functions be named as a "noun", not a "verb"? 💭 
```suggestion
    CameraPermission()
```

luongvo: @Tuubz Fixed in https://github.com/nimblehq/android-templates/pull/375/commits/108f99bbe905146a8ae4a93da059979976a1ef7a

Wadeewee (approved):
LGTM 🚀

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeScreen.kt
```diff
@@ -37,13 +39,42 @@ fun HomeScreen(
...
+
     HomeScreenContent(
         uiModels = uiModels,
         showLoading = showLoading,
         onItemClick = viewModel::navigateToSecond
     )
 }
 
+/**
+ * The permission check composable needs to be separate from the HomeScreenContent composable to avoid re-composition
```
Tuubz, line 52: @luongvo Optional:

```suggestion
 * [CameraPermission] needs to be separate from [HomeScreenContent] to avoid re-composition
```

Would you like to update it? If not, I'll merge it now 🤖

luongvo: Fixed in https://github.com/nimblehq/android-templates/pull/375/commits/eb95572869dd994af2a445fe91e9c785d9ac9374

PR #382, opened 2023-01-06
[#228] Add TokenAuthenticator to template-xml

### LINKED ISSUE #228: Add TokenAuthenticator setup
## Why

Most of apps use API request and have `refresh token` mechanism to request new access token in case of expired. We don't have the setup for the initial `TokenAuthenticator` yet so we should add it into Coroutine template.

## Who Benefits?

`Developers` who use this template to initialize their project with network request.

I think this is **essential** because most of apps have API request and have refresh token mechanism. Having a standard setup would help us a lot in development and reduce mistakes.

Example of implementation: https://github.com/nimblehq/toyota-wallet-android/blob/main/data/src/main/java/co/omise/gcpf/service/authenticator/ApplicationRequestAuthenticator.kt

kaungkhantsoe:
- Closes #228 

## What happened 👀

Provide a description of the **changes** this pull request brings to the codebase. Additionally, when the pull request is still being worked on, a checklist of the planned changes is welcome to track progress.

## Insight 📝

Describe in detail why this solution is the most appropriate, which solution you tried but did not go with, and how to test the changes. References to relevant documentation are welcome as well.

## Proof Of Work 📹

Show us the implementation: screenshots, GIFs, etc.

### DIFF
--- a/template-xml/app/src/main/java/co/nimblehq/template/xml/di/Authenticate.kt
+++ b/template-xml/app/src/main/java/co/nimblehq/template/xml/di/Authenticate.kt
@@ -1,0 +1,7 @@
+package co.nimblehq.template.xml.di
+
+import javax.inject.Qualifier
+
+@Qualifier
+@Retention(AnnotationRetention.BINARY)
+annotation class Authenticate
--- a/template-xml/app/src/main/java/co/nimblehq/template/xml/di/modules/AuthModule.kt
+++ b/template-xml/app/src/main/java/co/nimblehq/template/xml/di/modules/AuthModule.kt
@@ -1,0 +1,22 @@
+package co.nimblehq.template.xml.di.modules
+
+import co.nimblehq.template.xml.data.repository.SessionManagerImpl
+import co.nimblehq.template.xml.data.repository.TokenRefresherImpl
+import co.nimblehq.template.xml.data.service.AuthService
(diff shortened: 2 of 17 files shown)


Tuubz:
@luongvo @kaungkhantsoe I think let's postpone this for now, this issue is a lot bigger than I originally estimated.

It's best to create another issue first, to integrate a new API that works with a (refresh) token.

What do you think? 🙏

ryan-conway:
>It's best to create another issue first, to integrate a new API that works with a (refresh) token.

@Tuubz Do you mean to integrate an API into the template? It might be a better idea to add that to the sample projects instead (perhaps the survey API?) and keep only the authenticator stuff in the template, what do you think?

Tuubz:
@ryan-conway Sorry for the late reply, but yes, you're correct 🙏 

Please have a look at this thread: https://nimble-co.slack.com/archives/G41UCTGHL/p1614156128022800

IMO, I think we should go with GitHub API (convenient) or create a Mock API with Postman (flexible).

I'll create an RFC for this, what do you think? 🙏

ryan-conway:
@Tuubz An RFC sounds like a good idea 😄

Tuubz:
@ryan-conway @kaungkhantsoe RFC created: https://github.com/nimblehq/android-templates/discussions/434 🚀

Issue #373, opened 2022-12-14
Update Compose dependencies using BOM for versioning all libs


lydiasama:
## Why

After Google released [BOM](https://developer.android.com/jetpack/compose/setup#bom-version-mapping) for the compose libraries, it helps us to manage the latest versions of libraries we are using. Therefore, we don't need to handle it independently. However, if we want to fix version to some libraries, we can do it by [overriding the library version](https://developer.android.com/jetpack/compose/setup#how_do_i_use_a_different_library_version_than_whats_designated_in_the_bom).

![Screenshot 2565-12-14 at 12 29 48 PM](https://user-images.githubusercontent.com/12026942/207513856-670f3e5c-195e-4cbe-8580-e54a5bc47cbf.png)

## Who Benefits?

Android devs

hoangnguyen92dn:
@lydiasama Applying BOM for dependencies versioning management, we must be aware of new versions released from Google 🤔 It might cause unexpected issues, and we are always forced to update to adapt with new releases 🐛 How do you think we could keep the specific version for some core of libraries instead of all libraries? 😄

lydiasama:
@hoangnguyen92dn To use `BOM` != `@latest`. You can see [here](https://developer.android.com/jetpack/compose/setup#bom-version-mapping), each BOM version specifies the libraries version. So there is no concern about keeping up to date with every lib. Furthermore, if there are some libraries you don't want to use the same version as BOM suggests, you can override it as I explain in the PR description.

hoangnguyen92dn:
@lydiasama I see the idea, but once you update the BOM version, all the dependencies in the library group will be updated as well and we have to verify to make sure functionalities work fine after upgrading. 

Seems we could rely on this explanation from Google, we could give it a try

> The BOM itself has links to the stable versions of the different Compose libraries, in such a way that they work well together.

https://developer.android.com/jetpack/compose/setup#using-the-bom

lydiasama:
@hoangnguyen92dn From my experience, when we update the libraries, we usually update all relevant libraries together, no? 🤔  I know your concern but I think BOM is practical and can be used. What do you think?

lydiasama:
@hoangnguyen92dn Again, when updating a library, if you don't want them to be affected all, you can override it one by one first and then use the next version of BOM later. it's cleaner than handling one-by-one version and creating a lot of version variables, IMHO.

hoangnguyen92dn:
@lydiasama 
[...]

PR #388, opened 2023-01-10
[#373] Update Compose dependecies using BOM version

Wadeewee:
https://github.com/nimblehq/android-templates/issues/373

## What happened 👀

After Google released [BOM](https://developer.android.com/jetpack/compose/setup#bom-version-mapping) for the compose libraries, it helps us to manage the latest versions of libraries we are using. Therefore, we don't need to handle it independently. However, if we want to fix version to some libraries, we can do it by [overriding the library version](https://developer.android.com/jetpack/compose/setup#how_do_i_use_a_different_library_version_than_whats_designated_in_the_bom).

## Insight 📝

Once we applied `BOM` we can get rid of the specifying version of these libs:
- `androidx.compose.ui:ui`
- `androidx.compose.ui:ui-tooling`
- `androidx.compose.foundation:foundation`
- `androidx.compose.material:material`

But, there're two libs that's relate with `compose`,  but it doesn't type of `BOM`:

- `androidx.navigation:navigation-compose`
- `androidx.hilt:hilt-navigation-compose`

So, Regarding the libs above, we can't include them in `BOM` version.

## Proof Of Work 📹

The app can normally run in both `template-compose` and `sample-compose`.

Screen Shot 2566-01-10 at 17 14 23

Screen Shot 2566-01-10 at 17 13 52

### DIFF
--- a/sample-compose/app/build.gradle.kts
+++ b/sample-compose/app/build.gradle.kts
@@ -114,13 +114,15 @@ dependencies {
 
     implementation(fileTree(mapOf("dir" to "libs", "include" to listOf("*.jar"))))
 
+    implementation(platform("androidx.compose:compose-bom:${Versions.COMPOSE_BOM_VERSION}"))
+    implementation("androidx.compose.ui:ui")
+    implementation("androidx.compose.ui:ui-tooling")
+    implementation("androidx.compose.foundation:foundation")
+    implementation("androidx.compose.material:material")
+
     implementation("androidx.core:core-ktx:${Versions.ANDROIDX_CORE_KTX_VERSION}")
     implementation("androidx.lifecycle:lifecycle-runtime-ktx:${Versions.ANDROIDX_LIFECYCLE_VERSION}")
 
-    implementation("androidx.compose.ui:ui:${Versions.COMPOSE_VERSION}")
-    implementation("androidx.compose.ui:ui-tooling:${Versions.COMPOSE_VERSION}")
-    implementation("androidx.compose.foundation:foundation:${Versions.COMPOSE_VERSION}")
-    implementation("androidx.compose.material:material:${Versions.COMPOSE_VERSION}")
     implementation("androidx.navigation:navigation-compose:${Versions.COMPOSE_NAVIGATION_VERSION}")
 
     implementation("com.google.dagger:hilt-android:${Versions.HILT_VERSION}")
--- a/sample-compose/app/build.gradle.kts
+++ b/sample-compose/app/build.gradle.kts
@@ -114,16 +114,17 @@ dependencies {
 
     implementation(fileTree(mapOf("dir" to "libs", "include" to listOf("*.jar"))))
 
+    implementation("androidx.core:core-ktx:${Versions.ANDROIDX_CORE_KTX_VERSION}")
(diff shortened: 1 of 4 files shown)


template-compose/app/build.gradle.kts
```diff
@@ -114,13 +114,15 @@ dependencies {
...
+    implementation("androidx.compose.material:material")
+
     implementation("androidx.core:core-ktx:${Versions.ANDROIDX_CORE_KTX_VERSION}")
     implementation("androidx.lifecycle:lifecycle-runtime-ktx:${Versions.ANDROIDX_LIFECYCLE_VERSION}")
 
-    implementation("androidx.compose.ui:ui:${Versions.COMPOSE_VERSION}")
-    implementation("androidx.compose.ui:ui-tooling:${Versions.COMPOSE_VERSION}")
-    implementation("androidx.compose.foundation:foundation:${Versions.COMPOSE_VERSION}")
-    implementation("androidx.compose.material:material:${Versions.COMPOSE_VERSION}")
     implementation("androidx.navigation:navigation-compose:${Versions.COMPOSE_NAVIGATION_VERSION}")
```
luongvo, line 126: I think we should move the Compose BOM back here to have the same group with other compose components, such as `navigation-compose`. We can move `com.google.accompanist:accompanist-permissions` up to here for the same group too 🤔 

```
    implementation(platform("androidx.compose:compose-bom:${Versions.COMPOSE_BOM_VERSION}"))
    implementation("androidx.compose.ui:ui")
    implementation("androidx.compose.ui:ui-tooling")
    implementation("androidx.compose.foundation:foundation")
    implementation("androidx.compose.material:material")

    implementation("androidx.navigation:navigation-compose:${Versions.COMPOSE_NAVIGATION_VERSION}")
    implementation("com.google.accompanist:accompanist-permissions:${Versions.ACCOMPANIST_PERMISSIONS_VERSION}")
```

sample-compose/app/build.gradle.kts
```diff
@@ -114,13 +114,15 @@ dependencies {
...
+    implementation("androidx.compose.material:material")
+
     implementation("androidx.core:core-ktx:${Versions.ANDROIDX_CORE_KTX_VERSION}")
     implementation("androidx.lifecycle:lifecycle-runtime-ktx:${Versions.ANDROIDX_LIFECYCLE_VERSION}")
 
-    implementation("androidx.compose.ui:ui:${Versions.COMPOSE_VERSION}")
-    implementation("androidx.compose.ui:ui-tooling:${Versions.COMPOSE_VERSION}")
-    implementation("androidx.compose.foundation:foundation:${Versions.COMPOSE_VERSION}")
-    implementation("androidx.compose.material:material:${Versions.COMPOSE_VERSION}")
     implementation("androidx.navigation:navigation-compose:${Versions.COMPOSE_NAVIGATION_VERSION}")
```
luongvo, line 126: same, grouping issue

luongvo (commented):
rest lgtm

FYI: 
- BOM is not created by Google, it belongs to Apache Maven instead https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#bill-of-materials-bom-poms
- Someone asks Dagger (Hilt) to use BOM too https://github.com/google/dagger/issues/2122

chornerman (approved):
LGTM 🚀

luongvo (approved):

Tuubz (approved):
Cool! 🚀

PR #389, opened 2023-01-11
[#159][Part1] [Compose] Support datastore replacing shared preference

lydiasama:
https://github.com/nimblehq/android-templates/issues/159

## What happened 👀

Part 1: Integrate for sample-compose
Part 2: Integrate for sample-xml
Part 3: Integrate for both template-xml & template-compose

## Insight 📝

- Add the DataStore Preferences dependencies to the `Sample-Compose` project.
- Add use cases for `GetFirstTimeLaunchPreferencesUseCase` and `UpdateFirstTimeLaunchPreferencesUseCase`.
- At first, I attempt to call the use cases by the following approach:
```kotlin
init {
        execute {
            showLoading()
            getModelsUseCase()
                .catch {
                    _error.emit(it)
                }
                .collect { result ->
                    val uiModels = result.toUiModels()
                    _uiModels.emit(uiModels)
                }
            getFirstTimeLaunchUseCase()
                .catch {
                    _error.emit(it)
                }
                .collect { firstTimeLaunch ->
                    Timber.d("firstTimeLaunch: $firstTimeLaunch")
                    _firstTimeLaunch.emit(firstTimeLaunch)
                    if (firstTimeLaunch) {
                        updateFirstTimeLaunchUseCase(false)
                    }
                    Timber.d("firstTimeLaunch: after if")
                }
            hideLoading()
        }
    }
```
But it has never come to do the `hideLoading()` function. It seems it never ends the `collect()` function. So I changed the approach to be:

```kotlin
init {
        execute {
            showLoading()
            val getModelsFlow = getModelsUseCase()
            val getFirstTimeLaunchPreferencesFlow = getFirstTimeLaunchPreferencesUseCase()

            getModelsFlow.combine(getFirstTimeLaunchPreferencesFlow) { uiModels, firstTimeLaunch ->
                _uiModels.emit(uiModels.toUiModels())

                _firstTimeLaunch.emit(firstTimeLaunch)
                if (firstTimeLaunch) {
                    updateFirstTimeLaunchPreferencesUseCase(false)
                }
            }.catch {
                _error.emit(it)
                hideLoading()
            }.collect {
                hideLoading()
            }
        }
    }
```

And what's interesting is, when I tried the `.onCompletion`, it will do the `onCompletion` block when gets error, but not execute it when it succeeds. So in success case, it will not hide loading if we do the following code:
```kotlin
init {
        execute {
            val getModelsFlow = getModelsUseCase()
            val getFirstTimeLaunchPreferencesFlow = getFirstTimeLaunchPreferencesUseCase()
           ...
            }.onStart {
                showLoading()
            }.onCompletion {
                hideLoading()                
            }.catch {
                _error.emit(it)
            }.collect()
        }
    }
```

After discussing with @Wadeewee, we can wrap the `appPreferencesDataStore.data` with `flow`, so that we can use `.first()` to get only the first emit from the DataStore preferences and it can finish the `collect()` function.

💡  Here is the solution:

```kotlin
init {
        execute {
            val getModelsFlow = getModelsUseCase()
            val getFirstTimeLaunchPreferencesFlow = getFirstTimeLaunchPreferencesUseCase()
           ...
            }.onStart {
                showLoading()
            }.onCompletion {
                hideLoading()
            }.catch {
                _error.emit(it)
            }.collect()
        }
    }
```

- 💡 I have an idea to combine the `_uiModels`, `_firstTimeLaunch`, and `showLoading` flow in `HomeViewState`. So we can get rid of their public variables. Moreover, we can just observe only 

```kotlin
val homeViewState: HomeViewState by viewModel.homeViewState.collectAsState()
```
in `HomeScreen`. BTW, I'm not sure if we could reduce the recomposition by doing this because when the HomeViewState's properties are changed, it might trigger to recompose anyway. So I'll keep the codebase as the same.

- I added these attributes to `AndroidManifest` because when I test uninstall and reinstall the app, sometimes the preferences does not remove entirely. But instead, it keep some caches. So it's better to not allow caching.
Ref: https://stackoverflow.com/a/33176898 

```
        android:allowBackup="false"
        android:fullBackupContent="false"
```

## Proof Of Work 📹

Success case:

https://user-images.githubusercontent.com/12026942/211796491-5669d9af-f4b8-4c86-ad8f-4012c404744f.mp4

Error case:

https://user-images.githubusercontent.com/12026942/212041785-ba686014-38fa-4dc8-9f1d-1fce29927fa6.mp4

How does it look like in the device's storage:

Screenshot 2566-01-11 at 6 18 42 PM


lydiasama:
@Tuubz Updated. 🚀

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeViewModel.kt
```diff
@@ -12,26 +13,51 @@ import javax.inject.Inject
...
+        HomeViewState(
+            uiModels = uiModels,
+            firstTimeLaunch = firstTimeLaunch,
+            showLoading = showLoading
+        )
+    }.stateIn(
+        scope = viewModelScope,
+        started = SharingStarted.WhileSubscribed(5000),
+        initialValue = HomeViewState()
+    )
```
Wadeewee, line 40: @lydiasama, I think we should avoid combining these streams together:
- `_uiModels`
- `_firstTimeLaunch`
- `showLoading`

If I'm not wrong, it's because whenever one of the streams was triggered (**emit**), the rest streams were triggered, too(**trigger** **latest** **value**).  And I'm pretty sure this `_firstTimeLaunch` stream, we required it only once time to show the toast. So, the streams should be independent of each other in this case. What do you think? 🙏 cc @nimblehq/android-chapter

sample-compose/data/src/main/java/co/nimblehq/sample/compose/data/repository/AppPreferencesRepositoryImpl.kt
```diff
@@ -0,0 +1,44 @@
...
+                emit(emptyPreferences())
+            } else {
+                throw exception
+            }
+        }.map { preferences ->
+            preferences[PreferencesKeys.FIRST_TIME_LAUNCH] ?: true
+        }.first()
+
+        emit(firstLaunchPreferences)
+    }
```
Wadeewee, line 37: @lydiasama I have an idea. Instead of `emit(emptyPreferences())` when `IOException` and then filter it to `true`. What about if we give retries(**maybe just 3 times**) 👇 , but if it's still `IOException`, then we should let it throw an error as normal. Does it make sense? 🤔 

Also, I think in this way, we don't need to `handle errors` or `emit` value here, So,  just use the `mapper` -> `flowTransform` as it handles `(error/emit)` already. What do you think? cc @nimblehq/android-chapter 
```suggestion
    override fun getFirstTimeLaunchPreferences(): Flow<Boolean> = flowTransform {
        appPreferencesDataStore.data
            .retry(retries = 3) { exception ->
                (exception is IOException).also { isIOException ->
                    if (isIOException) delay(1000)
                }
            }
            .map { preferences ->
                preferences[PreferencesKeys.FIRST_TIME_LAUNCH] ?: true
            }.first()
    }
```

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeViewModel.kt
```diff
@@ -12,26 +13,51 @@ import javax.inject.Inject
...
-                    _uiModels.emit(uiModels)
+            val getModelsFlow = getModelsUseCase()
+            val getFirstTimeLaunchPreferencesFlow = getFirstTimeLaunchPreferencesUseCase()
+
+            getModelsFlow.combine(getFirstTimeLaunchPreferencesFlow) { uiModels, firstTimeLaunch ->
+                _uiModels.emit(uiModels.toUiModels())
+
+                _firstTimeLaunch.emit(firstTimeLaunch)
+                if (firstTimeLaunch) {
+                    updateFirstTimeLaunchPreferencesUseCase(false)
```
Wadeewee, line 52: What about if we return `Unit` instead of `Boolean`? Then we can get rid of the condition in `HomeScreen` inside `LaunchedEffect` scope.
```suggestion
                if (firstTimeLaunch) {
                 _firstTimeLaunch.emit(Unit)
                    updateFirstTimeLaunchPreferencesUseCase(false)
```

sample-compose/data/src/main/java/co/nimblehq/sample/compose/data/repository/AppPreferencesRepositoryImpl.kt
lydiasama: @Wadeewee Regarding retrying, I don't think we should do it. Since we don't usually do it when using the SharePreferences, so I think we should keep the same approach on the DataStore as well.

For the error mapping that you suggested, as you said `we don't need to handle errors or emit value here`, so I think we shouldn't `mapError` function from the `flowTransform`. Because even we map the error from the DataStore to that function, it will return as `else` case anyway that could waste the CPU performance. cc @nimblehq/android-chapter

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeViewModel.kt
lydiasama: Nice suggestion. @Wadeewee  👍🏻

sample-compose/data/src/main/java/co/nimblehq/sample/compose/data/repository/AppPreferencesRepositoryImpl.kt
Wadeewee: > Regarding retrying, I don't think we should do it. Since we don't usually do it when using the SharePreferences

@lydiasama, In `SharePreferences` we don't need a handler `IOExeption` when read/write. That's why I said It's weird that the `code lab` suggested handling `IOExeption` when using `DataStore`, and because we have handled `IOExeption`, that's why I suggest retrying to avoid emitting `empty` value. 🙏 


> so I think we should keep the same approach on the DataStore as well.

Regarding this ☝️ If we use the same approach, May I know how we handle this case:

- When entering the app but `not for the first time`, That's mean we have set `false` for  `updateFirstTimeLaunch` once we enter it for the first time.
- And then we get `IOExeption` (**The current value should be false**), but whenever `IOExeption`, we will emit `emptyPreferences()` and then map the value to `true`. Is that mean we will show `toast` and `update value` again? 🤔 

Please correct me if I missed something. cc @Tuubz

lydiasama: @Wadeewee Whenever you call a function, it will use the CPU every time. And for this case, when you go into the `mapError` function, it will return the same exception you pass. So it's useless to do that. 😅

Wadeewee: > I don't think we need to repeat the fetching 3 times here

Oh, @kaungkhantsoe, I don't suppose to retry 3 times as well,  just give an example of using `retry` to handling `IOExeption`

lydiasama: @Wadeewee I think @kaungkhantsoe  means when it returns error `IOException`, it means there was an issue about the device. So even if we retry to connect the device again, it will return the same `IOException` forever. Hence, we don't need to keep retrying it.

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeViewModel.kt
lydiasama: @Wadeewee After retrying your approach, it won't show the toast properly. So I think let's do it as-is. 🙏🏻

lydiasama: @Wadeewee We can't declare the `MutableStateFlow` without the initial value, right? 😢 So this would emit to the `HomeScreen` by default. We can't use the Unit.

Screenshot 2566-01-26 at 5 13 36 PM

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeViewModel.kt
lydiasama: @Wadeewee I rethink about this and found out we passed the `uiModels` and `showLoading` variable to the `HomeScreenContent` like:

```kotlin
HomeScreenContent(
        uiModels = uiModels,
        showLoading = showLoading,
        onItemClick = viewModel::navigateToSecond
    )
```

So when whether the `uiModels` or `showLoading` have an update, it will recompose this `HomeScreenContent` again. I'm not sure about if we pass the `homeViewState` to the `HomeScreenContent` instead, will it recompose when there is a variable updated? We might need some suggestion from @kaungkhantsoe 🙋🏻

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeViewModel.kt
lydiasama: @Wadeewee Unfortunately, I tried, but it doesn't work. So let me keep the Boolean 🙏🏻

Wadeewee: > I tried, but it doesn't work

@lydiasama, Not sure, Is this you stuck: 

By using `MutableStateFlow` and giving the `initial value` as `Unit` It will doesn't work because whenever we enter the screen, it will trigger the `default` value `immediately`. So, we need to use `MutableSharedFlow` instead to let it trigger only when we want to emit it. The case is like when we emit ` _navigator` or `_error`.

If not this ☝️ Please let me know what are you stuck why it doesn't work 🙏

lydiasama: @Wadeewee Yes, I tried with the `MutibleSharedFlow`, but it doesn't work. Could you please try on your side if I missed something?

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeViewModel.kt
Wadeewee: > Will it recompose when there is a variable updated?

@lydiasama , Regarding recompose, that's why we don't pass any state to `HomeScreenContent` to make it `Stateless` function:

```kotlin
HomeScreenContent(
        uiModels = uiModels,
        showLoading = showLoading,
        onItemClick = viewModel::navigateToSecond
    )
```
And because of `HomeScreenContent` is `Stateless` function then, we need to collect state in `Stateful` function, which is the `parent/root` function -> `HomeScreen` to avoid `recomposition`

```kotlin
fun HomeScreen( //------Stateful Fucntion--------//
) {
....
    val showLoading: IsLoading by viewModel.showLoading.collectAsState()
    val uiModels: List<UiModel> by viewModel.uiModels.collectAsState()
    
     HomeScreenContent(//------Stateless Fucntion--------//
        uiModels = uiModels,
        showLoading = showLoading,
        onItemClick = viewModel::navigateToSecond
    )
}
```

> I'm not sure about if we pass the homeViewState to the HomeScreenContent instead:

And Regarding this, I think it's the same:

```kotlin
val homeViewState: HomeViewState by viewModel.homeViewState.collectAsState()

---------------VS---------------

val showLoading: IsLoading by viewModel.showLoading.collectAsState()
val uiModels: List<UiModel> by viewModel.uiModels.collectAsState()

/**
* Both ways we keep and collect it in `Stateful function`(HomeScreen), 
* and then we pass only `value` to `Stateless function`(HomeScreenContent`)
*/

```
That's from my understanding, Please let me know if you have any suggestion @kaungkhantsoe cc @Tuubz 🙇‍♀️

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeViewModel.kt
Wadeewee: @lydiasama May I know do you still keep `firstTimeLaunch` in `HomeViewState` by using `MutibleSharedFlow` 🤔

sample-compose/data/src/main/java/co/nimblehq/sample/compose/data/repository/AppPreferencesRepositoryImpl.kt
Wadeewee: @lydiasama, TBH, I don't think we need to handle `IOException` for the current approach. It shouldn't take more time to get(**read**) the value that causes it `IOException`. It would be great if we could reproduce it.  but if we decide to emit `empty` when `IOException`, What do you think about this case? Can it be happen?

> - When entering the app but not for the first time, That's mean we have set false for updateFirstTimeLaunch once we enter it for the first time.
> - And then we get IOExeption (The current value should be false), but whenever IOExeption, we will emit emptyPreferences() and then map the value to true. Is that mean we will show toast and update value again? 🤔

cc @Tuubz

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeViewModel.kt
chornerman: Could you please try this 🙏🏼 

```
// HomeViewModel
@HiltViewModel
    private val getModelsUseCase: GetModelsUseCase,
    private val getFirstTimeLaunchPreferencesUseCase: GetFirstTimeLaunchPreferencesUseCase,
    private val updateFirstTimeLaunchPreferencesUseCase: UpdateFirstTimeLaunchPreferencesUseCase,
    dispatchers: DispatchersProvider
) : BaseViewModel(dispatchers) {

    private val _uiModels = MutableStateFlow<List<UiModel>>(emptyList())
        get() = _uiModels

    private val _firstTimeLaunch = MutableSharedFlow<Unit>()
    val firstTimeLaunch: SharedFlow<Unit>
        get() = _firstTimeLaunch

    init {
        execute {
            val getModelsFlow = getModelsUseCase()
            val getFirstTimeLaunchPreferencesFlow = getFirstTimeLaunchPreferencesUseCase()

            getModelsFlow.combine(getFirstTimeLaunchPreferencesFlow) { uiModels, firstTimeLaunch ->
                _uiModels.emit(uiModels.toUiModels())

                if (firstTimeLaunch) {
                    _firstTimeLaunch.emit(Unit)
                    updateFirstTimeLaunchPreferencesUseCase(false)
                }
            }.onStart {
                showLoading()
            }.onCompletion {
                hideLoading()
            }.catch {
                _error.emit(it)
            }.collect()
        }
    }

    fun navigateToSecond(uiModel: UiModel) {
        execute { _navigator.emit(AppDestination.Second.buildDestination(uiModel.id)) }
    }
}
```

```
// HomeScreen
@Composable
fun HomeScreen(
) {
    val showLoading: IsLoading by viewModel.showLoading.collectAsState()
    val uiModels: List<UiModel> by viewModel.uiModels.collectAsState()
//    val firstTimeLaunch: Boolean by viewModel.firstTimeLaunch.collectAsState()

    val context = LocalContext.current
    LaunchedEffect(viewModel.error) {
        viewModel.error.collect { error ->
            val message = error.userReadableMessage(context)
            Toast.makeText(context, message, Toast.LENGTH_SHORT).show()
        }
    }
    LaunchedEffect(viewModel.firstTimeLaunch) {
//        delay(1000)
        viewModel.firstTimeLaunch.collect {
            Toast.makeText(context, "This is the first time launch", Toast.LENGTH_SHORT).show()
        }
//        if (firstTimeLaunch) {
//            Toast.makeText(context, "This is the first time launch", Toast.LENGTH_SHORT).show()
//        }
    }
    LaunchedEffect(viewModel.navigator) {
        viewModel.navigator.collect { destination -> navigator(destination) }
    }

    CameraPermission()

    HomeScreenContent(
        uiModels = uiModels,
        showLoading = showLoading,
        onItemClick = viewModel::navigateToSecond
    )
}
```
Thanks @Wadeewee for the help 🙏🏼

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeViewModel.kt
lydiasama: @Wadeewee I think it's the same too. So I would prefer to wrap it inside the `homeViewState` to raise the best practice about handling state here. What do you think?

sample-compose/data/src/main/java/co/nimblehq/sample/compose/data/repository/AppPreferencesRepositoryImpl.kt
lydiasama: @Wadeewee 
> And then we get IOExeption (The current value should be false), but whenever IOExeption, we will emit emptyPreferences() and then map the value to true. Is that mean we will show toast and update value again? 

About this case, I think if it gets IOException, then returning the default preferences > return `true` makes sense to me. Since the IOException handling is recommended by Google's codelab, so I think we should handle it too.

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeViewModel.kt
lydiasama: @chornerman I tried, but it doesn't show the toast message at all. Can you show the toast message by this code, I'm curious?

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeScreen.kt
```diff
@@ -19,26 +19,34 @@ import co.nimblehq.sample.compose.ui.screens.AppBar
...
+
     val context = LocalContext.current
     LaunchedEffect(viewModel.error) {
         viewModel.error.collect { error ->
             val message = error.userReadableMessage(context)
             Toast.makeText(context, message, Toast.LENGTH_SHORT).show()
         }
     }
+    LaunchedEffect(firstTimeLaunch) {
+        delay(1000)
```
doannimble, line 41: @lydiasama I know that it is only the sample, but could we put the delay in `ViewModel`? I'm not feeling good when we block the UI thread 🤔

doannimble (commented):
rest lgtm

sample-compose/data/src/main/java/co/nimblehq/sample/compose/data/repository/AppPreferencesRepositoryImpl.kt
kaungkhantsoe: @Wadeewee According to [this issue tracker](https://issuetracker.google.com/issues/173522155), not using a single instance of the datastore can cause `IOException`. cc @lydiasama

lydiasama: @kaungkhantsoe I think I implemented the single instance of the datastore already. 🤔 You can see in the module, I used `@singleton`. Or did I miss something?

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeViewModel.kt
kaungkhantsoe: From the [official](documentation), what `combine` does is
> Returns a [Flow](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-flow/index.html) whose values are generated with [transform](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/combine.html) function by combining the most recently emitted values by each flow.

And just like @Wadeewee said, we should normally use `combine` when the streams are dependent on each other. But from Compose view, like @lydiasama said `HomeScreenContent` will try to recompose. It doesn't mean all the contents of `HomeScreenContent` will recompose. Assume the following as an example:
```
HomeScreenContent(val text1: String, val text2: String) {
   Text(text1)
   Text(text2)
}
```
Let's say originally `text1=Hello` and `text2=World`. If we try to pass `text2=Team` to `HomeScreenContent`, it will try to recompose every child. If you check the recomposition counts of `Text(text1)` and `Text(text2)`, you will see `Text(text1)` skipped the recomposition and `Text(text2)` recomposed once. 

In summary, if a parent compose view got different value parameters, it will try to recompose every child compose views. But only the child compose views with different values will be recomposed and the other will skip the recomposition.

I tried to remove some items in our item list using `HomeScreenState` and without using `HomeScreenState`. The recomposition count is the same as follows: 🙏 

|With HomeScreenState|Without HomeScreenState|
|--|--|
|https://user-images.githubusercontent.com/32578035/215387523-027bef91-2e9e-4f84-bdaf-dce875d0c188.mov|https://user-images.githubusercontent.com/32578035/215387571-2bfb4a26-83f3-4333-a392-186f5f0ef06f.mov|

sample-compose/data/src/main/java/co/nimblehq/sample/compose/data/repository/AppPreferencesRepositoryImpl.kt
kaungkhantsoe: @lydiasama I am just answering what could cause `IOException` with `DataStore`. 😬 You did it right 👍

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeViewModel.kt
lydiasama: @kaungkhantsoe I think so. So I think there's no difference if we use `HomeScreenState` or not. And also, I think the code is cleaner when using it because we can wrap every stream inside one parameter. What do you think? cc @Wadeewee @nimblehq/android-chapter

sample-compose/data/src/main/java/co/nimblehq/sample/compose/data/repository/AppPreferencesRepositoryImpl.kt
```diff
@@ -0,0 +1,44 @@
...
+    override fun isFirstTimeLaunchPreferences(): Flow<Boolean> = flow {
+        val firstLaunchPreferences = appPreferencesDataStore.data
+            .catch { exception ->
+                if (exception is IOException) {
+                    Log.e(
+                        AppPreferencesRepositoryImpl::class.simpleName,
+                        "Error reading preferences.",
+                        exception
+                    )
+                    emit(emptyPreferences())
```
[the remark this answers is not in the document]
lydiasama: @chornerman I answered [here](https://github.com/nimblehq/android-templates/pull/389#discussion_r1088821962). I think handling the `IOException` is good to follow since it is what Google suggested. And the way we are handling it is just our sample to do, right? So in this case, I wanna create the `Empty Preference`, which means it will return the `default` value to the view. In this way, we will not see the `IOException` on the view and will see `This is the first time` instead. 
Thinking of the real use case, users don't need to know if there is something wrong with the device. Imagine we got exactly `IOException`, then what will we handle further? Just throw it and show a toast message, and the user will be blocked? I think we shouldn't handle it like that. Instead, returning an `Empty preference with default value` can resolve this. So it will not block users and will continue to flow smoothly.
cc @nimblehq/android-chapter

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeViewModel.kt
```diff
@@ -12,26 +12,39 @@ import javax.inject.Inject
...
-                    _error.emit(it)
-                }
-                .collect { result ->
-                    val uiModels = result.toUiModels()
-                    _uiModels.emit(uiModels)
+            val getModelsFlow = getModelsUseCase()
+            val getFirstTimeLaunchPreferencesFlow = getFirstTimeLaunchPreferencesUseCase()
+
+            getModelsFlow.combine(getFirstTimeLaunchPreferencesFlow) { uiModels, firstTimeLaunch ->
+                _uiModels.emit(uiModels.toUiModels())
```
kaungkhantsoe, line 35: Quick question: Why are we emitting new value like this manually? What is the difference between
```
_uiModels.emit(uiModels.toUiModels())
```
and
```
_uiModels.value = uiModels.toUiModels()
```

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeScreen.kt
```diff
@@ -19,26 +19,34 @@ import co.nimblehq.sample.compose.ui.screens.AppBar
...
+import kotlinx.coroutines.delay
 
 @Composable
 fun HomeScreen(
     viewModel: HomeViewModel = hiltViewModel(),
     navigator: (destination: AppDestination) -> Unit
 ) {
+    val showLoading: IsLoading by viewModel.showLoading.collectAsState()
+    val uiModels: List<UiModel> by viewModel.uiModels.collectAsState()
+    val firstTimeLaunch: Boolean by viewModel.firstTimeLaunch.collectAsState()
```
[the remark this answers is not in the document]
lydiasama: @doannimble I thought about it too. But if we change this, how about `showLoading`? Shall we change it to `isShowLoading` for consistency? 🤔

sample-compose/app/src/main/java/co/nimblehq/sample/compose/di/modules/PreferencesRepositoryModule.kt
```diff
@@ -0,0 +1,33 @@
...
+@InstallIn(SingletonComponent::class)
+class PreferencesRepositoryModule {
+
+    @Provides
+    @Singleton
+    fun provideAppPreferencesDataStore(
+        @ApplicationContext applicationContext: Context
+    ): DataStore<Preferences> {
+        return applicationContext.appPreferencesDataStore
+    }
```
kaungkhantsoe, line 28: Let's do it in one line for consistency
```suggestion
    ): DataStore<Preferences> =
        applicationContext.appPreferencesDataStore
```

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeScreen.kt
lydiasama: @doannimble The point is we have too many toast messages to show here. We cannot guarantee when it emits the `firstTimeLanch` stream here because it's all asynchronous. Does it make more sense to put this delay on the `ViewModel`?

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeScreen.kt
```diff
@@ -19,26 +19,34 @@ import co.nimblehq.sample.compose.ui.screens.AppBar
...
 import co.nimblehq.sample.compose.ui.userReadableMessage
 import com.google.accompanist.permissions.*
+import kotlinx.coroutines.delay
 
 @Composable
 fun HomeScreen(
     viewModel: HomeViewModel = hiltViewModel(),
     navigator: (destination: AppDestination) -> Unit
 ) {
+    val showLoading: IsLoading by viewModel.showLoading.collectAsState()
```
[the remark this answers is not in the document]
lydiasama: @doannimble It's from @luongvo's implementation. 😂 I just moved it up. Shall we change it to Java style like this?

sample-compose/app/src/main/java/co/nimblehq/sample/compose/di/modules/PreferencesRepositoryModule.kt
```diff
@@ -0,0 +1,33 @@
...
+import co.nimblehq.sample.compose.data.repository.AppPreferencesRepositoryImpl
+import co.nimblehq.sample.compose.domain.repository.AppPreferencesRepository
+import dagger.Module
+import dagger.Provides
+import dagger.hilt.InstallIn
+import dagger.hilt.android.qualifiers.ApplicationContext
+import dagger.hilt.components.SingletonComponent
+import javax.inject.Singleton
+
+private val Context.appPreferencesDataStore by preferencesDataStore("app_preferences")
```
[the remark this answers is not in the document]
lydiasama: @doannimble Interesting. Let me try. 🙏🏻

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeViewModel.kt
lydiasama: @kaungkhantsoe It's the same. 😁 
https://stackoverflow.com/a/66673965
I can see your PR is using `_navigate.emit` too. https://github.com/nimblehq/android-templates/pull/390/files#diff-4f89083574fd3be2e7422e65bfd3d0f0ae5e5c2321733e28a66a82c4fc44605fR39. 😂

sample-compose/data/src/main/java/co/nimblehq/sample/compose/data/repository/AppPreferencesRepositoryImpl.kt
Wadeewee: > About this case, I think if it gets IOException, then returning the default preferences > return true makes sense to me. 

@lydiasama, I just want to clarify this ☝️, So are you saying, It makes sense to `show a toast` with the message `firstInstall` again, even if it's not a first install, right? 🙏

sample-compose/data/src/main/java/co/nimblehq/sample/compose/data/repository/AppPreferencesRepositoryImpl.kt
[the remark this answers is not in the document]
Wadeewee: > Thinking of the real use case, users don't need to know if there is something wrong with the device. Imagine we got exactly IOException, then what will we handle further? Just throw it and show a toast message, and the user will be blocked?

@lydiasama, I agree with you on this point, So that should we still `combine` the source of `getModelsUseCase()` and `getModelsUseCase()` in `VM` as it will be block when one of the sources returns an error? What do you think? 🙏

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeViewModel.kt
Wadeewee: @lydiasama, I can confirm it works, as I have seen when he debugs. @chornerman, Do you still have the code on your side? If it's still there so, could you please test again and screen record for her 🙇‍♀️

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeViewModel.kt
kaungkhantsoe: @lydiasama I didn't notice it when I was doing my PR. I just edited `buildDestination` to `createRoute` 😬 . It is alright. I just asked because I am curious. Both works.

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeViewModel.kt
kaungkhantsoe: There is a solution if we don't want to use `combine`. We can separate functions for setting each states by `copy`.

sample-compose/data/src/main/java/co/nimblehq/sample/compose/data/repository/AppPreferencesRepositoryImpl.kt
lydiasama: @Wadeewee It's correct. That's my intention to handle it that way. Since this is just an example, we don't have any requirements for handling it. So let's keep it simple pls. 🙏🏻

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeViewModel.kt
lydiasama: @Wadeewee @chornerman Please show the record. 🙏🏻 Because I did what exactly you guys did, but it didn't work as we expected.

sample-compose/data/src/main/java/co/nimblehq/sample/compose/data/repository/AppPreferencesRepositoryImpl.kt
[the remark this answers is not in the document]
lydiasama: @Wadeewee My point in combining them all together is I'd like to demonstrate how we handle the chain use cases in a real project. When working on a client project, we usually combine all use cases to ensure the result will not be the `race-condition` case. 

Besides, I think it's acceptable to do it in this case. The worst case of the DataStore is the `IOException` that was covered with `Empty preferences with the default value`. So I don't think there will be another exception returns from `getFirstTimeLaunchPreferencesUseCase`. After testing several times, I didn't see the exception at all.

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeViewModel.kt
chornerman: @lydiasama @Wadeewee Here it is 🙏🏼 


https://user-images.githubusercontent.com/13492460/215442129-9f95edd1-d18d-47ec-b50f-e41c724b00ea.mp4

lydiasama: @chornerman Thanks Chorn. It might be because my Samsung has a permission popup at the bottom, and the toast message shows on that period. 😢

chornerman: @lydiasama That is weird 🤔 I tested it on Samsung (Android 12) and it seems fine

https://user-images.githubusercontent.com/13492460/215448038-75a290b4-d586-41bb-93bf-74ad662ccd29.mp4

ANW, if this approach is not working on some devices, I'm okay with using boolean 🙌🏼

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeScreen.kt
thiennguyen0196: Not so sure about this delay here @lydiasama 🤔 When I tested it on my device without the `delay(1000)` it works fine and can show the Toast `This is the first time launch` normally 😄 

https://user-images.githubusercontent.com/25218255/216553501-2a08c739-2c7b-44f6-8920-4a093d845678.mp4

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeScreen.kt
```diff
@@ -19,26 +19,34 @@ import co.nimblehq.sample.compose.ui.screens.AppBar
...
+    val uiModels: List<UiModel> by viewModel.uiModels.collectAsState()
+    val firstTimeLaunch: Boolean by viewModel.firstTimeLaunch.collectAsState()
+
     val context = LocalContext.current
     LaunchedEffect(viewModel.error) {
```
thiennguyen0196, line 34: [Optional]
@lydiasama I know this is not your changes code 😅  but since you used `collectAsState`, I think it may be better if we can take advantage of the it as the same of others `showLoading` `uiModels` `firstTimeLaunch` in here with the `error` 👏 
```suggestion
    val error: Throwable? by viewModel.error.collectAsState(initial = null)
        LaunchedEffect(error) {
        error?.let {
            val message = it.userReadableMessage(context)
            Toast.makeText(context, message, Toast.LENGTH_SHORT).show()
        }
    }
```

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeScreen.kt
```diff
@@ -19,26 +19,34 @@ import co.nimblehq.sample.compose.ui.screens.AppBar
...
             Toast.makeText(context, message, Toast.LENGTH_SHORT).show()
         }
     }
+    LaunchedEffect(firstTimeLaunch) {
+        delay(1000)
+        if (firstTimeLaunch) {
+            Toast.makeText(context, "This is the first time launch", Toast.LENGTH_SHORT).show()
+        }
+    }
     LaunchedEffect(viewModel.navigator) {
```
thiennguyen0196, line 46: [Optional]
Ditto here with the usage of `collectAsState`
```suggestion
    val destination: AppDestination? by viewModel.navigator.collectAsState(initial = null)
    LaunchedEffect(destination) {
        destination?.let {
            navigator(it)
        }
    }
```

thiennguyen0196 (commented):
Finish my first round of review, there are just some optional recommendations in usage of `collectAsState` 😄

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeScreen.kt
lydiasama: @thiennguyen0196 I don't think the `error` stream is like others as it's more like the trigger function to show the toast message, not to update or recomposition any widget here. 🤔 And also if we change it to state, it will have only `null` or `unit`, which does not have any benefit IMO.

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeScreen.kt
lydiasama: The same answer to this one. 🙏🏻

lydiasama:
@thiennguyen0196 
> Not so sure about this delay here @lydiasama 🤔 When I tested it on my device without the delay(1000) it works fine and can show the Toast This is the first time launch normally 

Yes, I can remove the delay too. Just when I tested on my device: Samsung Note9, Android10, somehow it does not show the toast message properly. Sometimes it shows overlapping with others or flashes a bit. But I think I will remove it.

sample-compose/app/src/main/java/co/nimblehq/sample/compose/di/modules/PreferencesRepositoryModule.kt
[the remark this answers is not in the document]
lydiasama: Screenshot 2566-02-07 at 11 39 07 AM
Screenshot 2566-02-07 at 11 39 16 AM

@doannimble Could you please suggest me through the github suggestion? 🙏🏻 
I tried from my understanding, but it returns error from the AS.

lydiasama:
@doannimble @kaungkhantsoe @thiennguyen0196 @Wadeewee  @chornerman Thank you so much for your reviews. I updated all. 🙏🏻

sample-compose/app/src/main/java/co/nimblehq/sample/compose/di/modules/PreferencesRepositoryModule.kt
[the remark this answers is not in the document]
doannimble: @lydiasama Sorry my bad. I got confused between delegating and extending. The current approach is correct, we don't need to change anymore 🙏

doannimble (approved):
lgtm now 💯

luongvo:
@lydiasama is this part 1 or part 2? 😂 
image

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeViewState.kt
```diff
@@ -0,0 +1,10 @@
+package co.nimblehq.sample.compose.ui.screens.home
+
+import co.nimblehq.sample.compose.lib.IsLoading
+import co.nimblehq.sample.compose.model.UiModel
+
+data class HomeViewState(
```
luongvo, line 6: Do we need this `HomeViewState` anymore? 🤔

sample-compose/data/src/main/java/co/nimblehq/sample/compose/data/repository/AppPreferencesRepositoryImpl.kt
```diff
@@ -0,0 +1,44 @@
...
+class AppPreferencesRepositoryImpl @Inject constructor(
+    private val appPreferencesDataStore: DataStore<Preferences>
+) : AppPreferencesRepository {
+
+    private object PreferencesKeys {
+        val FIRST_TIME_LAUNCH = booleanPreferencesKey("FIRST_TIME_LAUNCH")
+    }
+
+    override fun isFirstTimeLaunchPreferences(): Flow<Boolean> = flow {
+        val firstLaunchPreferences = appPreferencesDataStore.data
```
luongvo, line 20: Naming it as `isFirstTimeLaunch` would be more consistent with `fun updateFirstTimeLaunchPreferences(isFirstTimeLaunch: Boolean)`

```suggestion
        val isFirstTimeLaunch = appPreferencesDataStore.data
```

sample-compose/data/src/main/java/co/nimblehq/sample/compose/data/repository/AppPreferencesRepositoryImpl.kt
```diff
@@ -0,0 +1,44 @@
...
+            .catch { exception ->
+                if (exception is IOException) {
+                    Log.e(
+                        AppPreferencesRepositoryImpl::class.simpleName,
+                        "Error reading preferences.",
+                        exception
+                    )
```
luongvo, line 27: I think we can get rid of this log and replace it with a comment like what Google suggested https://developer.android.com/codelabs/android-preferences-datastore#5. Logging with `e` level will throw a critical message in red in the log system.

If we still want to log, let's use a lower level such as `d` or `i`. 🙏

sample-compose/data/src/main/java/co/nimblehq/sample/compose/data/repository/AppPreferencesRepositoryImpl.kt
```diff
@@ -0,0 +1,44 @@
...
+                    emit(emptyPreferences())
+                } else {
+                    throw exception
+                }
+            }.map { preferences ->
+                preferences[PreferencesKeys.FIRST_TIME_LAUNCH] ?: true
+            }.first()
+
+        emit(firstLaunchPreferences)
+    }
```
luongvo, line 37: I'm not sure why do we need to wrap the result of `appPreferencesDataStore.data` inside a `flow {}` block again with `.first` to get the value since its result is a `Flow<Boolean>` already. Why don't we just pass the result to the main fun like this?

```suggestion
    override fun isFirstTimeLaunchPreferences(): Flow<Boolean> = appPreferencesDataStore.data
        .catch { exception ->
            if (exception is IOException) {
                Log.e(
                    AppPreferencesRepositoryImpl::class.simpleName,
                    "Error reading preferences.",
                    exception
                )
                emit(emptyPreferences())
            } else {
                throw exception
            }
        }.map { preferences ->
            preferences[FIRST_TIME_LAUNCH] ?: true
        }
```

sample-compose/domain/src/main/java/co/nimblehq/sample/compose/domain/repository/AppPreferencesRepository.kt
```diff
@@ -0,0 +1,11 @@
+package co.nimblehq.sample.compose.domain.repository
+
+import kotlinx.coroutines.flow.Flow
+
+interface AppPreferencesRepository {
+
+    fun isFirstTimeLaunchPreferences(): Flow<Boolean>
+
+    suspend fun updateFirstTimeLaunchPreferences(isFirstTimeLaunch: Boolean)
```
luongvo, line 9: I think we should not name these APIs with `Preferences` postfix as the wrapper is an `AppPreferences` already. This is the same when we define a Repository and all API functions inside not flowing with a `Repository` postfix.

Plus, `isFirstTimeLaunch` is not a `Preferences` object like the [Google example](https://developer.android.com/codelabs/android-preferences-datastore#5) because they have `UserPreferences` class as a wrapper.

```suggestion
    fun isFirstTimeLaunch(): Flow<Boolean>

    suspend fun updateFirstTimeLaunch(isFirstTimeLaunch: Boolean)
```

sample-compose/domain/src/main/java/co/nimblehq/sample/compose/domain/usecase/IsFirstTimeLaunchPreferencesUseCase.kt
```diff
@@ -0,0 +1,14 @@
+package co.nimblehq.sample.compose.domain.usecase
+
+import co.nimblehq.sample.compose.domain.repository.AppPreferencesRepository
+import kotlinx.coroutines.flow.Flow
+import javax.inject.Inject
+
+class IsFirstTimeLaunchPreferencesUseCase @Inject constructor(
```
luongvo, line 7: However, we can keep the `Preferences` postfix here to explicit we will get it from `Preferences` 🤓

sample-compose/domain/src/main/java/co/nimblehq/sample/compose/domain/usecase/IsFirstTimeLaunchPreferencesUseCase.kt
```diff
@@ -0,0 +1,14 @@
...
+    private val appPreferencesRepository: AppPreferencesRepository
+) {
+
+    operator fun invoke(): Flow<Boolean> {
+        return appPreferencesRepository.isFirstTimeLaunchPreferences()
+    }
+}
```
luongvo, line 14: Can we have unit tests for these new usecase? 👋

sample-compose/app/src/main/java/co/nimblehq/sample/compose/di/modules/PreferencesRepositoryModule.kt
```diff
@@ -0,0 +1,30 @@
...
+class PreferencesRepositoryModule {
+
+    @Provides
+    @Singleton
+    fun provideAppPreferencesDataStore(@ApplicationContext applicationContext: Context)
+        : DataStore<Preferences> = applicationContext.appPreferencesDataStore
+
+    @Provides
+    fun provideAppPreferencesRepository(appPreferencesDataStore: DataStore<Preferences>)
+        : AppPreferencesRepository = AppPreferencesRepositoryImpl(appPreferencesDataStore)
```
luongvo, line 29: Kinda weird formatting, should we break to the second line from `=` operator?

```suggestion
    fun provideAppPreferencesDataStore(@ApplicationContext applicationContext: Context): DataStore<Preferences> =
        applicationContext.appPreferencesDataStore

    @Provides
    fun provideAppPreferencesRepository(appPreferencesDataStore: DataStore<Preferences>): AppPreferencesRepository =
        AppPreferencesRepositoryImpl(appPreferencesDataStore)
```

luongvo (commented):
@lydiasama Please mention some documentation for references in this PR. That helps us reduce some questions and comments on the implementation without knowing why we need to do that. 🤗 
- https://developer.android.com/codelabs/android-preferences-datastore
- https://medium.com/androiddevelopers/all-about-preferences-datastore-cc7995679334

For example, this is how Google recommend us in error handling https://developer.android.com/codelabs/android-preferences-datastore#5 🤓 

Rest lgtm, good work 💪

sample-compose/data/src/main/java/co/nimblehq/sample/compose/data/repository/AppPreferencesRepositoryImpl.kt
lydiasama: @luongvo At first, I implemented as your suggestion. But I found that the receiver will never go into the collect() function and keeps waiting for the next value.

luongvo: @lydiasama That's the benefit of `Flow` lol to observe value changes in ViewModel, like we can observe value changes in the Room database with LivedData. Please look at the [first sample code of this question in stackoverflow](https://stackoverflow.com/questions/70567126/wait-for-preferences-datastore-to-retrieve-data-from-preferences-datastore-in-wo), they use it as an observer. So:
- We should not put Loading state with `isFirstTimeLaunchPreferencesUseCase()` execution.
- When ever new value comes, it will delegate the new value to the UI layer. If we don't want to keep the observer, use `first()`, but in VM layer instead.

```
        execute {
            showLoading()
            getModelsUseCase()
                .catch {
                    _error.emit(it)
                }
                .collect { result ->
                    val uiModels = result.toUiModels()
                    _uiModels.emit(uiModels)
                }
            hideLoading()
        }
        execute {
            // Use first() instead of collect() if we don't want to observe the value changes from DataStore  
            isFirstTimeLaunchPreferencesUseCase()
                .catch {
                    _error.emit(it)
                }
                .collect { isFirstTimeLaunch ->
                    _isFirstTimeLaunch.emit(isFirstTimeLaunch)
                    if (isFirstTimeLaunch) {
                        updateFirstTimeLaunchPreferencesUseCase(false)
                    }
                }
        }
```

lydiasama: @luongvo I see. Then I'll move `.first()` at the use case level. Thanks.

luongvo: @lydiasama I think we should use `first()` at ViewModel layer because `first()` will return the raw boolean value instead of Flow.

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeScreen.kt
thiennguyen0196: @lydiasama sorry for my late response as I was doing other tasks 😅 Since this is just optional and your explanation was quite clear enough, I think it is okay for me too 👍

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeScreen.kt
thiennguyen0196: Same as [this](https://github.com/nimblehq/android-templates/pull/389#discussion_r1099849312)

thiennguyen0196 (commented):
Apart from @luongvo 's comments, the rest LGTM, I will approve it when all the discussions are resolved. Really nice work there @lydiasama 👏

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeViewState.kt
lydiasama: @luongvo I'd like to get your opinion on this one too. 🙏🏻  Shall we combine all together in this `HomeViewState` or just do it separately like as-is.

luongvo: @lydiasama I think it's not related to what we need to achieve in this PR. Should we move it into a new discussion and a new task to refactor if needed?

I personally feel good with both, having `ViewState` is good too, but it's hard to manage screen states when having different data types, like `Survey` data loading state, `Header` data state, etc. This is kinda a big issue to discuss.

lydiasama: @luongvo  I agree. 🙌🏻

sample-compose/app/src/main/java/co/nimblehq/sample/compose/di/modules/PreferencesRepositoryModule.kt
```diff
@@ -0,0 +1,30 @@
...
+import dagger.hilt.InstallIn
+import dagger.hilt.android.qualifiers.ApplicationContext
+import dagger.hilt.components.SingletonComponent
+import javax.inject.Singleton
+
+private val Context.appPreferencesDataStore by preferencesDataStore("app_preferences")
+
+@Module
+@InstallIn(SingletonComponent::class)
+class PreferencesRepositoryModule {
```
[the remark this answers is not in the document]
lydiasama: > @lydiasama is this part 1 or part 2? 😂 I don't see part 2 but part 3.
> 
> image

@luongvo, @Tuubz requested me to change it to Part1 😄

sample-compose/data/src/main/java/co/nimblehq/sample/compose/data/repository/AppPreferencesRepositoryImpl.kt
```diff
@@ -0,0 +1,40 @@
...
+) : AppPreferencesRepository {
+
+    private object PreferencesKeys {
+        val FIRST_TIME_LAUNCH = booleanPreferencesKey("FIRST_TIME_LAUNCH")
+    }
+
+    override fun isFirstTimeLaunch(): Flow<Boolean> = appPreferencesDataStore.data
+        .catch { exception ->
+            if (exception is IOException) {
+                Log.i(
```
kaungkhantsoe, line 22: Can we remove this or use `Timber` here?

kaungkhantsoe (approved):
LGTM 🚀

sample-compose/data/src/main/java/co/nimblehq/sample/compose/data/repository/AppPreferencesRepositoryImpl.kt
thiennguyen0196: I believe we do not include `Timber` in the gradle of `data` layer @kaungkhantsoe 😄 
https://github.com/nimblehq/android-templates/blob/68d7608b5c2113a7450650a3a2a6b0cf77b6bf58/sample-compose/data/build.gradle.kts#L58-L83

lydiasama:
@luongvo @thiennguyen0196 Resolved all. Plus, unit test added.

thiennguyen0196 (approved):
LGTM now

luongvo (commented):
rest lgtm, 1 last issue 👋

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeScreen.kt
```diff
@@ -25,25 +25,31 @@ fun HomeScreen(
...
 
-    val showLoading: IsLoading by viewModel.showLoading.collectAsState()
-    val uiModels: List<UiModel> by viewModel.uiModels.collectAsState()
-
     CameraPermission()
 
     HomeScreenContent(
         uiModels = uiModels,
-        showLoading = showLoading,
+        showLoading = isLoading,
```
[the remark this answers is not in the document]
lydiasama: @longnd Indeed!. Nice catch. 👍🏻

luongvo (approved):
LGTM 🎉

sample-compose/domain/src/main/java/co/nimblehq/sample/compose/domain/repository/AppPreferencesRepository.kt
```diff
@@ -0,0 +1,11 @@
...
+
+import kotlinx.coroutines.flow.Flow
+
+interface AppPreferencesRepository {
+
+    fun isFirstTimeLaunch(): Flow<Boolean>
+
+    suspend fun updateFirstTimeLaunch(isFirstTimeLaunch: Boolean)
+}
+
```
chornerman, line 11: Redundant blank line 😄

chornerman (commented):
Minor comment 🚀

Wadeewee (approved):

chornerman (approved):
LGTM 👏🏼 👏🏼

sample-compose/data/src/main/java/co/nimblehq/sample/compose/data/repository/AppPreferencesRepositoryImpl.kt
```diff
@@ -0,0 +1,40 @@
...
+
+import android.util.Log
+import androidx.datastore.core.DataStore
+import androidx.datastore.preferences.core.*
+import co.nimblehq.sample.compose.domain.repository.AppPreferencesRepository
+import kotlinx.coroutines.flow.*
+import java.io.IOException
+import javax.inject.Inject
+
+class AppPreferencesRepositoryImpl @Inject constructor(
```
luongvo, line 11: @lydiasama Wait, I think we should not add `androidx.datastore:datastore-preferences` to the data module. Is it better to only add it to the `app` module and keep this AppPreferencesRepository implementation in the app module as well at `co.nimblehq.sample.compose.repository`?

This is the same when we need to integrate Firebase features such as FirebaseMessaging or FirebaseRemoteConfig. We need to put repository's implementations in the app module as well 💭

sample-compose/data/build.gradle.kts
```diff
@@ -58,6 +58,7 @@ dependencies {
     implementation(project(Module.DOMAIN))
 
     implementation("androidx.core:core-ktx:${Versions.ANDROIDX_CORE_KTX_VERSION}")
+    implementation("androidx.datastore:datastore-preferences:${Versions.ANDROIDX_DATASTORE_PREFERENCES_VERSION}")
```
luongvo, line 61: Hence we don't need this dependency.

luongvo (changes requested):

sample-compose/data/src/main/java/co/nimblehq/sample/compose/data/repository/AppPreferencesRepositoryImpl.kt
lydiasama: @luongvo I considered the `Datastore` as a resource. So I think it makes sense to be in the data module. What do you think if we discuss about this in the discussion?

lydiasama: If we consider in a standard Android project, it is common to place SharedPreferences in the app module, as it is often closely tied to the application's lifecycle and functions as a convenient storage location for user preferences and settings. This approach is straightforward and simple, as the SharedPreferences can be accessed directly from the Context of the application.

However, in a more modular or architectural approach, such as MVVM or Clean Architecture, it may be preferable to place the SharedPreferences in the data layer of the application. This approach allows for a clearer separation of concerns, as the data layer is responsible for managing and persisting data, including user preferences. Additionally, this makes it easier to test the data layer and provides a clear API for accessing and modifying SharedPreferences from the rest of the application. @luongvo

luongvo: @lydiasama Ok, so if we decide to put all data-related stuff in the `data` module, could we get rid of `androidx.datastore:datastore-preferences` in the app module?

lydiasama: @luongvo We can't because for now, our project structure set the di inside of the `app` module. So we need to keep the dependency for it.

Screenshot 2566-02-10 at 3 57 48 PM

lydiasama: @luongvo You can see the same structure with `RepositoryModule` that we call API. We implemented the `RepositoryModule` inside of `app` module, but `RepositoryImpl` is in the `data` module as well.

luongvo: @lydiasama No, it's not the same because the app module does not include retrofit, hence the app module should not include DataStore dependency too 💭

lydiasama: @nimblehq/android-chapter What do you guys think?

lydiasama: @luongvo I created a discussion [here](https://github.com/nimblehq/android-templates/discussions/406)

luongvo: @lydiasama Since we had a new discussion on this issue. I think we can handle this issue in part 2 or part 3 if needed https://github.com/nimblehq/android-templates/pull/374. 👌

luongvo (approved):
lgtm

kaungkhantsoe:
@lydiasama Conflict 🦀

lydiasama:
@kaungkhantsoe Thanks ❤️  Resolved!

ryan-conway (approved):

PR #372, opened 2022-12-13
[#352] [Template] 2/2: Setup the nav graph and navigation logic

luongvo:
Solves https://github.com/nimblehq/android-templates/issues/352

## What happened 👀

Bring all necessary changes from Compose Sample project #367 to the Compose Template.

## Insight 📝

- There is no Second screen at the template level.

## Proof Of Work 📹

All the logic should be the same as what we did in Compose Sample project ✅

PR #390, opened 2023-01-16
[#377] [Compose] Add a sample to navigate to the ThirdScreen with a custom parcelable data

### LINKED ISSUE #377: [Compose] Add a sample to navigate to the ThirdScreen with a custom parcelable data
## Why

Coming through [this discussion](https://github.com/nimblehq/android-templates/pull/372#discussion_r1058045844), we need to have a sample implementation to demonstrate how can we pass a parcelable data between destinations. Unfortunately, it's not that easy:
- https://betterprogramming.pub/realize-jetpack-compose-navigation-2889401f52b
- https://stackoverflow.com/questions/65610003/pass-parcelable-argument-with-compose-navigation
- https://stackoverflow.com/questions/67553012/passing-model-between-destinations/67553630#67553630
- https://gist.github.com/Aidanvii7/46f9f014ba4dc58a2ac10625b198769d

We should find an optimized way to apply it to the template.

## Who Benefits?

Developers

kaungkhantsoe:
- Closes #377 

## What happened 👀

Coming through [this discussion](https://github.com/nimblehq/android-templates/pull/372#discussion_r1058045844), we need to have a sample implementation to demonstrate how can we pass a parcelable data between destinations. Unfortunately, it's not that easy:
- https://betterprogramming.pub/realize-jetpack-compose-navigation-2889401f52b
- https://stackoverflow.com/questions/65610003/pass-parcelable-argument-with-compose-navigation
- https://stackoverflow.com/questions/67553012/passing-model-between-destinations/67553630#67553630
- https://gist.github.com/Aidanvii7/46f9f014ba4dc58a2ac10625b198769d

We should find an optimized way to apply it to the template.

## Insight 📝

- Tried to implement the way the links provided and I found [this issue](https://github.com/nimblehq/android-templates/issues/377#issuecomment-1379951504)
- Big thanks to Huey's [suggestion](https://github.com/nimblehq/android-templates/issues/377#issuecomment-1381417694), I have successfully implemented to pass a `Parcelable data` between screens 🥳 
- Implemented a `Long Press` behavior and a `DropdownMenu` to be able to go to `Third Screen`
- Used `savedStateHandle` to pass `Parcelable data` between screens
- Updated Compose Navigation Version to `2.5.3` to be able to prove that the current implementation works with the latest version

## Proof Of Work 📹

https://user-images.githubusercontent.com/32578035/212663394-cf53a958-e23e-4960-aa4b-3432726f77c6.mov

### DIFF
--- a/sample-compose/app/src/main/java/co/nimblehq/sample/compose/model/UiModel.kt
+++ b/sample-compose/app/src/main/java/co/nimblehq/sample/compose/model/UiModel.kt
@@ -6,8 +6,7 @@ import kotlinx.parcelize.Parcelize
 
 @Parcelize
 data class UiModel(
-    val id: String = "",
-    val url: String = "https://www.google.com/"
+    val id: String = ""
 ) : Parcelable
 
 private fun Model.toUiModel() = UiModel(id.toString())
--- a/sample-compose/app/src/main/java/co/nimblehq/sample/compose/model/UiModel.kt
+++ b/sample-compose/app/src/main/java/co/nimblehq/sample/compose/model/UiModel.kt
@@ -6,7 +6,8 @@ import kotlinx.parcelize.Parcelize
 
 @Parcelize
 data class UiModel(
-    val id: String
+    val id: String = "",
+    val url: String = "https://www.google.com/"
 ) : Parcelable
 
 private fun Model.toUiModel() = UiModel(id.toString())
--- a/sample-compose/app/src/main/java/co/nimblehq/sample/compose/model/UiModel.kt
+++ b/sample-compose/app/src/main/java/co/nimblehq/sample/compose/model/UiModel.kt
@@ -6,7 +6,7 @@ import kotlinx.parcelize.Parcelize
 
 @Parcelize
 data class UiModel(
-    val id: String = ""
+    val id: String
 ) : Parcelable
 
 private fun Model.toUiModel() = UiModel(id.toString())
--- a/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppDestination.kt
+++ b/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppDestination.kt
@@ -12,7 +12,7 @@ sealed class AppDestination(val route: String = "") {
 
     open var destination: String = route
 
-    open var parcel: Pair<String, Any?> = "" to null
+    open var parcelableArgument: Pair<String, Any?> = "" to null
 
     object Up : AppDestination()
 
@@ -31,7 +31,7 @@ sealed class AppDestination(val route: String = "") {
 
     object Third : AppDestination("third") {
         fun addParcel(value: UiModel) = apply {
-            parcel = KeyModel to value
+            parcelableArgument = KeyModel to value
         }
     }
 }
--- a/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppDestination.kt
+++ b/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppDestination.kt
@@ -1,14 +1,18 @@
 package co.nimblehq.sample.compose.ui
 
 import androidx.navigation.*
+import co.nimblehq.sample.compose.model.UiModel
 
 const val KeyId = "id"
+const val ModelKey = "model"
 
 sealed class AppDestination(val route: String = "") {
 
     open val arguments: List<NamedNavArgument> = emptyList()
 
     open var destination: String = route
+
+    open var parcel : Pair<String,UiModel?> = "" to null
 
     object Up : AppDestination()
 
@@ -20,8 +24,14 @@ sealed class AppDestination(val route: String = "") {
             navArgument(KeyId) { type = NavType.StringType }
         )
 
-        fun buildDestination(id: String) = apply {
+        fun createRoute(id: String) = apply {
             destination = "second/$id"
+        }
+    }
+
+    object Third: AppDestination("third") {
+        fun addParcel(value: UiModel) = apply {
+            parcel = ModelKey to value
         }
     }
 }
--- a/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppDestination.kt
+++ b/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppDestination.kt
@@ -4,7 +4,7 @@ import androidx.navigation.*
 import co.nimblehq.sample.compose.model.UiModel
 
 const val KeyId = "id"
-const val ModelKey = "model"
+const val KeyModel = "model"
 
 sealed class AppDestination(val route: String = "") {
 
@@ -12,7 +12,7 @@ sealed class AppDestination(val route: String = "") {
 
     open var destination: String = route
 
-    open var parcel : Pair<String,UiModel?> = "" to null
+    open var parcel: Pair<String, Any?> = "" to null
 
     object Up : AppDestination()
 
@@ -29,9 +29,9 @@ sealed class AppDestination(val route: String = "") {
         }
     }
 
-    object Third: AppDestination("third") {
+    object Third : AppDestination("third") {
         fun addParcel(value: UiModel) = apply {
-            parcel = ModelKey to value
+            parcel = KeyModel to value
         }
     }
 }
--- a/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppNavigation.kt
+++ b/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppNavigation.kt
@@ -20,7 +20,6 @@ fun AppNavigation(
         composable(destination = AppDestination.Home) {
             HomeScreen(
                 navigator = { destination ->
-//                    navController.navigate(destination)
                     navController.navigate(destination, destination.parcel)
                 }
             )
@@ -36,7 +35,7 @@ fun AppNavigation(
         composable(destination = AppDestination.Third) {
             ThirdScreen(
                 navigator = { destination -> navController.navigate(destination) },
-                model = navController.previousBackStackEntry?.savedStateHandle?.get<UiModel>(ModelKey) ?: UiModel()
+                model = navController.previousBackStackEntry?.savedStateHandle?.get<UiModel>(KeyModel) ?: UiModel()
             )
         }
     }
@@ -66,7 +65,7 @@ private fun <T> NavHostController.navigate(appDestination: AppDestination, parce
     when (appDestination) {
         is AppDestination.Up -> navigateUp()
         else -> {
-            currentBackStackEntry?.savedStateHandle?.set(parcel.first,parcel.second)
+            currentBackStackEntry?.savedStateHandle?.set(parcel.first, parcel.second)
             navigate(route = appDestination.destination)
         }
     }
--- a/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppNavigation.kt
+++ b/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppNavigation.kt
@@ -20,7 +20,7 @@ fun AppNavigation(
         composable(destination = AppDestination.Home) {
             HomeScreen(
                 navigator = { destination ->
-                    navController.navigate(destination, destination.parcel)
+                    navController.navigate(destination, destination.parcelableArgument)
                 }
             )
         }
@@ -35,7 +35,7 @@ fun AppNavigation(
         composable(destination = AppDestination.Third) {
             ThirdScreen(
                 navigator = { destination -> navController.navigate(destination) },
-                model = navController.previousBackStackEntry?.savedStateHandle?.get<UiModel>(KeyModel) ?: UiModel()
+                model = navController.previousBackStackEntry?.savedStateHandle?.get<UiModel>(KeyModel)
             )
         }
     }
@@ -55,27 +55,19 @@ private fun NavGraphBuilder.composable(
 }
 
 /**
- * Navigate to [AppDestination] with navArgument
- */
-private fun NavHostController.navigate(appDestination: AppDestination) {
-    when (appDestination) {
-        is AppDestination.Up -> navigateUp()
-        else -> navigate(route = appDestination.destination)
-    }
-}
-
-/**
- * Navigate to AppDestination with Parcelable Data
+ * Navigate to provided [AppDestination] with a Pair of key value String and Data [parcel]
  * Caution to use this method. This method use savedStateHandle to store the Parcelable data.
  * When previousBackstackEntry is popped out from navigation stack, savedStateHandle will return null and cannot retrieve data.
  * eg.Login -> Home, the Login screen will be popped from the back-stack on logging in successfully.
- * Navigate to provided [AppDestination] with a Pair of key value String and Data [parcel]
  */
-private fun <T> NavHostController.navigate(appDestination: AppDestination, parcel: Pair<String, T>) {
+private fun NavHostController.navigate(appDestination: AppDestination, parcel: Pair<String, Any?>? = null) {
     when (appDestination) {
         is AppDestination.Up -> navigateUp()
         else -> {
-            currentBackStackEntry?.savedStateHandle?.set(parcel.first, parcel.second)
+            parcel?.let {
+                val (key, value) = it
+                currentBackStackEntry?.savedStateHandle?.set(key, value)
+            }
             navigate(route = appDestination.destination)
         }
     }
--- a/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppNavigation.kt
+++ b/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppNavigation.kt
@@ -64,8 +64,7 @@ private fun NavHostController.navigate(appDestination: AppDestination, parcel: P
     when (appDestination) {
         is AppDestination.Up -> navigateUp()
         else -> {
-            parcel?.let {
-                val (key, value) = it
+            parcel?.let { (key, value) ->
                 currentBackStackEntry?.savedStateHandle?.set(key, value)
             }
             navigate(route = appDestination.destination)
--- a/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppNavigation.kt
+++ b/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppNavigation.kt
@@ -3,8 +3,10 @@ package co.nimblehq.sample.compose.ui
 import androidx.compose.runtime.Composable
 import androidx.navigation.*
 import androidx.navigation.compose.*
+import co.nimblehq.sample.compose.model.UiModel
 import co.nimblehq.sample.compose.ui.screens.home.HomeScreen
 import co.nimblehq.sample.compose.ui.screens.second.SecondScreen
+import co.nimblehq.sample.compose.ui.screens.third.ThirdScreen
 
 @Composable
 fun AppNavigation(
@@ -15,16 +17,26 @@ fun AppNavigation(
         navController = navController,
         startDestination = startDestination
     ) {
-        composable(AppDestination.Home) {
+        composable(destination = AppDestination.Home) {
             HomeScreen(
-                navigator = { destination -> navController.navigate(destination) }
+                navigator = { destination ->
+//                    navController.navigate(destination)
+                    navController.navigate(destination, destination.parcel)
+                }
             )
         }
 
-        composable(AppDestination.Second) { backStackEntry ->
+        composable(destination = AppDestination.Second) { backStackEntry ->
             SecondScreen(
                 navigator = { destination -> navController.navigate(destination) },
                 id = backStackEntry.arguments?.getString(KeyId).orEmpty()
+            )
+        }
+
+        composable(destination = AppDestination.Third) {
+            ThirdScreen(
+                navigator = { destination -> navController.navigate(destination) },
+                model = navController.previousBackStackEntry?.savedStateHandle?.get<UiModel>(ModelKey) ?: UiModel()
             )
         }
     }
@@ -47,5 +59,15 @@ private fun NavHostController.navigate(appDestination: AppDestination) {
     when (appDestination) {
         is AppDestination.Up -> navigateUp()
         else -> navigate(route = appDestination.destination)
+    }
+}
+
+private fun <T> NavHostController.navigate(appDestination: AppDestination, parcel: Pair<String, T>) {
+    when (appDestination) {
+        is AppDestination.Up -> navigateUp()
+        else -> {
+            currentBackStackEntry?.savedStateHandle?.set(parcel.first,parcel.second)
+            navigate(route = appDestination.destination)
+        }
     }
 }
--- a/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppNavigation.kt
+++ b/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppNavigation.kt
@@ -54,6 +54,9 @@ private fun NavGraphBuilder.composable(
     )
 }
 
+/**
+ * Navigate to [AppDestination] with navArgument
+ */
 private fun NavHostController.navigate(appDestination: AppDestination) {
     when (appDestination) {
         is AppDestination.Up -> navigateUp()
@@ -61,6 +64,13 @@ private fun NavHostController.navigate(appDestination: AppDestination) {
     }
 }
 
+/**
+ * Navigate to AppDestination with Parcelable Data
+ * Caution to use this method. This method use savedStateHandle to store the Parcelable data.
+ * When previousBackstackEntry is popped out from navigation stack, savedStateHandle will return null and cannot retrieve data.
+ * eg.Login -> Home, the Login screen will be popped from the back-stack on logging in successfully.
+ * Navigate to provided [AppDestination] with a Pair of key value String and Data [parcel]
+ */
 private fun <T> NavHostController.navigate(appDestination: AppDestination, parcel: Pair<String, T>) {
     when (appDestination) {
         is AppDestination.Up -> navigateUp()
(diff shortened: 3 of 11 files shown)


sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppDestination.kt
```diff
@@ -1,15 +1,19 @@
 package co.nimblehq.sample.compose.ui
 
 import androidx.navigation.*
+import co.nimblehq.sample.compose.model.UiModel
 
 const val KeyId = "id"
+const val ModelKey = "model"
```
hoangnguyen92dn, line 7: Should we put the `Key` as a prefix instead? At least it is consistent with `KeyId`
```suggestion
const val KeyModel = "model"
```

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppDestination.kt
```diff
@@ -1,15 +1,19 @@
...
 
 sealed class AppDestination(val route: String = "") {
 
     open val arguments: List<NamedNavArgument> = emptyList()
 
     open var destination: String = route
 
+    open var parcel : Pair<String,UiModel?> = "" to null
```
hoangnguyen92dn, line 15: I would recommend having more generic type to support different UiModel. Any please reformat btw 😄 

```suggestion
    open var parcel: Pair<String, Any?> = "" to null
```

I am also thinking about refactoring the `arguments`, seems we don't need to use it anymore, and confusing to have it here. cc @luongvo

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppNavigation.kt
```diff
@@ -15,18 +17,28 @@ fun AppNavigation(
         navController = navController,
         startDestination = startDestination
     ) {
-        composable(AppDestination.Home) {
+        composable(destination = AppDestination.Home) {
             HomeScreen(
-                navigator = { destination -> navController.navigate(destination) }
+                navigator = { destination ->
+//                    navController.navigate(destination)
```
hoangnguyen92dn, line 23: Could we remove this?

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeScreen.kt
```diff
@@ -44,7 +44,8 @@ fun HomeScreen(
     HomeScreenContent(
         uiModels = uiModels,
         showLoading = showLoading,
-        onItemClick = viewModel::navigateToSecond
+        onItemClick = viewModel::navigateToSecond,
+        onItemEdit = viewModel::navigateToThird
```
hoangnguyen92dn, line 48: I suggest renaming the callback a bit for consistency and to make more sense. We could have the callback with the name of the event used to trigger correctly; furthermore, when we change the handler, we don't need to reflect the name, for example we change the long click event to create a new item or remove it, in that case we have to change `onItemEdit` -> `onItemRemove` ...

```suggestion
        onItemLongClick = viewModel::navigateToThird
```

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppDestination.kt
kaungkhantsoe: We need it if we use `navArgument` instead of `savedStateHandle` to pass the data. @hoangnguyen92dn

hoangnguyen92dn: @kaungkhantsoe I see, I mean, could we change to use only one approach for passing data? It would avoid confusion which one they could use case by case from the team.

kaungkhantsoe: IMO, I would like to leave them as options since using `navArgument` is recommended by the Android Doc. We can also use it if we don't have to pass an object. 💭

chornerman (approved):
Rest LGTM 🚀

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppDestination.kt
hoangnguyen92dn: @kaungkhantsoe If so, could we add the comment to explain when we could use `argument` and `parcel` to clarify the newcomer's confusion? cc @luongvo @Tuubz Please share your though since we will apply this approach for our template. I am thinking about the issue that when `previousBackstackEntry` could be popped out, so we could not receive the data on the destination. Ex: Login -> Home, the Login screen will be popped from the back-stack on logging in successfully.

And I think it better if we reorder the variables a bit: 

```
    open var destination: String = route

    open val arguments: List<NamedNavArgument> = emptyList()

    open var parcel : Pair<String,UiModel?> = "" to null
```

doannimble (approved):
LGTM 🚀

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppDestination.kt
kaungkhantsoe: @hoangnguyen92dn You mean like this? Is it good enough? 
Screenshot 2023-01-31 at 12 42 11 PM

sample-compose/app/src/main/java/co/nimblehq/sample/compose/model/UiModel.kt
```diff
@@ -6,7 +6,8 @@ import kotlinx.parcelize.Parcelize
 
 @Parcelize
 data class UiModel(
-    val id: String
+    val id: String = "",
+    val url: String = "https://www.google.com/"
```
Wadeewee, line 10: May I know any purpose for adding `url` in the model 🤔  I didn't see where we will use it only show in `ThirdScreen`, Can we have only `id`? 🙏 
```suggestion

```

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppDestination.kt
hoangnguyen92dn: @kaungkhantsoe Looks good to me. One minor comment about the `@param`, I think we can get rid of it. Refer: https://kotlinlang.org/docs/coding-conventions.html#documentation-comments

sample-compose/app/src/main/java/co/nimblehq/sample/compose/model/UiModel.kt
kaungkhantsoe: @Wadeewee It is to ensure we can navigate with a Parcelable data which include `url`. There are 2 ways to navigate with parcelable data. Using `savedInstanceState` and `navArgument`. We can serialize the whole model and use it inside `navArgument` too. In that case, we need to encode `url` before serializing the object. Because the navigation uses `nav route` and it somehow assume the `url` as part of the `nav route`.

Wadeewee: > It is to ensure that we can navigate with a Parcelable Object which include url

@kaungkhantsoe, Not sure If I get your point, but removing `url` it's still a `Parcelable Object` 🤔 

Here's my result when remove `url`:

https://user-images.githubusercontent.com/28002315/215727331-80ccea3f-3011-4d59-94ba-b55b354de3a0.mov

Please correct me If I missed something 🙏  cc @hoangnguyen92dn @Tuubz

kaungkhantsoe: @Wadeewee Yes, it is working since we are using `savedInstanceState` for now. There is another way to navigate without using `savedInstanceState`. We haven't implemented it yet. I can remove it for now. I just want us to be aware of the danger of `url` here.

kaungkhantsoe: `savedInstanceState` has a drawback as describe in [this comment](https://github.com/nimblehq/android-templates/pull/390#discussion_r1090170856).

kaungkhantsoe: Got it. Since `url` has that kind of problem with navigation, we cannot be sure whether `savedInstanceState` might or might not have a problem later. I will remove it for now. cc @Tuubz @Wadeewee

hoangnguyen92dn: > In that case, we need to encode url before serializing the object. Because the navigation uses nav route and it somehow assume the url as part of the nav route.

@kaungkhantsoe Hold onnnn... With the `Third` App destination, we don't add any extra information to the nav route, it's simply `"third"` like `"home"`, the parcel object will be retained in the `savedInstanceState`, it doesn't related to the nav route.

hoangnguyen92dn: Btw, I am fine with the object containing `url` or not, the idea is passing the data object to destination 😄 @kaungkhantsoe @Wadeewee

kaungkhantsoe: You can check out [this thread](https://stackoverflow.com/a/67133534) and its comments why we could have a problem with `url`. We haven't implemented the `serialization` way for now because we are worry that it might block the thread if we serialize a big object. Unlike using `savedInstanceState`, `serialization` way doesn't have a drawback as described in [this comment](https://github.com/nimblehq/android-templates/pull/390#discussion_r1090170856)  cc @hoangnguyen92dn @Wadeewee

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/Item.kt
```diff
@@ -1,29 +1,48 @@
...
             .fillMaxWidth()
-            .clickable(onClick = { onClick(uiModel) })
+            .combinedClickable(
+                onClick = { onClick(uiModel) },
+                onLongClick = { expanded = !expanded }
+            )
     ) {
         Text(
-            modifier = Modifier.padding(dimensions.spacingNormal),
+            modifier = Modifier
```
Wadeewee, line 33: @hoangnguyen92dn @kaungkhantsoe @Tuubz May I get your opinion on this 🙏 

What do you guys think if we provide `modifier` from the root/parent(**HomeScreen**) of `Composable` that we want to use this `Composable`(**Item**)? 🤔 

If I'm not wrong, it might affect when `re-composition`  in the case when we have another screen that also uses the `Item` function or when we have a child function that's separate from the parent.

```suggestion
            modifier = modifier
```

lydiasama (approved):

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppNavigation.kt
```diff
@@ -15,18 +17,27 @@ fun AppNavigation(
         navController = navController,
         startDestination = startDestination
     ) {
-        composable(AppDestination.Home) {
+        composable(destination = AppDestination.Home) {
             HomeScreen(
-                navigator = { destination -> navController.navigate(destination) }
+                navigator = { destination ->
+                    navController.navigate(destination, destination.parcel)
```
Wadeewee, line 23: Question 🙋‍♀️   @kaungkhantsoe @hoangnguyen92dn,  I found something interesting in this [link](https://gist.github.com/Aidanvii7/46f9f014ba4dc58a2ac10625b198769d#file-navcontrollercomposeutils-kt) It seems like from the example they created a `NavController` extension. Not sure you have tried to follow this. I feel like It will be flexible when we have to add/customize `NavOptions` or `Navigator.Extras` as well, but it might require more refactored. 🙈 

Any insight for this? 🙏  cc @luongvo @Tuubz

hoangnguyen92dn: @Wadeewee I found the approach you mentioned before, and it looks promising with extensions and implementation. But until now, it's out of date with that approach in here:

https://gist.github.com/Aidanvii7/46f9f014ba4dc58a2ac10625b198769d#file-navcontrollercomposeutils-kt-L23
```
    val argumentBundle: Bundle = addedEntry.arguments ?: Bundle().also {
        addedEntry.arguments = it // 👈 This doesn't support var to reassign the argument value
    }
```

And @kaungkhantsoe also mentioned that on the [discussion](https://github.com/nimblehq/android-templates/issues/377#issuecomment-1379951504) 😄 

image

Wadeewee: Ah, I see thanks for your explanation @hoangnguyen92dn 🙇‍♀️

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppNavigation.kt
```diff
@@ -43,9 +54,29 @@ private fun NavGraphBuilder.composable(
...
     }
 }
+
+/**
+ * Navigate to AppDestination with Parcelable Data
+ * Caution to use this method. This method use savedStateHandle to store the Parcelable data.
+ * When previousBackstackEntry is popped out from navigation stack, savedStateHandle will return null and cannot retrieve data.
+ * eg.Login -> Home, the Login screen will be popped from the back-stack on logging in successfully.
+ * Navigate to provided [AppDestination] with a Pair of key value String and Data [parcel]
+ */
```
Tuubz, line 73: Seems like they are implying the same message 🙏 
```suggestion
/**
 * Navigate to provided [AppDestination] with a Pair of key value String and Data [parcel]
 * Caution to use this method. This method use savedStateHandle to store the Parcelable data.
 * When previousBackstackEntry is popped out from navigation stack, savedStateHandle will return null and cannot retrieve data.
 * eg.Login -> Home, the Login screen will be popped from the back-stack on logging in successfully.
 */
```

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppNavigation.kt
```diff
@@ -15,18 +17,27 @@ fun AppNavigation(
...
             SecondScreen(
                 navigator = { destination -> navController.navigate(destination) },
                 id = backStackEntry.arguments?.getString(KeyId).orEmpty()
             )
         }
+
+        composable(destination = AppDestination.Third) {
+            ThirdScreen(
+                navigator = { destination -> navController.navigate(destination) },
+                model = navController.previousBackStackEntry?.savedStateHandle?.get<UiModel>(KeyModel) ?: UiModel()
```
Wadeewee, line 38: @kaungkhantsoe I think we can allow `nullable` for `parcel`. What If we have many variables in `UiModel`? That means we need to set the default value in each variable. 🤔 

```suggestion
                model = navController.previousBackStackEntry?.savedStateHandle?.get<UiModel>(KeyModel)
```

So, by this way we no need to set the default value for `id` in `UiModel` cc @hoangnguyen92dn @Tuubz

Wadeewee: And one more thing, @kaungkhantsoe, I think we should create a new object `ModelParcel` and use this object for passing arguments instead of `UiModel`, What do you think @hoangnguyen92dn @Tuubz

sample-compose/app/src/main/res/values/strings.xml
```diff
@@ -7,4 +7,8 @@
 
     <string name="second_title_bar">Second</string>
     <string name="second_id_title">ID: %1$s</string>
+
+    <string name="third_title_bar">Third</string>
+    <string name="third_data_title">Data: %1$s</string>
```
Tuubz, line 12: Simplified

```suggestion
    <string name="third_data_title">Data: %s</string>
```

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/Item.kt
```diff
@@ -1,29 +1,48 @@
...
-    Row(
+    var expanded by remember { mutableStateOf(false) }
+
+    Box(
         modifier = Modifier
             .fillMaxWidth()
-            .clickable(onClick = { onClick(uiModel) })
+            .combinedClickable(
+                onClick = { onClick(uiModel) },
+                onLongClick = { expanded = !expanded }
```
Tuubz, line 29: Optional, might be easier to read

```suggestion
                onLongClick = { expanded = true }
```

Tuubz (commented):
@kaungkhantsoe Sorry for the late review, if possible let's try to get this merged today. 🙇

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/third/ThirdScreen.kt
```diff
@@ -0,0 +1,58 @@
...
+            modifier = Modifier
+                .fillMaxSize()
+                .padding(paddingValues)
+        ) {
+            Text(
+                text = stringResource(R.string.third_data_title, data.toString()),
+                textAlign = TextAlign.Center,
+                modifier = Modifier
+                    .align(Alignment.Center)
+                    .padding(10.dp)
```
Wadeewee, line 46: May I know why we need to set padding here 🙏 
```suggestion

```

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppNavigation.kt
hoangnguyen92dn: @Wadeewee Do you mean a wrapper class, likely `Bundle`, `Extra` to pack the real data object inside? It would me cumbersome since we have to create a new wrapper and in the destination one more step to get it out from the wrapper. I don't think we would need it 😄 Generic type is more convenient.

Tuubz: +1 for allowing `nullable` for `parcel`, it will be easier to know if something went wrong on the next screen if so 🙏

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/Item.kt
Tuubz: @Wadeewee Are you suggesting we should pass `modifier` through the constructor? 💭 

Sorry for asking for clarification, I'm trying to better understand your suggestion 🙇 

```
fun Item(
    uiModel: UiModel,
    onClick: (UiModel) -> Unit,
    onLongClick: (UiModel) -> Unit,
    modifier: Modifier // this is new
)
```

kaungkhantsoe: I am not sure I understand you @Wadeewee. Can you elaborate more on why `modifier` would affect recomposition?

Wadeewee: > Are you suggesting we should pass modifier through the constructor?

Yes, @Tuubz, and Sorry 🙏 If I write the message not clearly.

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppNavigation.kt
Wadeewee: > Do you mean a wrapper class, likely Bundle, Extra to pack the real data object inside? 

Oh, @hoangnguyen92dn, I didn't mean like this ☝️ and What's I mean 👇  :

```kotlin
// ---- Use for UI Leyer ---- //
data class UiModel(
    val id: String
)

private fun Model.toUiModel() = UiModel(id.toString())

fun List<Model>.toUiModels() = this.map { it.toUiModel() }
```

```kotlin
// ---- Use for passing argument---- //
@Parcelize
data class ModelParcel(
    val id: String
) : Parcelable

private fun UiModel.toModelParcel() = ModelParcel(id = id)
```

```kotlin
 composable(destination = AppDestination.Third) {
            ThirdScreen(
                navigator = { destination -> navController.navigate(destination) },
                model = navController.previousBackStackEntry?.savedStateHandle?.get<ModelParcel>(KeyModel)
            )
        }
```
Actually, we don't use a kind of `UiModel` object to passing `argument` to another screen that's why I suggest to create a new object with suffix `Parcel` cc @Tuubz

kaungkhantsoe: @Wadeewee So it is the same object but with a different suffix name and parcelized?

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/Item.kt
Tuubz: @Wadeewee If we do that, then it means all the components in this screen will make use of the same `modifier` 

Sounds risky and difficult to manage 💭

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppNavigation.kt
hoangnguyen92dn: @Wadeewee Gotcha 👍 It's good to separate the UiModel and Data Model for passing to the destination. As I mentioned, the sample is supposed to demo how we pass the data between screens, creating a new `ModelParcel` could share unnecessary boilerplate code IMO. cc @Tuubz

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/Item.kt
kaungkhantsoe: We can pass the modifier easily but I think it depends on how you will use it. I don't think `modifier` can affect `recomposition` in this case. Currently, it is just a simple UI sample so I suggest we should leave it as it is. 🙏 cc @Wadeewee @Tuubz

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppNavigation.kt
```diff
@@ -43,9 +54,21 @@ private fun NavGraphBuilder.composable(
...
+ */
+private fun NavHostController.navigate(appDestination: AppDestination, parcel: Pair<String, Any?>? = null) {
     when (appDestination) {
         is AppDestination.Up -> navigateUp()
-        else -> navigate(route = appDestination.destination)
+        else -> {
+            parcel?.let {
+                val (key, value) = it
+                currentBackStackEntry?.savedStateHandle?.set(key, value)
+            }
```
Tuubz, line 70:
```suggestion
            parcel?.let { (key, value) -> currentBackStackEntry?.savedStateHandle?.set(key, value) }
```

sample-compose/app/src/main/java/co/nimblehq/sample/compose/model/UiModel.kt
```diff
@@ -6,7 +6,7 @@ import kotlinx.parcelize.Parcelize
 
 @Parcelize
 data class UiModel(
-    val id: String
+    val id: String = ""
```
Tuubz, line 9: Default value is no longer needed 
```suggestion
    val id: String
```

Tuubz (approved):
@kaungkhantsoe Some final touchesssss 🔥 

Rest LGTM, I'll merge after we have resolved @Wadeewee's comments

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppNavigation.kt
Wadeewee: > So it is the same object but with a different suffix name and parcelized?

@kaungkhantsoe, As I said in this [comment](https://github.com/nimblehq/android-templates/pull/390#discussion_r1092751052)

> Actually, we don't use a kind of UiModel object to passing argument to another screen

As it's a kind of `UiModel`, it was created/mapped from the response to render the `UI` on that screen. That's why we create an object with the suffix `Parcel` instead, and we do this way actually in client projects because when we navigate to the next screen, we mostly have a new `UiModel`, let's say `UiModelDetail` For more ref: You can research in `DeeMoney` project.

> As I mentioned, the sample is supposed to demo how we pass the data between screens, creating a new ModelParcel could share unnecessary boilerplate code

And also, @hoangnguyen92dn said here ☝️As it's the sample is supposed to demo how we pass the data between screens. So, It's an option for you to update then, but tbh, I don't think it's unnecessary because we usually do it in the client project. cc @hoangnguyen92dn @Tuubz 🙏

Tuubz: I think keeping it as `UiModel` is fine, since it's still being used to compose the UI 🚀

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/Item.kt
Wadeewee: @kaungkhantsoe @Tuubz, I might be wrong that I said it could effect `re-composition` 🙇‍♀️  but here're my ref why we should have a `modifier` as a parameter when we create a `Composable` function(**Item**), and basically it's like another base `Composable`, e.g., (Text, Colum, Button and so on) all of them have a parameter `modifier`

> * @param modifier The modifier to be applied to the layout. // ---(Layout = HomeScreen)--- //
 
- https://stackoverflow.com/a/71465727
- https://stackoverflow.com/a/71475970
- [Jetpack-Compose-Playground](https://foso.github.io/Jetpack-Compose-Playground/general/modifier/)

Screen Shot 2566-02-01 at 13 12 45
- And I noticed from the [Follow best practices](https://developer.android.com/jetpack/compose/performance/bestpractices) on the Composable `ContactList` they also have a parameter `modifier: Modifier = Modifier` as well.

Screen Shot 2566-02-01 at 13 20 03

Actually, I just want to know what you think about this cc @hoangnguyen92dn

Tuubz: @Wadeewee I see 

Adding `Modifier` to the constructor is good for flexibility/reusability, which I agree can be good for some cases 👍 

But I'm not 100% convinced this should be the case everywhere 💭 

So it might be better to create an RFC for this, what do you think? 🙏

Wadeewee: Yeah, I agree @Tuubz, we should have an `RFC` for this, I don't expect to update on this PR too 🚀 just ask for your idea 🙏

Wadeewee (approved):

PR #407, opened 2023-02-16
[#159][Part2] [XML] Support datastore replacing shared preference

lydiasama:
https://github.com/nimblehq/android-templates/issues/159

## What happened 👀

- [x] Part 1: Integrate for sample-compose
- [x] Part 2: Integrate for sample-xml
- [ ] Part 3: Integrate for both template-xml & template-compose

## Insight 📝

- Add the DataStore Preferences dependencies to the `Sample-Xml` project.
- Add use cases for `GetFirstTimeLaunchPreferencesUseCase` and `UpdateFirstTimeLaunchPreferencesUseCase`.
- Rename the existing `UseCase` to `GetModelsUseCase` accordingly.
- Display toast message when `isFirstTimeLaunch` flag is `true`
- Update 
``` 
android:allowBackup="false"
android:fullBackupContent="false"
```
in the AndroidManifest file to prevent share preferences caching on the device after uninstalling.

## Insight 📝

- Used the same dependencies and logic implementation from #389 
- The toast message from my POW might show twice. But after debugging, I double-checked it emits the event just once. ✅ 

## Proof Of Work 📹

https://user-images.githubusercontent.com/12026942/219269815-6bb7d0e0-c297-412f-a092-9f1bcb149851.mov


luongvo:
@lydiasama @ryan-conway @Tuubz I think the XML template will be deprecated soon. I thought the next step should be integrating to XML and Compose templates instead of spending effort for XML sample support 😢 

Anw, since @lydiasama finished her work on this part, it's nice to have.

lydiasama:
> @lydiasama @ryan-conway @Tuubz I think the XML template will be deprecated soon. I thought the next step should be integrating to XML and Compose templates instead of spending effort for XML sample support 😢
> 
> Btw, since @lydiasama finished her work on this part, it's nice to have.

@luongvo The maintainer, @Tuubz assigned me both sample repos (https://github.com/nimblehq/android-templates/pull/374#discussion_r1055081624). So yeah... 😂

kaungkhantsoe (approved):
LGTM

sample-xml/domain/src/main/java/co/nimblehq/sample/xml/domain/usecase/IsFirstTimeLaunchPreferencesUseCase.kt
```diff
@@ -0,0 +1,14 @@
+package co.nimblehq.sample.xml.domain.usecase
+
+import co.nimblehq.sample.xml.domain.repository.AppPreferencesRepository
+import kotlinx.coroutines.flow.Flow
+import javax.inject.Inject
+
+class IsFirstTimeLaunchPreferencesUseCase @Inject constructor(
```
luongvo, line 7: Could you please add tests for these 2 use-cases? Basically copy them from compose-templates.

sample-xml/app/src/main/java/co/nimblehq/sample/xml/di/modules/PreferencesModule.kt
```diff
@@ -0,0 +1,32 @@
...
+
+@Module
+@InstallIn(SingletonComponent::class)
+class PreferencesModule {
+
+    @Provides
+    @Singleton
+    fun provideAppPreferencesDataStore(
+        @ApplicationContext applicationContext: Context
+    ): DataStore<Preferences> = applicationContext.appPreferencesDataStore
```
luongvo, line 26: Should we remove the usage of `NormalSharedPreferences`?

sample-xml/app/src/main/java/co/nimblehq/sample/xml/ui/screens/home/HomeFragment.kt
```diff
@@ -52,6 +52,13 @@ class HomeFragment : BaseFragment<FragmentHomeBinding>() {
         viewModel.error bindTo ::displayError
         viewModel.navigator bindTo navigator::navigate
         viewModel.showLoading bindTo ::showLoading
```
luongvo, line 54: Do we need to apply refactoring `showLoading` to `isLoading` too?

sample-xml/app/src/main/java/co/nimblehq/sample/xml/ui/screens/home/HomeViewModel.kt
```diff
@@ -12,18 +12,24 @@ import javax.inject.Inject
...
+    private val updateFirstTimeLaunchPreferencesUseCase: UpdateFirstTimeLaunchPreferencesUseCase,
     dispatchers: DispatchersProvider
 ) : BaseViewModel(dispatchers) {
 
     private val _uiModels = MutableStateFlow<List<UiModel>>(emptyList())
     val uiModels: StateFlow<List<UiModel>>
         get() = _uiModels
 
+    private val _isFirstTimeLaunch = MutableStateFlow(false)
+    val isFirstTimeLaunch: StateFlow<Boolean>
```
luongvo, line 26: @lydiasama We can apply using SharedFlow to avoid duplicated Toast showing on the Home screen.

luongvo (commented):
rest lgtm

sample-xml/data/src/main/java/co/nimblehq/sample/xml/data/repository/AppPreferencesRepositoryImpl.kt
```diff
@@ -0,0 +1,40 @@
...
+
+import android.util.Log
+import androidx.datastore.core.DataStore
+import androidx.datastore.preferences.core.*
+import co.nimblehq.sample.xml.domain.repository.AppPreferencesRepository
+import kotlinx.coroutines.flow.*
+import java.io.IOException
+import javax.inject.Inject
+
+class AppPreferencesRepositoryImpl @Inject constructor(
```
luongvo, line 11: @lydiasama It's good to have unit-test for this Repository class too. I wrote a unit-test class in the new project which applies this new DataStore logic, please feel free to check it out as a reference for your implementation.

🚧  Note that, we might not be able to use MockK to mock & test `updateFirstTimeLaunch`, so I need to prepare a TestDataStore to test it `as an instrument test with Robolectric`.

```
private const val TEST_DATASTORE_NAME: String = "test_datastore"

@ExperimentalCoroutinesApi
@RunWith(RobolectricTestRunner::class)
class AppPreferencesRepositoryTest {

    private val testContext: Context = ApplicationProvider.getApplicationContext()
    private val testCoroutineDispatcher = UnconfinedTestDispatcher()
    private val testCoroutineScope = TestScope(testCoroutineDispatcher + Job())
    private val testDataStore: DataStore<Preferences> = PreferenceDataStoreFactory.create(
        scope = testCoroutineScope,
        produceFile =
        { testContext.preferencesDataStoreFile(TEST_DATASTORE_NAME) }
    )

    private val mockDataStore: DataStore<Preferences> = mockk()
    private lateinit var repository: AppPreferencesRepository

    @Before
    fun setup() {
        repository = AppPreferencesRepositoryImpl(mockDataStore)
    }

    @Test
    fun `When calling getItSupportInfo request successfully, it returns success response`() =
        runTest {
            val mockPreferences = mockk<Preferences>()
            every { mockPreferences[stringPreferencesKey("IT_SUPPORT_PHONE_NUMBER")] } returns "02-134-3556"
            every { mockPreferences[stringPreferencesKey("IT_SUPPORT_PHONE_NUMBER_EXT")] } returns "123"
            every { mockDataStore.data } returns flowOf(mockPreferences)

            repository.getItSupportInfo().test {
                expectMostRecentItem() shouldBe MockUtil.itSupportInfo
            }
        }

    @Test
    fun `When calling getItSupportInfo request failed with IOException, it returns empty ItSupportInfo`() =
        runTest {
            val expected = IOException()
            every { mockDataStore.data } returns flow { throw expected }

            repository.getItSupportInfo().test {
                expectMostRecentItem() shouldBe ItSupportInfo(
                    phoneNumber = "",
                    phoneNumberExtension = ""
                )
            }
        }

    @Test
    fun `When calling getItSupportInfo request failed with other exceptions, it returns wrapped error`() = runTest {
        val expected = Exception()
        every { mockDataStore.data } returns flow { throw expected }

        repository.getItSupportInfo().test {
            awaitError() shouldBe expected
        }
    }

    /**
     * FIXME Can't use MockK to mock DataStore.edit or DataStore.updateData, switch to use a test DataStore.
     */
    @Test
    fun `When calling updateItSupportInfo, it updates ItSupportInfo into DataStore`() =
        runTest {
            repository = AppPreferencesRepositoryImpl(testDataStore)

            val expected = MockUtil.itSupportInfo

            repository.updateItSupportInfo(expected)

            repository.getItSupportInfo().test {
                expectMostRecentItem() shouldBe expected
            }
        }
}
```

luongvo: Please consider adding it to the `sample-compose` project too.

sample-xml/app/src/main/java/co/nimblehq/sample/xml/ui/screens/home/HomeViewModel.kt
lydiasama: @luongvo I wonder if it's the issue of emitting several times. Since I debugged as: 
> The toast message from my POW might show twice. But after debugging, I double-checked it emits the event just once.

I found that it emits just once which is correct. 🤔 And also in sample compose, we use `StateFlow` as well. What do you think?

luongvo: @lydiasama Using `StateFlow` will re-trigger to bind the current value when `bindViewModel()` is recalled. That's why we used `SharedFlow` to display the `error message`, but `StateFlow` for `Loading` because re-showing the error message does not make sense, however, re-updating loading progress visibility is good. 

We might don't have the same issue in Compose template & sample projects because we use different ways to bind & stream data from VM to Composable.

luongvo: To have a better explanation of this issue, say your HomeViewModel has `error as StateFlow` with pre-emitted Throwable value, going back to HomeFragment from another fragment will re-trigger `bindViewModel()` and re-showing the error message 💥

sample-xml/app/src/main/java/co/nimblehq/sample/xml/di/modules/PreferencesModule.kt
```diff
@@ -0,0 +1,32 @@
...
+import javax.inject.Singleton
+
+private val Context.appPreferencesDataStore by preferencesDataStore("app_preferences")
+
+@Module
+@InstallIn(SingletonComponent::class)
+class PreferencesModule {
+
+    @Provides
+    @Singleton
```
luongvo, line 23: @lydiasama I don't think we need to set `DataStore<Preferences>` as Singleton. It is enough to put it inside a `SingletonComponent` module to keep its lifespan. Please check this for more detail: https://medium.com/digigeek/hilt-components-and-scopes-in-android-b96546cb07df

> Scoping is expensive. Use it if you really need it !!

So, please consider this issue for XML & Compose samples or templates too 🤗

lydiasama: @luongvo Please correct me if I'm wrong but from your reference, it seems we still need to add it.

> Component determines the scope of the instance. But when we ask Hilt to give us an OkHttpClient in two different places, will it be singleton or not ? By default, it is not singleton. If you want an application wide singleton OkHttpClient instance, you should annotate your OkHttpClient provider method with @Singleton.

kaungkhantsoe: @lydiasama's idea is kind of like the implementation from [this thread](https://stackoverflow.com/a/66101769). There is an updated version from 2022 and from Google developer's [article](https://medium.com/androiddevelopers/datastore-and-dependency-injection-ea32b95704e3). Since we are using hilt and providing the dependency, I think we should go with the builder pattern instead of delegate pattern here. What do you guys think? cc @luongvo

luongvo: Yes, I agree with @kaungkhantsoe. That seems to be a better + newer approach to apply. cc @lydiasama. 

Regarding my concern about using `@Singleton`. You're right, it seems we still need to keep only 1 instance of DataStore for safety. 🙏

luongvo:
@lydiasama please let me know if you have any concerns or if you need any help to finish this PR and the main task. I hope we can have the final implementation soon for the template updates 🤗

lydiasama:
@luongvo I'm trying my best to finish it today. 😢 Currently, I have 4 projects to work on. 🙏🏻

lydiasama:
@luongvo @kaungkhantsoe For the approach you suggested, I think I need to have another PR to update both `sample-comppose` too. So I would go with as-is implementation since it passed the review on `sample-compose` already. Or @kaungkhantsoe You can propose a PR later too. Please understand. 🙏🏻

kaungkhantsoe:
@lydiasama We can update the implementation using the builder pattern in a separate PR. I will help you out with the PR. cc @luongvo

lydiasama:
@kaungkhantsoe Big thanks. 🤗  Otherwise, it will reinvent the wheels. 🙏🏻

kaungkhantsoe:
Could you please create the issue and assign it to me? 🙏

sample-xml/data/src/main/java/co/nimblehq/sample/xml/data/repository/AppPreferencesRepositoryImpl.kt
lydiasama: @luongvo Resolved in bba361ff

sample-xml/domain/src/main/java/co/nimblehq/sample/xml/domain/usecase/IsFirstTimeLaunchPreferencesUseCase.kt
lydiasama: @longnd  Resolved in 08943cfc

lydiasama:
@luongvo All updated. Could you please help me recheck again? 🙏🏻

luongvo (commented):
Some more minor issues, the rest lgtm @lydiasama

lydiasama:
@luongvo Updated. 🙏🏻

luongvo (approved):
LGTM 💯

sample-compose/data/src/test/java/co/nimblehq/sample/compose/data/repository/AppPreferencesRepositoryImplTest.kt
```diff
@@ -0,0 +1,95 @@
...
+            repository.isFirstTimeLaunch().test {
+                awaitError() shouldBe expected
+            }
+        }
+
+    /**
+     * FIXME Can't use MockK to mock DataStore.edit or DataStore.updateData, switch to use a test DataStore.
+     */
+    @Test
+    fun `When calling updateFirstTimeLaunch, it updates ItSupportInfo into DataStore`() =
```
luongvo, line 83: please replace the `ItSupportInfo` word, it comes from my sample code 🤓

sample-xml/data/src/test/java/co/nimblehq/sample/xml/data/repository/AppPreferencesRepositoryImplTest.kt
```diff
@@ -0,0 +1,95 @@
...
+            repository.isFirstTimeLaunch().test {
+                awaitError() shouldBe expected
+            }
+        }
+
+    /**
+     * FIXME Can't use MockK to mock DataStore.edit or DataStore.updateData, switch to use a test DataStore.
+     */
+    @Test
+    fun `When calling updateFirstTimeLaunch, it updates ItSupportInfo into DataStore`() =
```
luongvo, line 83: same

luongvo (commented):
new minor issues

sample-xml/app/src/main/java/co/nimblehq/sample/xml/di/modules/PreferencesModule.kt
```diff
@@ -0,0 +1,32 @@
...
+    fun provideAppPreferencesDataStore(
+        @ApplicationContext applicationContext: Context
+    ): DataStore<Preferences> = applicationContext.appPreferencesDataStore
+
+    @Provides
+    fun provideAppPreferencesRepository(
+        appPreferencesDataStore: DataStore<Preferences>
+    ): AppPreferencesRepository = AppPreferencesRepositoryImpl(appPreferencesDataStore)
+}
```
AVI5HEK, line 32: We can simply bind `AppPreferencesRepositoryImpl` to `AppPreferencesRepository` since it's already using the `@Inject` annotation, which can give [better performance](https://blog.mindorks.com/dependency-injection-with-dagger2-advance/#:~:text=This%20provides%20better%20performance.).
```suggestion
    @Binds
    abstract fun bindAppPreferencesRepository(
        pppPreferencesRepositoryImpl: AppPreferencesRepositoryImpl
    ): AppPreferencesRepository
}
```

ryan-conway (approved):

sample-xml/app/src/main/java/co/nimblehq/sample/xml/di/modules/PreferencesModule.kt
lydiasama: @AVI5HEK Thanks for your suggestion. Better to have this in the part1 tho. 😭 This is part2 that we apply the same implementation from the Part1 #389 .

sample-compose/data/src/test/java/co/nimblehq/sample/compose/data/repository/AppPreferencesRepositoryImplTest.kt
lydiasama: @luongvo I fixed it at Part3 #416  🙏🏻

sample-xml/app/src/main/java/co/nimblehq/sample/xml/di/modules/PreferencesModule.kt
lydiasama: @AVI5HEK Wait, I'm afraid that we could not @Binds it because there will be both `@Provide` and `@Binds` inside the same module. So it will throw error like this:

Screenshot 2566-03-01 at 3 38 48 PM

Or did I miss something? 🙏🏻

lydiasama: @kaungkhantsoe Sorry, I don't have a permission to see your PR from gulf. 🙏🏻

AVI5HEK: @lydiasama I was going to say the same thing as @kaungkhantsoe  did. You can put the concrete provider implementation inside a `companion object` to fix this error.

kaungkhantsoe: Something like this @lydiasama
```
abstract class StorageModule {
      companion object {
         @Provides
         fun provideSecuredLocalStorage(@ApplicationContext context: Context) =
             EncryptedSharedPreferences(context)
     }

     @Singleton
     @Binds
     abstract fun provideAppPersistence(appPersistenceImpl: AppPersistenceImpl): AppPersistence
 }
```

lydiasama: @AVI5HEK Thanks for your details. I got it now. 👍🏻

lydiasama: @AVI5HEK @kaungkhantsoe Resolved at 4eb0acf in Part3 🙏🏻

Issue #425, opened 2023-03-07
[Kotlin Flow] Using "launchIn" instead of "collect" for collecting Flows in ViewModel


luongvo:
## Why

Using `launchIn` [instead of `collect`](https://github.com/nimblehq/android-templates/blob/74b85b4f5a2a95e5e2ea815a09dde4312d0fe8b3/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeViewModel.kt#L30-L41) for collecting Flows in ViewModel is recommended way, because:
- `collect()` will block the coroutine until the flow has finished emitting. This behavior is sometimes desired, but it’s not in most cases.
- `launchIn` is less code to write, we don’t have to have indentations, and easier to understand.
- We can eliminate the `execute` operator in the BaseVIewModel.

Sample implementation:

```
        logInUseCase(
            email = email,
            password = password
        )
            .onStart { showLoading() }
            .onCompletion { hideLoading() }
            .catch { e -> _error.emit(e) }
            .onEach {
                _isLoggedIn.emit(true)
                navigateToHome()
            }
            .launchIn(viewModelScope)
```

RFC discussion: https://github.com/nimblehq/android-templates/discussions/408

References: 
- https://handstandsam.com/2021/02/19/the-best-way-to-collect-a-flow-in-kotlin-launchin/
- https://itnext.io/differences-in-methods-of-collecting-kotlin-flows-3d1d4efd1c2

## Who Benefits?

Developers

PR #441, opened 2023-03-10
[#425] [Kotlin Flow] Using "launchIn" instead of "collect" for collecting Flows in ViewModel

luongvo:
https://github.com/nimblehq/android-templates/issues/425

## What happened 👀

- Use `launchIn` instead of `collect` for collecting Flows in ViewModel.
- Replace the old `execute` protected fun in BaseViewModel with the new `launch` fun (renaming for consistency).
- Add `injectLoading` fun to reduce the logic for `showLoading` and `hideLoading` at `onStart` and `onCompletion` in each Flow.

## Insight 📝

- The new `launch` fun should accept `EmptyCoroutineContext` as default ([same as the original fun](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/launch.html)). [With `EmptyCoroutineContext`, the `Dispatchers.Default` is used](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/launch.html) 👉 **main thread**, to launch the Job.

  image

- [To launch a Flow in a specific CoroutineContext](https://developer.android.com/kotlin/flow), such as `IO` for storage or network execution. We must use [`flowOn`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/flow-on.html), e.g. `.flowOn(dispatchers.io)`. Hence, we don't need to pass `DispatchersProvider` into the BaseViewModel.
- Note that `flowOn` changes the CoroutineContext of the upstream flow.

  image

- This update ensures we will not launch all unnecessary Flows on Dispatcher.IO at all times.

  Before ❌  https://github.com/nimblehq/android-templates/blob/5a037e8ca67447e70610463bb77f2146a56a6e66/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/base/BaseViewModel.kt#L49-L52
  
  After ✅  https://github.com/nimblehq/android-templates/blob/869f82ea73f98e9a47d5934485ff0a3810a5cb7a/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeViewModel.kt#L29-L37

- `.catch { }` should always be put at the end of a Flow collector, in front of `.launchIn` to catch all the prior errors.

## Proof Of Work 📹

- All the core app logic remains the same as before ✅ 
- All tests are passed ✅

### DIFF
--- a/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/base/BaseViewModel.kt
+++ b/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/base/BaseViewModel.kt
@@ -4,8 +4,10 @@ import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import co.nimblehq.sample.compose.lib.IsLoading
 import co.nimblehq.sample.compose.ui.AppDestination
+import kotlinx.coroutines.*
 import kotlinx.coroutines.flow.*
-import kotlinx.coroutines.launch
+import kotlin.coroutines.CoroutineContext
+import kotlin.coroutines.EmptyCoroutineContext
 
 @Suppress("PropertyName")
 abstract class BaseViewModel : ViewModel() {
@@ -41,8 +43,8 @@ abstract class BaseViewModel : ViewModel() {
         }
     }
 
-    protected fun launch(job: suspend () -> Unit) =
-        viewModelScope.launch {
+    protected fun launch(context: CoroutineContext = EmptyCoroutineContext, job: suspend () -> Unit) =
+        viewModelScope.launch(context) {
             job.invoke()
         }
 
--- a/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/base/BaseViewModel.kt
+++ b/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/base/BaseViewModel.kt
@@ -4,27 +4,22 @@ import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import co.nimblehq.sample.compose.lib.IsLoading
 import co.nimblehq.sample.compose.ui.AppDestination
-import co.nimblehq.sample.compose.util.DispatchersProvider
-import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.flow.*
 import kotlinx.coroutines.launch
 
 @Suppress("PropertyName")
-abstract class BaseViewModel(private val dispatchersProvider: DispatchersProvider) : ViewModel() {
+abstract class BaseViewModel : ViewModel() {
 
     private var loadingCount: Int = 0
 
     private val _isLoading = MutableStateFlow(false)
-    val isLoading: StateFlow<IsLoading>
-        get() = _isLoading
+    val isLoading: StateFlow<IsLoading> = _isLoading
 
     protected val _error = MutableStateFlow<Throwable?>(null)
-    val error: StateFlow<Throwable?>
-        get() = _error
+    val error: StateFlow<Throwable?> = _error
 
     protected val _navigator = MutableSharedFlow<AppDestination>()
-    val navigator: SharedFlow<AppDestination>
-        get() = _navigator
+    val navigator: SharedFlow<AppDestination> = _navigator
 
     /**
      * To show loading manually, should call `hideLoading` after
@@ -46,8 +41,12 @@ abstract class BaseViewModel(private val dispatchersProvider: DispatchersProvide
         }
     }
 
-    fun execute(coroutineDispatcher: CoroutineDispatcher = dispatchersProvider.io, job: suspend () -> Unit) =
-        viewModelScope.launch(coroutineDispatcher) {
+    protected fun launch(job: suspend () -> Unit) =
+        viewModelScope.launch {
             job.invoke()
         }
+
+    protected fun <T> Flow<T>.injectLoading(): Flow<T> = this
+        .onStart { showLoading() }
+        .onCompletion { hideLoading() }
 }
--- a/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeViewModel.kt
+++ b/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeViewModel.kt
@@ -1,11 +1,11 @@
 package co.nimblehq.sample.compose.ui.screens.home
 
+import androidx.lifecycle.viewModelScope
 import co.nimblehq.sample.compose.domain.usecase.*
 import co.nimblehq.sample.compose.model.UiModel
 import co.nimblehq.sample.compose.model.toUiModel
 import co.nimblehq.sample.compose.ui.AppDestination
 import co.nimblehq.sample.compose.ui.base.BaseViewModel
-import co.nimblehq.sample.compose.util.DispatchersProvider
 import dagger.hilt.android.lifecycle.HiltViewModel
 import kotlinx.coroutines.flow.*
 import javax.inject.Inject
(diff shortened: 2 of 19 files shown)


AVI5HEK (commented):
Overall LGTM.

sample-xml/app/src/main/java/co/nimblehq/sample/xml/ui/screens/home/HomeViewModel.kt
```diff
@@ -12,39 +13,33 @@ import javax.inject.Inject
 
 @HiltViewModel
 class HomeViewModel @Inject constructor(
-    private val getModelsUseCase: GetModelsUseCase,
+    dispatchers: DispatchersProvider,
+    getModelsUseCase: GetModelsUseCase,
     private val isFirstTimeLaunchPreferencesUseCase: IsFirstTimeLaunchPreferencesUseCase,
```
ryan-conway, line 18: Since we've updated `getModelsUseCase` to not be a `val` anymore, we might as well update these too (and other `HomeViewModel` classes) for consistency, what do you think? 😄 

The rest LGTM 💪🏻

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/base/BaseViewModel.kt
```diff
@@ -46,8 +43,12 @@ abstract class BaseViewModel(private val dispatchersProvider: DispatchersProvide
...
-    fun execute(coroutineDispatcher: CoroutineDispatcher = dispatchersProvider.io, job: suspend () -> Unit) =
-        viewModelScope.launch(coroutineDispatcher) {
+    protected fun launch(context: CoroutineContext = EmptyCoroutineContext, job: suspend () -> Unit) =
+        viewModelScope.launch(context) {
             job.invoke()
         }
+
+    protected fun <T> Flow<T>.injectLoading(): Flow<T> = this
+        .onStart { showLoading() }
+        .onCompletion { hideLoading() }
```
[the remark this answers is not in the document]
luongvo: @AVI5HEK I don't think combining `injectLoading` and `onEach` into 1 new operator is a good idea. It's for shorter code, yes; however, they're different logic, so combining them could bring confusion. Especially sometimes, we don't want to handle the loading state for some Flow collecting flows. 💭

sample-xml/app/src/main/java/co/nimblehq/sample/xml/ui/screens/home/HomeViewModel.kt
luongvo: Fixed in https://github.com/nimblehq/android-templates/pull/441/commits/4ca6e75bde07ff851340dae3aca03fa74e9ab2b8

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeViewModel.kt
```diff
@@ -12,39 +13,33 @@ import javax.inject.Inject
 
 @HiltViewModel
 class HomeViewModel @Inject constructor(
-    private val getModelsUseCase: GetModelsUseCase,
+    dispatchers: DispatchersProvider,
```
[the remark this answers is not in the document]
luongvo: Fixed in https://github.com/nimblehq/android-templates/pull/441/commits/4ca6e75bde07ff851340dae3aca03fa74e9ab2b8

ryan-conway (approved):
👏🏻

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeViewModel.kt
```diff
@@ -12,39 +15,33 @@ import javax.inject.Inject
...
 class HomeViewModel @Inject constructor(
-    private val getModelsUseCase: GetModelsUseCase,
-    private val isFirstTimeLaunchPreferencesUseCase: IsFirstTimeLaunchPreferencesUseCase,
-    private val updateFirstTimeLaunchPreferencesUseCase: UpdateFirstTimeLaunchPreferencesUseCase,
-    dispatchers: DispatchersProvider
-) : BaseViewModel(dispatchers) {
+    dispatchersProvider: DispatchersProvider,
+    getModelsUseCase: GetModelsUseCase,
+    isFirstTimeLaunchPreferencesUseCase: IsFirstTimeLaunchPreferencesUseCase,
+    updateFirstTimeLaunchPreferencesUseCase: UpdateFirstTimeLaunchPreferencesUseCase,
```
minhnimble, line 21: Just curious, but do we have any convention on how we group the ViewModel's attributes? As I saw from the previous code, we are placing `dispatchers: DispatchersProvider` at the very bottom, but we are moving it to the top now. Should we always group the use cases at the top like the same approach as the old code? 🤔

sample-xml/app/src/main/java/co/nimblehq/sample/xml/ui/screens/home/HomeViewModel.kt
```diff
@@ -12,39 +13,33 @@ import javax.inject.Inject
...
 class HomeViewModel @Inject constructor(
-    private val getModelsUseCase: GetModelsUseCase,
-    private val isFirstTimeLaunchPreferencesUseCase: IsFirstTimeLaunchPreferencesUseCase,
-    private val updateFirstTimeLaunchPreferencesUseCase: UpdateFirstTimeLaunchPreferencesUseCase,
-    dispatchers: DispatchersProvider
-) : BaseViewModel(dispatchers) {
+    dispatchersProvider: DispatchersProvider,
+    getModelsUseCase: GetModelsUseCase,
+    isFirstTimeLaunchPreferencesUseCase: IsFirstTimeLaunchPreferencesUseCase,
+    updateFirstTimeLaunchPreferencesUseCase: UpdateFirstTimeLaunchPreferencesUseCase,
```
minhnimble, line 19: Same concern here 🙏

template-compose/app/src/main/java/co/nimblehq/template/compose/ui/screens/home/HomeViewModel.kt
```diff
@@ -11,26 +12,22 @@ import javax.inject.Inject
 
 @HiltViewModel
 class HomeViewModel @Inject constructor(
-    private val useCase: UseCase,
-    dispatchers: DispatchersProvider
-) : BaseViewModel(dispatchers) {
+    dispatchersProvider: DispatchersProvider,
+    useCase: UseCase,
```
minhnimble, line 16: And here 🙏

minhnimble (commented):
Just a small concern regarding the ViewModel's attributes order, the rest LGTM 🙏

AVI5HEK (approved):

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeViewModel.kt
luongvo: @minhnimble Good question 👍 I can't find an actual convention for this, however, from [this code sample](https://kotlinlang.org/docs/coding-conventions.html#class-properties-and-parameters), I think we should put `non val` properties at the end 👍 

So in final, the order could be like this 💭 
- public val
- non public val
- private val
- non private val

minhnimble (approved):
The rest LGTM 🚀

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeViewModel.kt
minhnimble: @luongvo FYI, a Discussion is created for this convention here https://github.com/nimblehq/android-templates/discussions/451 🙏

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/base/BaseViewModel.kt
```diff
@@ -46,8 +43,12 @@ abstract class BaseViewModel(private val dispatchersProvider: DispatchersProvide
         }
     }
 
-    fun execute(coroutineDispatcher: CoroutineDispatcher = dispatchersProvider.io, job: suspend () -> Unit) =
-        viewModelScope.launch(coroutineDispatcher) {
+    protected fun launch(context: CoroutineContext = EmptyCoroutineContext, job: suspend () -> Unit) =
+        viewModelScope.launch(context) {
             job.invoke()
         }
```
kaungkhantsoe, line 49: Here is a little bit of imporvement but I feel like we are over-engineering this part. It is almost the same as calling `viewModelScope.launch` by ourselves. What do you guys think? @nimblehq/android-chapter 
```suggestion
    protected fun launch(context: CoroutineContext = EmptyCoroutineContext, job: suspend CoroutineScope.() -> Unit) =
        viewModelScope.launch(context, block = job)
```

luongvo: @kaungkhantsoe Actually, I tried to follow the current approach to shorten the default execution, then we can get rid of some max line length Detekt issues. I follow your options here @nimblehq/android-chapter  😄

kaungkhantsoe (approved):

chornerman (approved):
LGTM 🚀

Issue #427, opened 2023-03-07
[Compose] Defining a "collectAsEffect" extension for Compose to reduce boilerplate code


luongvo:
## Why

On the way of defining some valuable extensions for Compose to reduce boilerplate code, we can define a `collectAsEffect`:

```
@Composable
fun <T> Flow<T>.collectAsEffect(
    context: CoroutineContext = EmptyCoroutineContext,
    block: (T) -> Unit
) {
    LaunchedEffect(key1 = Unit) {
        onEach(block).flowOn(context).launchIn(this)
    }
}
```

So with a long implementation like this 🤔 ,

```
LaunchedEffect(viewModel.error) {
    viewModel.error.collect { error -> error.showToast(context) }
}

LaunchedEffect(viewModel.navigator) {
    viewModel.navigator.collect { destination -> navigator(destination) }
}
```

we can now shorten it by using the above `collectAsEffect` extension 👏 

```
viewModel.error.collectAsEffect { error -> error.showToast(context) }
viewModel.navigator.collectAsEffect { destination -> navigator(destination) }
```

RFC discussion: https://github.com/nimblehq/android-templates/discussions/408

References: 
- https://github.com/elson-dsouza/PopMovies/blob/dce98cd63540e6c88b8913a978c78edaf2fe035e/app/src/main/java/com/example/elson/popmovies/compose/EventsSupport.kt
- https://github.com/igamgam97/MovieDB/blob/a2cd3e0dd171115c9d422f96944e0b19a41cee95/core/common/src/main/java/com/noveogroup/moviecatalog/core/common/FlowExt.kt
- https://github.com/Cilestal/Kotlin-Multiplatform-MVI-sample-Android-Compose-iOS-SwiftUI-/blob/7e68ec688ef7c0685a883a708c06454fe60d8d0c/androidApp/src/main/java/com/cilestal/mvi_mpp/android/extension/ComposeExt.kt
- https://github.com/adiRandom/grocever-android/blob/eab84e8a6294a7ad429a51f5353620d35e6b033b/app/src/main/java/app/adi_random/dealscraper/usecase/FlowUtils.kt
- https://github.com/adiRandom/grocever-android/blob/39fc41cd46ef0f6497b289059359c3bb56e1b0c3/app/src/main/java/app/adi_random/dealscraper/usecase/FlowUtils.kt

## Who Benefits?

Developers

PR #449, opened 2023-03-27
[#427] [Compose] Defining a "collectAsEffect" extension for Compose to reduce boilerplate code

luongvo:
https://github.com/nimblehq/android-templates/issues/427

## What happened 👀

- Defined a Flow extension `collectAsEffect` to reduce boilerplate code with LaunchEffect.
- Shortened all `SharedFlow` flows by using the collectAsEffect extension 👏
- Added a default error handling for the `Compose Template` project.

## Insight 📝

- As [discussed in the RFC](https://github.com/nimblehq/android-templates/discussions/408#discussioncomment-5060034), we can use this extension for a one-time action or multiple-times action of a **SharedFlow** when value changes to shorten the logic of using `LaunchedEffect`. 
- Since we used `onEach(block).flowOn(context).launchIn(this)`, we don't need to expose `key1` for value changes to re-subscribing. The `onEach(block).flowOn(context).launchIn(this)` automatically emits new values from Flow itself without the help from `LaunchedEffect(key1 = )` ✅ 
- For **StateFlow**, we should continue using `collectAsStateWithLifecycle()`.

## Proof Of Work 📹

- The SharedFlow collecting logic in the Composable screen is shortened and more simple to focus on the UI logic ✅ 

- The `navigator` logic from the Compose Sample project works normally ✅ 

https://user-images.githubusercontent.com/16315358/228491641-8e3f05ec-1a78-4337-a22b-a81f2ced4883.mp4

- The new default `error` logic from the Compose Template project works normally > no networking error ✅ 

https://user-images.githubusercontent.com/16315358/228496668-005d87a0-4803-4b98-9ca4-cc4bfd5436e5.mp4


sample-compose/app/src/main/java/co/nimblehq/sample/compose/extensions/FlowExt.kt
```diff
@@ -0,0 +1,19 @@
...
+@SuppressLint("ComposableNaming")
+@Composable
+fun <T> Flow<T>.collectAsEffect(
+    context: CoroutineContext = EmptyCoroutineContext,
+    block: (T) -> Unit,
+) {
+    LaunchedEffect(key1 = Unit) {
+        onEach(block).flowOn(context).launchIn(this)
+    }
+}
```
minhnimble, line 19: As the team raised about [allowing modification for the LaunchedEffect's key1](https://github.com/nimblehq/android-templates/discussions/408#discussioncomment-5060034), I also think we should support passing value for this key as it can inform `LaunchedEffect to whether the coroutine should restart or not when the key1 parameter value changes. 🙏

You can refer to the following link as a good example use case regarding the need to modify this key e.g., re-executing the API:
https://betterprogramming.pub/jetpack-compose-side-effects-launchedeffect-with-example-99c2f51ff463

sample-compose/app/src/main/java/co/nimblehq/sample/compose/extensions/FlowExt.kt
```diff
@@ -0,0 +1,19 @@
+package co.nimblehq.sample.compose.extensions
+
+import android.annotation.SuppressLint
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
+import kotlinx.coroutines.flow.*
+import kotlin.coroutines.CoroutineContext
+import kotlin.coroutines.EmptyCoroutineContext
+
+@SuppressLint("ComposableNaming")
```
minhnimble, line 10: 🙈

luongvo: It's not a Composable view, it's an extension so we should not follow the Composable naming convention 😛

sample-compose/app/src/main/java/co/nimblehq/sample/compose/extensions/FlowExt.kt
luongvo: @minhnimble Since we used `onEach(block).flowOn(context).launchIn(this)`, we don't need to expose `key1` for value changes to re-subscribing. The `onEach(block).flowOn(context).launchIn(this)` is automatically emit new value from Flow itself without the help from `LaunchedEffect(key1 = )` 😄 

I already explained the unnecessary of exposing `key1` in PR's description.
image

minhnimble: Okay, I got it now. Previously I had a look through the PR description, but it was still ambiguous for me that you just mentioned:

> We don't need to expose a key to replace the Unit, because we already used `onEach(block).flowOn(context).launchIn(this)`.

=> Maybe you should explain a bit deeper in the description into how the fact that `onEach(block).flowOn(context).launchIn(this)` can automatically emit new values from Flow itself without the help from `LaunchedEffect` with key. 🙏

template-compose/app/src/main/java/co/nimblehq/template/compose/ui/ErrorMapping.kt
```diff
@@ -10,3 +11,6 @@ fun Throwable.userReadableMessage(context: Context): String {
         else -> message
     } ?: context.getString(R.string.error_generic)
 }
+
+fun Throwable.showToast(context: Context) =
+    Toast.makeText(context, userReadableMessage(context), Toast.LENGTH_SHORT).show()
```
minhnimble, line 16: I think we might need a better naming for this file as with this new extension fun, it is no longer used for `ErrorMapping` only => `ThrowableExtension.kt` maybe? 🤔

template-compose/app/src/main/java/co/nimblehq/template/compose/ui/screens/home/HomeScreen.kt
```diff
@@ -22,6 +25,10 @@ fun HomeScreen(
     viewModel: HomeViewModel = hiltViewModel(),
     navigator: (destination: AppDestination) -> Unit
 ) {
+    val context = LocalContext.current
+    viewModel.error.collectAsEffect { e -> e.showToast(context) }
```
ryan-conway, line 29: Any reason we don't make use of the `Toaster` class here? 🤔

sample-compose/app/src/main/java/co/nimblehq/sample/compose/extensions/FlowExt.kt
luongvo: @minhnimble Yeah. Updated 👍

template-compose/app/src/main/java/co/nimblehq/template/compose/ui/ErrorMapping.kt
luongvo: @minhnimble This `ErrorMapping` is not a common thing like other extensions. It's specific for the error showing of a project, and might not be used in other projects because we could switch between custom messages from different `Throwable`s.

image

So I don't think we should rename it to `ThrowableExtension` or move it to the common extension package. This looks like a project component instead of a common code to be reused 🤔

template-compose/app/src/main/java/co/nimblehq/template/compose/ui/screens/home/HomeScreen.kt
luongvo: @ryan-conway ComposableScreen is not the same as a Fragment so we can easily bind a `Toaster` component. Let me check whether we can do anything to make use of it. 🤔

luongvo: @ryan-conway I tried to make it work with the ComposableScreen. However, from the updated code, I don't see much benefit of having a Toaster component with Jetpack Compose anymore 🤷‍♂️ You can see my insights from the diff below:

image

image

Instead of having a Toaster component, we can simply have a Context extension to show the Toast 💡

luongvo: @ryan-conway Fixed in https://github.com/nimblehq/android-templates/pull/449/commits/82c99e598454daeaab58d3916514ada106a9ade5 👉 Replace the simple `Toaster` component by an extension.

chornerman (approved):
LGTM 🚀

sample-compose/app/src/main/java/co/nimblehq/sample/compose/extensions/FlowExt.kt
```diff
@@ -0,0 +1,19 @@
...
+@Composable
+fun <T> Flow<T>.collectAsEffect(
+    context: CoroutineContext = EmptyCoroutineContext,
+    block: (T) -> Unit,
```
kaungkhantsoe, line 14: Can be improved a little bit so that we can also run suspend functions like `delay()`
```suggestion
    block: suspend (T) -> Unit,
```

template-compose/app/src/main/java/co/nimblehq/template/compose/extensions/FlowExt.kt
```diff
@@ -0,0 +1,19 @@
...
+import androidx.compose.runtime.LaunchedEffect
+import kotlinx.coroutines.flow.*
+import kotlin.coroutines.CoroutineContext
+import kotlin.coroutines.EmptyCoroutineContext
+
+@SuppressLint("ComposableNaming")
+@Composable
+fun <T> Flow<T>.collectAsEffect(
+    context: CoroutineContext = EmptyCoroutineContext,
+    block: (T) -> Unit,
```
kaungkhantsoe, line 14: This too

kaungkhantsoe (commented):
Rest LGTM

sample-compose/app/src/main/java/co/nimblehq/sample/compose/extensions/FlowExt.kt
luongvo: Fixed in https://github.com/nimblehq/android-templates/pull/449/commits/618fd7d9f2031c8695ee58ee2f6d59a644f1b7a2

template-compose/app/src/main/java/co/nimblehq/template/compose/ui/ErrorMapping.kt
minhnimble: well-noted @luongvo it is a good point to move the `showToast` to `ContextExt` too, let's go 🚀

minhnimble (approved):
LGTM 🚀

sample-compose/app/src/main/java/co/nimblehq/sample/compose/extensions/ContextExt.kt
```diff
@@ -0,0 +1,7 @@
+package co.nimblehq.sample.compose.extensions
+
+import android.content.Context
+import android.widget.Toast
+
+fun Context.showToast(message: String) =
+    Toast.makeText(this, message, Toast.LENGTH_SHORT).show()
```
thiennguyen0196, line 7: (Optional) I think we should make all the variables flexible for different usage 😄 
```suggestion
fun Context.showToast(message: String, length: Int = Toast.LENGTH_SHORT) =
    Toast.makeText(this, message, length).show()
```

template-compose/app/src/main/java/co/nimblehq/template/compose/extensions/ContextExt.kt
```diff
@@ -0,0 +1,7 @@
+package co.nimblehq.template.compose.extensions
+
+import android.content.Context
+import android.widget.Toast
+
+fun Context.showToast(message: String) =
+    Toast.makeText(this, message, Toast.LENGTH_SHORT).show()
```
thiennguyen0196, line 7: (Optional) Ditto here 😄 
```suggestion
fun Context.showToast(message: String, length: Int = Toast.LENGTH_SHORT) =
    Toast.makeText(this, message, length).show()
```

thiennguyen0196 (commented):
Nice implementation @luongvo 👏 
I just have 2 small optional recommendation, the rest LGTM 🚀

kaungkhantsoe (approved):

sample-compose/app/src/main/java/co/nimblehq/sample/compose/extensions/ContextExt.kt
luongvo: Actually, we can make it more flexible when used in an actual project. Let me help to make it available from the template.

luongvo: Fixed in https://github.com/nimblehq/android-templates/pull/449/commits/068afe2c3dbb379f7947df52637aa9f24505c1ac

thiennguyen0196 (approved):
lgtm now 🚀

Issue #419, opened 2023-03-01
Fix: The toast message for first time launching the app is showing duplicate for some devices in Sample projects


lydiasama:
## Issue

When testing with `Samsung Note9`, it shows the **first time launch** toast message correctly. But it shows duplicates when testing with the emulator device--`Pixel 3a`.

## Expected

Show the toast message only once at Home screen when first launch the app.

## Steps to reproduce

1. Launch the app for the first time
2. Click an item to navigate to another screen
3. Get back to the Home screen.

luongvo:
@lydiasama is this come from the current develop or your working branch? XML or Compose.
We need more detail such as a git hash as prerequisites to know where is the bug come from 👋

lydiasama:
@luongvo Updated the description.

PR #478, opened 2023-06-21
[#419] [Bug] Prevent the toast for first time app launch from reappearing after navigating back from the second screen

AVI5HEK:
https://github.com/nimblehq/android-templates/issues/419

## What happened 👀

When the app is launched for the first time, we show a toast saying "This is the first time launch". After navigating to the second screen by clicking an item on the list and then navigating back to home, the toast reappears.

Expected behavior: show the toast message only once on the Home screen when the app is launched for the first time.

## Insight 📝

* The `isFirstTimeLaunch` `StateFlow` is being observed from `HomeScreen`. `isFirstTimeLaunch` is updated only once when initializing the VM. So everytime the user navigates back from the second screen, the initial value `true` is observed by `HomeScreen` resulting in the bug.
* Fixed the issue by updating the `SharedPreferences` value as well as `isFirstTimeLaunch` only _after_ the toast is shown on the screen.
* Potential issue: if `UpdateFirstTimeLaunchPreferencesUseCase` runs into an error, `isFirstTimeLaunch` will still be updated with `false` and restarting the app will show the toast again.

## Proof Of Work 📹

https://github.com/nimblehq/android-templates/assets/8093908/7d69388b-da84-4b36-8307-66470e181306

### DIFF
--- a/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeViewModel.kt
+++ b/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeViewModel.kt
@@ -38,13 +38,13 @@ class HomeViewModel @Inject constructor(
             .catch { e -> _error.emit(e) }
             .launchIn(viewModelScope)
 
-        launch(dispatchersProvider.io) {
-            val isFirstTimeLaunch = isFirstTimeLaunchPreferencesUseCase()
-                .catch { e -> _error.emit(e) }
-                .first()
-
-            _isFirstTimeLaunch.emit(isFirstTimeLaunch)
-        }
+        isFirstTimeLaunchPreferencesUseCase()
+            .onEach { isFirstTimeLaunch ->
+                _isFirstTimeLaunch.emit(isFirstTimeLaunch)
+            }
+            .flowOn(dispatchersProvider.io)
+            .catch { e -> _error.emit(e) }
+            .launchIn(viewModelScope)
     }
 
     fun onFirstTimeLaunch() {
--- a/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeViewModel.kt
+++ b/sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeViewModel.kt
@@ -15,10 +15,10 @@ import javax.inject.Inject
 
 @HiltViewModel
 class HomeViewModel @Inject constructor(
-    dispatchersProvider: DispatchersProvider,
     getModelsUseCase: GetModelsUseCase,
     isFirstTimeLaunchPreferencesUseCase: IsFirstTimeLaunchPreferencesUseCase,
-    updateFirstTimeLaunchPreferencesUseCase: UpdateFirstTimeLaunchPreferencesUseCase,
+    private val updateFirstTimeLaunchPreferencesUseCase: UpdateFirstTimeLaunchPreferencesUseCase,
+    private val dispatchersProvider: DispatchersProvider,
 ) : BaseViewModel() {
 
     private val _uiModels = MutableStateFlow<List<UiModel>>(emptyList())
@@ -44,9 +44,13 @@ class HomeViewModel @Inject constructor(
                 .first()
 
             _isFirstTimeLaunch.emit(isFirstTimeLaunch)
-            if (isFirstTimeLaunch) {
-                updateFirstTimeLaunchPreferencesUseCase(false)
-            }
+        }
+    }
+
+    fun onFirstTimeLaunch() {
+        launch(dispatchersProvider.io) {
+            updateFirstTimeLaunchPreferencesUseCase(false)
+            _isFirstTimeLaunch.emit(false)
         }
     }
 
--- a/sample-compose/app/src/test/java/co/nimblehq/sample/compose/ui/screens/home/HomeScreenTest.kt
+++ b/sample-compose/app/src/test/java/co/nimblehq/sample/compose/ui/screens/home/HomeScreenTest.kt
@@ -106,10 +106,10 @@ class HomeScreenTest : BaseScreenTest() {
 
     private fun initViewModel() {
         viewModel = HomeViewModel(
-            coroutinesRule.testDispatcherProvider,
             mockGetModelsUseCase,
             mockIsFirstTimeLaunchPreferencesUseCase,
-            mockUpdateFirstTimeLaunchPreferencesUseCase
+            mockUpdateFirstTimeLaunchPreferencesUseCase,
+            coroutinesRule.testDispatcherProvider
         )
     }
 }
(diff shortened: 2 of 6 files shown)


sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeViewModel.kt
```diff
@@ -44,9 +44,13 @@ class HomeViewModel @Inject constructor(
                 .first()
 
             _isFirstTimeLaunch.emit(isFirstTimeLaunch)
```
Wadeewee, line 46: @AVI5HEK, It would be great if we handle the `scope` by `launchIn` like we do for `getModelsUseCase` above case. So we can get rid of calling `.first()`, Would you mind updating the existing logic 🙏  

```kotlin
        isFirstTimeLaunchPreferencesUseCase()
            .onEach { isFirstTimeLaunch ->
                _isFirstTimeLaunch.emit(isFirstTimeLaunch)
            }
            .flowOn(dispatchersProvider.io)
            .catch { e -> _error.emit(e) }
            .launchIn(viewModelScope)
```

sample-compose/app/src/test/java/co/nimblehq/sample/compose/ui/screens/home/HomeScreenTest.kt
```diff
@@ -106,10 +106,10 @@ class HomeScreenTest : BaseScreenTest() {
 
     private fun initViewModel() {
         viewModel = HomeViewModel(
-            coroutinesRule.testDispatcherProvider,
             mockGetModelsUseCase,
             mockIsFirstTimeLaunchPreferencesUseCase,
-            mockUpdateFirstTimeLaunchPreferencesUseCase
+            mockUpdateFirstTimeLaunchPreferencesUseCase,
```
Wadeewee, line 111: Can we add the test case for the `FirstTimeLaunch`?

Wadeewee (commented):
For the rest LGTM 👍

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeViewModel.kt
AVI5HEK: Updated in https://github.com/nimblehq/android-templates/pull/478/commits/d8c44c06ca24799ca04afcc1b59913dd1c4725d9

sample-compose/app/src/test/java/co/nimblehq/sample/compose/ui/screens/home/HomeScreenTest.kt
AVI5HEK: Updated in https://github.com/nimblehq/android-templates/pull/478/commits/d8c44c06ca24799ca04afcc1b59913dd1c4725d9

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/screens/home/HomeViewModel.kt
```diff
@@ -38,15 +38,19 @@ class HomeViewModel @Inject constructor(
...
-            _isFirstTimeLaunch.emit(isFirstTimeLaunch)
-            if (isFirstTimeLaunch) {
-                updateFirstTimeLaunchPreferencesUseCase(false)
+        isFirstTimeLaunchPreferencesUseCase()
+            .onEach { isFirstTimeLaunch ->
+                _isFirstTimeLaunch.emit(isFirstTimeLaunch)
             }
+            .flowOn(dispatchersProvider.io)
+            .catch { e -> _error.emit(e) }
+            .launchIn(viewModelScope)
```
kaungkhantsoe, line 47: Need to `catch` before `onEach` or it will throw I think.
```suggestion
        isFirstTimeLaunchPreferencesUseCase()
            .catch { e -> _error.emit(e) }
            .onEach { isFirstTimeLaunch ->
                _isFirstTimeLaunch.emit(isFirstTimeLaunch)
            }
            .flowOn(dispatchersProvider.io)
            .launchIn(viewModelScope)
```

AVI5HEK: The unit test for the error case seems to be working fine with the current code: https://github.com/nimblehq/android-templates/pull/478/commits/29ccca967a3952f212823388c49c57053600e98a

sample-compose/app/src/test/java/co/nimblehq/sample/compose/ui/screens/home/HomeScreenTest.kt
Wadeewee: @AVI5HEK, What about in `HomeScreenTest`?

AVI5HEK: @Wadeewee done. https://github.com/nimblehq/android-templates/pull/478/commits/2781508896349644162e42e186fa703ad23dd090

sample-compose/app/src/test/java/co/nimblehq/sample/compose/ui/screens/home/HomeScreenTest.kt
```diff
@@ -56,6 +56,28 @@ class HomeScreenTest : BaseScreenTest() {
...
 
+    @Test
+    fun `When entering the Home screen for the first time, it shows a toast confirming that`() {
+        every { mockIsFirstTimeLaunchPreferencesUseCase() } returns flowOf(true)
+
+        initComposable {
+            composeRule.waitForIdle()
+            advanceUntilIdle()
+
+            ShadowToast.showedToast(activity.getString(R.string.message_first_time_launch)) shouldBe true
```
Wadeewee, line 67: My last concern @AVI5HEK, Let's verify `UpdateFirstTimeLaunchPreferencesUseCase` as well 🚀 
```suggestion
            ShadowToast.showedToast(activity.getString(R.string.message_first_time_launch)) shouldBe true

            coVerify(exactly = 1) {
                mockUpdateFirstTimeLaunchPreferencesUseCase(false)
            }
```

AVI5HEK: @Wadeewee IMO, that's not a concern of the UI. We have that covered in the ViewModel test.

kaungkhantsoe (approved):

sample-compose/app/src/test/java/co/nimblehq/sample/compose/ui/screens/home/HomeScreenTest.kt
Wadeewee: I think it is good to have it but if you think it's unnecessary, So it's optional for you to update then 🚀

Wadeewee (approved):

ryan-conway (approved):

Issue #479, opened 2023-06-28
Fix: Moshi @Json annotation does not work properly on the release build


luongvo:
## Issue

Moshi does NOT seem to parse the JSON model fields with `@Json` annotation properly on the release build. All fields with `@Json` annotation that differs between the `field name` & JSON `key name` can not be parsed and fall back to `null` values.

![image](https://github.com/nimblehq/android-templates/assets/16315358/4f4afb5f-ed32-47d7-91ec-e53bd9ef0d58)

## Expected

JSON model fields with `@Json` annotation should be parsed correctly instead of `null`. 

## Steps to reproduce

1. Create a sample for `@Json` annotation with different field names & JSON key names.
2. Switch to the release build.
3. Run and check the parsing result.

PR #480, opened 2023-06-28
[#479] Fix: Moshi @Json annotation does not work properly on the release build

luongvo:
https://github.com/nimblehq/android-templates/issues/479

## What happened 👀

- Fix: keep all data models in proguard.
- Fix: @JSON does not work on the release build by `upgrading AGP to 7.4.2`.

## Insight 📝

- The Moshi `@Json annotation does not work on the release build` (while working with debug build) 👉 I did some research on this:
  - [Updating proguard rules does not work](https://stackoverflow.com/questions/57034937/moshi-kotlinjsonadapterfactory-cannot-parse-json-after-enabled-minify) ❌ 
  - [Adding `@Keep` or `@JsonClass(generateAdapter = false)` annotations does not work](https://stackoverflow.com/questions/60166217/moshi-json-annotation-does-not-work-with-proguard) ❌ 
  - [Switching to `@field:Json` works on the release build, but does not work with debug build](https://github.com/square/moshi/issues/446#issuecomment-367732968) ❌ 
  - [This might be a bug in Android Gradle Plugin](https://github.com/square/moshi/issues/1458#issuecomment-1001718894), [upgrading to AGP 7.4.2 solves the issue](https://stackoverflow.com/questions/70315232/data-class-metadata-is-removed-with-proguard-r8-for-kotlin-1-6-0/70331139#70331139) ✅ 

- Upgrading to AGP 7.4.2 will lead to this Lint issue https://github.com/bumptech/glide/issues/4940 (https://github.com/nimblehq/android-templates/actions/runs/5396819037/jobs/9800872801)
	```
	/home/vsts/work/1/s/app/src/main/AndroidManifest.xml: Error: When targeting Android 13 or higher, posting a permission requires holding the POST_NOTIFICATIONS permission (usage from com.chuckerteam.chucker.internal.support.NotificationHelper) [NotificationPermission]
	  
	     Explanation for issues of type "NotificationPermission":
	     When targeting Android 13 and higher, posting permissions requires holding
	     the runtime permission android.permission.POST_NOTIFICATIONS.
	```
	
	👉 To fix this, add a new debug `AndroidManifest.xml` with a additional `<uses-permission android:name="android.permission.POST_NOTIFICATIONS" />` ✅ 

- Another issue: We must keep them from code obfuscation by proguard rules.

## Proof Of Work 📹

| None of fixes | Add proguard rules for models | Upgrade AGP to 7.4.2 |
|--------|--------|--------|
| Screenshot_20230628-090526 | Screenshot_20230628-090814 | Screenshot_20230628-090031 |

### DIFF
--- a/sample-compose/app/src/debug/AndroidManifest.xml
+++ b/sample-compose/app/src/debug/AndroidManifest.xml
@@ -1,14 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="co.nimblehq.sample.xml">
+    package="co.nimblehq.sample.compose">
 
     <!--For Chucker in debug build only-->
     <uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
-
-    <application>
-        <activity
-            android:name=".EmptyHiltActivity"
-            android:exported="false" />
-    </application>
 
 </manifest>
--- a/sample-compose/app/src/main/java/co/nimblehq/sample/compose/model/UiModel.kt
+++ b/sample-compose/app/src/main/java/co/nimblehq/sample/compose/model/UiModel.kt
@@ -6,7 +6,11 @@ import kotlinx.parcelize.Parcelize
 
 @Parcelize
 data class UiModel(
-    val id: String
+    val id: String,
+    val userName: String,
 ) : Parcelable
 
-fun Model.toUiModel() = UiModel(id.toString())
+fun Model.toUiModel() = UiModel(
+    id.toString(),
+    userName.orEmpty(),
+)
--- a/sample-compose/app/src/main/java/co/nimblehq/sample/compose/model/UiModel.kt
+++ b/sample-compose/app/src/main/java/co/nimblehq/sample/compose/model/UiModel.kt
@@ -7,10 +7,10 @@ import kotlinx.parcelize.Parcelize
 @Parcelize
 data class UiModel(
     val id: String,
-    val userName: String,
+    val username: String,
 ) : Parcelable
 
 fun Model.toUiModel() = UiModel(
     id.toString(),
-    userName.orEmpty(),
+    username.orEmpty(),
 )
--- a/sample-compose/data/src/main/java/co/nimblehq/sample/compose/data/response/Response.kt
+++ b/sample-compose/data/src/main/java/co/nimblehq/sample/compose/data/response/Response.kt
@@ -4,9 +4,15 @@ import co.nimblehq.sample.compose.domain.model.Model
 import com.squareup.moshi.Json
 
 data class Response(
-    @Json(name = "id") val id: Int?
+    @Json(name = "id")
+    val id: Int?,
+    @Json(name = "username")
+    val userName: String?,
 )
 
-private fun Response.toModel() = Model(id = this.id)
+private fun Response.toModel() = Model(
+    id = this.id,
+    userName = this.userName,
+)
 
 fun List<Response>.toModels() = this.map { it.toModel() }
--- a/sample-compose/data/src/main/java/co/nimblehq/sample/compose/data/response/Response.kt
+++ b/sample-compose/data/src/main/java/co/nimblehq/sample/compose/data/response/Response.kt
@@ -6,13 +6,14 @@ import com.squareup.moshi.Json
 data class Response(
     @Json(name = "id")
     val id: Int?,
+    // A sample to ensure @Json annotation work properly with a differ between field name and JSON key name
     @Json(name = "username")
     val userName: String?,
 )
 
 private fun Response.toModel() = Model(
     id = this.id,
-    userName = this.userName,
+    username = this.userName,
 )
 
 fun List<Response>.toModels() = this.map { it.toModel() }
--- a/sample-compose/domain/src/main/java/co/nimblehq/sample/compose/domain/model/Model.kt
+++ b/sample-compose/domain/src/main/java/co/nimblehq/sample/compose/domain/model/Model.kt
@@ -2,5 +2,5 @@ package co.nimblehq.sample.compose.domain.model
 
 data class Model(
     val id: Int?,
-    val userName: String?,
+    val username: String?,
 )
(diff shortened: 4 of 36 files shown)


template-compose/data/src/test/java/co/nimblehq/template/compose/data/test/MockUtil.kt
```diff
@@ -29,4 +29,8 @@ object MockUtil {
     val errorResponse = ErrorResponse(
         message = "message"
     )
+
+    val responses = listOf(
+        co.nimblehq.template.compose.data.response.Response(id = 1)
```
kaungkhantsoe, line 34: Optional: We can import like `import co.nimblehq.sample.compose.data.response.*` or using `as`

kaungkhantsoe (approved):
Just an optional comment. LGTM

sample-compose/app/src/debug/AndroidManifest.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    package="co.nimblehq.sample.compose">
+
+    <!--For Chucker in debug build only-->
+    <uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
```
doannimble, line 6: @luongvo Just curious, why don't we put this inside the main manifest? It looks like the generated file 🤔

luongvo: @doannimble We only need this permission for Chucker to work on the debug build. If the app needs to post notifications in the system tray, we will move this to the main Manifest.

doannimble: @luongvo Ah I see, thanks for your clarification 🙏

luongvo: If the main Manifest has 2 permission definitions, the debug Manifest has 1, then the final debug build has 3 while the release build has only 2.

doannimble (approved):

sample-compose/data/src/main/java/co/nimblehq/sample/compose/data/response/Response.kt
```diff
@@ -4,9 +4,15 @@ import co.nimblehq.sample.compose.domain.model.Model
 import com.squareup.moshi.Json
 
 data class Response(
-    @Json(name = "id") val id: Int?
+    @Json(name = "id")
+    val id: Int?,
+    @Json(name = "username")
+    val userName: String?,
```
AVI5HEK, line 10: `username` is one word.
```suggestion
    val username: String?,
```

sample-compose/domain/src/main/java/co/nimblehq/sample/compose/domain/model/Model.kt
```diff
@@ -1,5 +1,6 @@
 package co.nimblehq.sample.compose.domain.model
 
 data class Model(
-    val id: Int?
+    val id: Int?,
+    val userName: String?,
```
AVI5HEK, line 5: Same here.
```suggestion
    val username: String?,
```

sample-compose/app/src/main/java/co/nimblehq/sample/compose/model/UiModel.kt
```diff
@@ -6,7 +6,11 @@ import kotlinx.parcelize.Parcelize
 
 @Parcelize
 data class UiModel(
-    val id: String
+    val id: String,
+    val userName: String,
```
AVI5HEK, line 10: And here.
```suggestion
    val username: String,
```

sample-compose/data/src/main/java/co/nimblehq/sample/compose/data/response/Response.kt
luongvo: @AVI5HEK Maybe it's an exception here to produce the sample test with a differ between field name and JSON key name 😛

lydiasama: @luongvo Can we have a comment to explain the sample here too? I'm concerned that after this PR, someone will update it to `username` as we don't know the purpose of a differ name example.

luongvo: @lydiasama Yes, we should indeed.

luongvo: @AVI5HEK @lydiasama Added in https://github.com/nimblehq/android-templates/pull/480/commits/991c23356f8d3add114f86ff6832d905e51efa07

sample-compose/domain/src/main/java/co/nimblehq/sample/compose/domain/model/Model.kt
luongvo: Fixed in https://github.com/nimblehq/android-templates/pull/480/commits/991c23356f8d3add114f86ff6832d905e51efa07

lydiasama (approved):

sample-compose/data/src/main/java/co/nimblehq/sample/compose/data/response/Response.kt
```diff
@@ -4,9 +4,16 @@ import co.nimblehq.sample.compose.domain.model.Model
 import com.squareup.moshi.Json
 
 data class Response(
-    @Json(name = "id") val id: Int?
+    @Json(name = "id")
+    val id: Int?,
+    // A sample to ensure @Json annotation work properly with a differ between field name and JSON key name
```
AVI5HEK, line 9: How about this as the comment?
```suggestion
    // This field name is used to ensure @Json annotations work properly even if the field name and JSON key name are different.
```

AVI5HEK (approved):
Just one optional suggestion for grammatical accuracy. The rest LGTM.

sample-compose/data/src/main/java/co/nimblehq/sample/compose/data/response/Response.kt
luongvo: IMO, it's good enough. Thanks for your suggestion 🙏

Issue #454, opened 2023-04-04
Update generating project example script to be in code block


lydiasama:
## Why

Regarding [this raised issue from slack](https://nimble-co.slack.com/archives/CSQG2QXFX/p1680590062084329), the copied text gets the wrong double quote `"` symbol when copying the example from the current documentation.

## Solution

Update the double backticks code (``) to the [Fenced Code Block](https://www.markdownguide.org/extended-syntax/#fenced-code-blocks) to prevent wrong double quote symbols.

## Who Benefits?

Developers

PR #483, opened 2023-07-06
[#454] Update generating project example script to be in Fenced Code Block

lydiasama:
- Closes https://github.com/nimblehq/android-templates/issues/454

## What happened 👀

Update code block script from double backticks code (``) to the [Fenced Code Block](https://www.markdownguide.org/extended-syntax/#fenced-code-blocks)

## Insight 📝

N/A

## Proof Of Work 📹

Screenshot 2566-07-06 at 11 10 22 AM

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,7 @@
 ---
 
 A collection of our Android templates:
+
 - XML template: **[template-xml](https://github.com/nimblehq/android-templates/tree/develop/template-xml)**
 - Compose template: **[template-compose](https://github.com/nimblehq/android-templates/tree/develop/template-compose)**
 
@@ -23,6 +24,7 @@ A collection of our Android templates:
 1. Clone or download this repository to your local machine, then extract and open the folder
 2. Run `cd scripts` to get into the scripts directory
 3. Run `kscript new_project.kts` to create a new project with the following arguments:
+
   ```   
     package-name=                      New package name (i.e., com.example.package)
     app-name=                          New app name (i.e., MyApp, "My App", "my-app")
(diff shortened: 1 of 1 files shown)


sleepylee (approved):

doannimble (approved):

luongvo (approved):

AVI5HEK (approved):

lydiasama:
Screenshot 2566-07-06 at 11 29 44 AM

Can we ignore this error? 😢

luongvo:
> Screenshot 2566-07-06 at 11 29 44 AM
> 
> Can we ignore this error? 😢

@lydiasama @ryan-conway I think we should create a task to switch to [using Chucker from Maven Central](https://github.com/ChuckerTeam/chucker#getting-started-) instead of Jitpack. Jitpack server is not stable sometimes, so use it only if there is no other way 🤓 

FYI: I faced the same issue in the Gulf project but forget to raise a ticket.

lydiasama:
@luongvo Haven't we set it from the maven? 🤔 

Screenshot 2566-07-06 at 11 39 28 AM

luongvo:
> @luongvo Haven't we set it from the maven? 🤔
> 
> Screenshot 2566-07-06 at 11 39 28 AM

@lydiasama maybe, we MUST move up the mavenCentral config here. That needs time for investigation 🤷‍♂️ 

image

chornerman (approved):

lydiasama:
@ryan-conway Since the CI error is related to Chunker dependency, shall we merge this PR first and then create another issue to resolve it? 🙏🏻

kaungkhantsoe (approved):

lydiasama:
> > @luongvo Haven't we set it from the maven? 🤔
> > Screenshot 2566-07-06 at 11 39 28 AM
> 
> @lydiasama maybe, we MUST move up the mavenCentral config here as both use the same dependency id. That needs time for investigation 🤷‍♂️
> 
> image

@luongvo I found the root cause. It's not related to Jitpack connection, instead, it's about the syntax we implemented. With [this solution](https://stackoverflow.com/a/66926254/11393626), It's resolved it https://github.com/nimblehq/android-templates/pull/484 🤓

Issue #485, opened 2023-07-12
Fix: Chucker dependency cannot be resolved when running verify newproject script


ryan-conway:
## Issue

[CI fails](https://github.com/nimblehq/android-templates/actions/runs/5471468933/jobs/9962858668?pr=483) when running the verify newproject script. 

```
* What went wrong:
Execution failed for task ':app:dataBindingMergeDependencyArtifactsProductionDebug'.
> Could not resolve all files for configuration ':app:productionDebugCompileClasspath'.
   > Failed to transform library-3.5.2.aar (com.github.chuckerteam.chucker:library:3.5.2) to match attributes {artifactType=android-databinding, org.gradle.category=library, org.gradle.libraryelements=jar, org.gradle.status=release, org.gradle.usage=java-api}.
      > Could not find library-3.5.2.jar (com.github.chuckerteam.chucker:library:3.5.2).
        Searched in the following locations:
            https://jitpack.io/com/github/chuckerteam/chucker/library/3.5.2/library-3.5.2.aar
            https://jitpack.io/com/github/chuckerteam/chucker/library/3.5.2/library-3.5.2.jar
```

~We should fetch Chucker from Maven instead of Jitpack to prevent this from occuring~

The syntax we implemented to get maven (Jitpack) should be in Kotlin format:

```kotlin
repositories {
  maven { url = uri("https://www.jitpack.io" ) }
}
```

Ref: https://stackoverflow.com/a/66926254/11393626

## Expected

CI should be able to execute the verify newproject script successfully

## Steps to reproduce

1. Open a PR
2. Let CI run

PR #486, opened 2023-07-12
[#485] Fix repositories import syntax to be in Kotlin format for maven Jitpack

lydiasama:
- Close https://github.com/nimblehq/android-templates/issues/485

## What happened 👀

- Update the maven import to be in Kotlin format

## Insight 📝

Ref: https://stackoverflow.com/a/66926254/11393626

## Proof Of Work 📹

The CI should be successful.

### DIFF
--- a/RxJavaTemplate[DEPRECATED]/build.gradle.kts
+++ b/RxJavaTemplate[DEPRECATED]/build.gradle.kts
@@ -21,7 +21,7 @@ plugins {
 allprojects {
     repositories {
         google()
-        maven(url = "https://jitpack.io")
+        maven { url = uri("https://www.jitpack.io") }
         mavenCentral()
     }
 }
--- a/sample-compose/build.gradle.kts
+++ b/sample-compose/build.gradle.kts
@@ -20,7 +20,7 @@ plugins {
 allprojects {
     repositories {
         google()
-        maven(url = "https://jitpack.io")
+        maven { url = uri("https://www.jitpack.io") }
         mavenCentral()
     }
 }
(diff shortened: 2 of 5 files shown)


ryan-conway (approved):

chornerman (approved):

RxJavaTemplate[DEPRECATED]/build.gradle.kts
```diff
@@ -21,7 +21,7 @@ plugins {
 allprojects {
     repositories {
         google()
-        maven(url = "https://jitpack.io")
+        maven { url = uri("https://www.jitpack.io") }
```
Comment (doannimble) [inline RxJavaTemplate[DEPRECATED]/build.gradle.kts:24]: @lydiasama Since this package is deprecated, I'm not sure we should do the update or not 😄

doannimble (approved):

RxJavaTemplate[DEPRECATED]/build.gradle.kts
Comment (lydiasama) [reply RxJavaTemplate[DEPRECATED]/build.gradle.kts:24]: @doannimble I'll keep updating until @ryan-conway said Stop it! 😂

Comment (lydiasama) [reply RxJavaTemplate[DEPRECATED]/build.gradle.kts:24]: @luongvo I'm not sure if I understand your expectation correctly, but do you have another intention outside of fixing this error by fetching Chucker from MavenCentral?

Comment (luongvo) [reply RxJavaTemplate[DEPRECATED]/build.gradle.kts:24]: @lydiasama No, just my concern about fetching the dependency from the recommended source.

Comment (lydiasama) [reply RxJavaTemplate[DEPRECATED]/build.gradle.kts:24]: @luongvo My concern is if we get rid of maven Jitpack, will it affect to another lib we use?

Comment (lydiasama) [reply RxJavaTemplate[DEPRECATED]/build.gradle.kts:24]: @luongvo @ryan-conway The PR is here. https://github.com/nimblehq/android-templates/pull/487

Issue #496, opened 2023-08-02
Fix: Remove Firebase integration in order to generate new project from the script


lydiasama:
## Issue

When generateing the project from the script, it requires the `google-services.json` file which we saved in the Github Secrets. To support general users outside Nimble, we need to remove it to make the script can be built successfully without the Firebase integration. Besides, Codemagic supports distributing the app to Firebase App Distribution without integrating the Firebase to the project.

## Expected

The generate new project script can run successfully.

PR #499, opened 2023-08-03
[#496] Distribute app without Firebase integration

lydiasama:
- Close https://github.com/nimblehq/android-templates/issues/496

## What happened 👀

Remove Firebase integration from project

## Insight 📝

- Codemagic can distribute the app to Firebase App Distribution without integrating the Firebase to the project.
- After merging this PR, the `google-services.json` in Github Secrets will be deleted.

Big thanks to @luongvo for noticing this and the POC branch. 🙇🏻‍♀️ 

## Proof Of Work 📹

Github Action to deploy on Firebase result:

https://github.com/nimblehq/android-templates/actions/runs/5712267786/job/15475425029

Codemagic to deploy on Firebase result:

Screenshot 2566-08-02 at 4 54 16 PM

### DIFF
--- a/codemagic.yaml
+++ b/codemagic.yaml
@@ -10,13 +10,6 @@ definitions:
       cache_paths:
         - $HOME/.gradle/caches
   scripts:
-    - &set_up_google_services_files_from_environment_variables
-      name: Set up google-services.json on template-compose
-      script: |
-        mkdir -p template-compose/app/src/production
-        echo $GOOGLE_SERVICES_JSON > template-compose/app/src/production/google-services.json
-        mkdir -p template-compose/app/src/staging
-        echo $GOOGLE_SERVICES_JSON_STAGING > template-compose/app/src/staging/google-services.json
     - &detekt_on_template_compose
       name: Run detekt on template-compose
       working_directory: ./template-compose
@@ -58,7 +51,6 @@ workflows:
           include: false
       cancel_previous_builds: true
     scripts:
-      - *set_up_google_services_files_from_environment_variables
       - *detekt_on_template_compose
       - *unit_test_on_template_compose
     artifacts:
@@ -95,9 +87,8 @@ workflows:
       events:
         - push
       branch_patterns:
-        - pattern: develop
+        - pattern: chore/poc-distribute-app-without-firebase-integration
     scripts:
-      - *set_up_google_services_files_from_environment_variables
       - *detekt_on_template_compose
       - *unit_test_on_template_compose
       - name: Build APK for staging
@@ -125,7 +116,6 @@ workflows:
       branch_patterns:
         - pattern: main
     scripts:
-      - *set_up_google_services_files_from_environment_variables
       - *detekt_on_template_compose
       - *unit_test_on_template_compose
       - name: Build APK for production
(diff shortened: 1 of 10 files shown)


codemagic.yaml
```diff
@@ -10,13 +10,6 @@ definitions:
       cache_paths:
         - $HOME/.gradle/caches
   scripts:
-    - &set_up_google_services_files_from_environment_variables
-      name: Set up google-services.json on template-compose
-      script: |
-        mkdir -p template-compose/app/src/production
-        echo $GOOGLE_SERVICES_JSON > template-compose/app/src/production/google-services.json
-        mkdir -p template-compose/app/src/staging
-        echo $GOOGLE_SERVICES_JSON_STAGING > template-compose/app/src/staging/google-services.json
```
luongvo, line 19: Should we also clean up all unnecessary steps in https://github.com/search?q=repo%3Animblehq%2Fandroid-templates+GOOGLE_SERVICES_JSON&type=code

luongvo, line 19: Should we also clean up all unnecessary steps in https://github.com/search?q=repo%3Animblehq%2Fandroid-templates+GOOGLE_SERVICES_JSON&type=code

lydiasama:
All updated. @luongvo  🙏🏻

luongvo (commented):
There are still 1 more place to temporary remove:
- https://github.com/nimblehq/android-templates/blob/3f234ab29313997961b4526956b877b3ee0e2a12/.github/workflows/review_pull_request.yml#L50-L58

lydiasama:
@luongvo I just recalled why I didn't remove it. If we use this `github action workflow` from `.cicdtemplate` in any project to distribute the app to the Firebase, that means we need to set up the Firebase to the project. So, we recommended saving the `google-service.json` file in the Github Secrets. 🤔  So IMHO, we don't need to remove that step to download the file in the `.cicdtempleate` folder. What do you think?

luongvo:
> If we use this github action workflow from .cicdtemplate in any project to distribute the app to the Firebase, that means we need to set up the Firebase to the project. 

Yes, we need to set up a new Firebase app on the Firebase console for distribution, but we **do not need to add Firebase to the app**. Hence, no google-services is required.

> And we recommended saving the google-service.json file in the Github Secrets. 🤔 So IMHO, we don't need to remove that step to download the file in the .cicdtempleate folder.

We can keep that variable on GitHub Secrets for later usage. For now, we do not use it to execute any commands or to build the app, so I recommend removing the step to download it to avoid confusion. 

@lydiasama Since you removed from both the `root workflows` and the `.cicdtemplate`, we should also remove it in https://github.com/nimblehq/android-templates/blob/3f234ab29313997961b4526956b877b3ee0e2a12/.github/workflows/review_pull_request.yml#L50-L58 🙏

luongvo:
@lydiasama @ryan-conway can we resolve and merge this PR asap? I'm working on KMM Templates and to reuse the Android app module, I need to have a successful build without this config 🙏

ryan-conway (approved):

luongvo (approved):

AVI5HEK (approved):

PR #505, opened 2023-08-18
[#512] [KMM Support] Update Kotlin DSL build dependencies & plugins configuration

### LINKED ISSUE #512: Update Kotlin DSL build dependencies & plugins configuration
## Why

As presented via [this presentation](https://docs.google.com/presentation/d/1faEx-hxFHUZJRQuZgy0zIDSzUBd5x1mxqtC1sGSfMeo/edit?usp=sharing), we need to **update the Kotlin DSL build dependencies & plugins configuration** to support the KMM Templates in a consistent way across templates.

- Define all dependencies in a `buildSrc > Dependencies` class.
- Define all plugins in a `buildSrc > Plugins` class.

## Who Benefits?

Developers

luongvo:
Closes #512

## What happened 👀

- Define all dependencies in a `buildSrc > Dependencies` class.
- Define all plugins in a `buildSrc > Plugins` class.
- Some small .gitignore refactoring.

## Insight 📝

- I used to not agree to define all dependencies in the `buildSrc` module in the early stage because I was concerned that it would break the new version suggestion flow in AS IDE. However, some recent updates make it more feasible to apply:
 - This practice has now become more popular to apply. Especially from Gradle 8 🌟 
   - https://medium.com/@sarguru1981/dependency-management-using-kotlin-dsl-28d06d678dfc
   - https://proandroiddev.com/better-dependencies-management-using-buildsrc-kotlin-dsl-eda31cdb81bf
   - https://betterprogramming.pub/gradle-dependency-management-with-buildsrc-and-kotlin-dsl-1de958eab166
- Defining a `DependencyHandler` looks too cumbersome. We can keep listing necessary dependencies in each module's build.gradle file to keep the flexibility of customization.
- All version variables in the `Versions` class do not need to follow by a `_VERSION` postfix as they're in the Versions class already ✅ 
- A dependency version definition could be reused between production and test code, so we don't need to group them in each dependency and test section ✅ 

> [!NOTE]  
> No version upgrading in this PR 🙏 

## Proof Of Work 📹

The app works properly after migrating.


template-compose/build.gradle.kts
```diff
@@ -6,15 +6,15 @@ buildscript {
     }
 
     dependencies {
-        classpath("com.android.tools.build:gradle:${Versions.BUILD_GRADLE_VERSION}")
-        classpath("com.google.dagger:hilt-android-gradle-plugin:${Versions.HILT_VERSION}")
-        classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:${Versions.KOTLIN_VERSION}")
+        classpath("com.android.tools.build:gradle:${Versions.GRADLE}")
+        classpath("com.google.dagger:hilt-android-gradle-plugin:${Versions.HILT}")
+        classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:${Versions.KOTLIN}")
```
kaungkhantsoe, line 11: Can we also configure for these dependencies as well?

kaungkhantsoe (approved):
Just one question. The rest LGTM.

template-compose/buildSrc/src/main/java/Configurations.kt
lydiasama, line 1: I prefer to combine them all in this `Configurations` file instead of separating them. 🤔  Could you please give me more insight on why we refactor this or just wanna change the implementation?

template-compose/buildSrc/src/main/java/Dependencies.kt
```diff
@@ -0,0 +1,71 @@
...
+
+        const val JUNIT = "junit:junit:${Versions.JUNIT}"
+
+        const val KOTEST = "io.kotest:kotest-assertions-core:${Versions.KOTEST}"
+        const val MOCKK = "io.mockk:mockk:${Versions.MOCKK}"
+
+        const val ROBOLECTRIC = "org.robolectric:robolectric:${Versions.ROBOLECTRIC}"
+
+        const val TEST_CORE = "androidx.test:core:${Versions.CORE}"
+        const val TURBINE = "app.cash.turbine:turbine:${Versions.TURBINE}"
```
lydiasama, line 69: The constants inside this `Test` have been ordered alphabetically. Let's order it alphabetically for another object too. 🙏🏻

chornerman (approved):
Beside others' comments, LGTM

template-compose/buildSrc/src/main/java/BuildTypes.kt
```diff
@@ -0,0 +1,4 @@
+object BuildTypes {
```
AVI5HEK, line 1: I personally prefer the singular form for singletons, which is consistent with the naming convention of any regular class, sealed class, or enums. From the consumer's POV, the API would seem consistent too. My motivation is from [this](https://android.googlesource.com/platform/frameworks/support/+/androidx-main/compose/docs/compose-api-guidelines.md#kotlin-style-singletons_constants_sealed-class-and-enum-class-values-why) doc where they recommend using PascalCase for the constants as well, but we can keep using uppercase in this case. 💡

AVI5HEK (commented):
Just a small concern.

template-compose/build.gradle.kts
luongvo: @kaungkhantsoe Nope, this deprecated config [will be removed when migrating to Gradle 8](https://github.com/nimblehq/android-templates/pull/508/files#diff-a86af81bc38d910132ed4be8d8a2ece3b82bb775bdae609344303399f3726afc).

template-compose/buildSrc/src/main/java/Configurations.kt
luongvo: @lydiasama `Versions` and `Dependencies` are also configurations, so keeping only `Flavors`, `BuildTypes`, and `Modules` here does not make sense. As we separate `Dependencies`, `Plugins`, and `Versions` into separate classes, doing the same for `Flavors`, `BuildTypes`, and `Modules` makes sense IMO.

template-compose/buildSrc/src/main/java/Dependencies.kt
luongvo: @lydiasama This is a good point. Actually, my intention is to group some dependencies in the same lib group together, so it's hard to adapt this request to order them alphabetically.

Current
image

Alphabetically ordered, the BOM - which applies for Compose dependencies only - is now far away from its dependencies 🤔 
image

luongvo: @lydiasama I would suggest to keep the current implementation 🙏

template-compose/buildSrc/src/main/java/BuildTypes.kt
luongvo: @AVI5HEK I agree to use singular for sealed classes, or enums. However, this is an object class, and it's closer to what we defined with AppColors or AppDimensions in Compose template https://github.com/nimblehq/android-templates/tree/develop/template-compose/app/src/main/java/co/nimblehq/template/compose/ui/theme.

ryan-conway (approved):

luongvo:
Hi @nimblehq/chapter-android @ryan-conway Can we have more approvals for the first PR merge?

doannimble (approved):
lgtm 💯

PR #529, opened 2023-10-06
[#552] Improve the base navigation with a "MainNavGraph"

luongvo:
Closes https://github.com/nimblehq/android-templates/issues/552

## What happened 👀

- Define a "MainNavGraph" (and also "OnboardingNavGraph" later) nested graph between `AppNavGraph` (AppNavigation) and the app screens as default to improve the app navigation logic.

## Insight 📝

- The final app navigation structure will look like this:

	```
	AppNavGraph
	├── OnboardingNavGraph
	│   └── LoginScreen
	└── MainNavGraph
	    └── HomeScreen
	```

- The `AppDestination` also has been refactored into 3 components:
  - `BaseDestination`: base destination logic, such as route, arguments, destination, or Up action.
  - `AppDestination`: all nav graphs, such as RootNavGraph, RootNavGraph
  - `MainDestination`: each nav graph's destinations, such as Main and Home.

## Proof Of Work 📹

The app works properly ✅


sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppDestination.kt
```diff
@@ -16,6 +18,10 @@ sealed class AppDestination(val route: String = "") {
 
     object Up : AppDestination()
 
+    object RootNavGraph : AppDestination("rootNavGraph")
```
kaungkhantsoe, line 21: @luongvo How about we create a new Destination Class for `Main`? I think we should have only one `NavGraph` per destination file at the top. It is clearer and can know where to put the destination. Or we won't have consistency or grouping.

**With separate Destination file**
```
sealed class ADestination {
 object ANavGraph : AppDestination()
 object ScreenA : AppDestination()
 object ScreenB : AppDestination()
}
```

**Without separate Destination file**
```
sealed class AppDestination {
 object ANavGraph : AppDestination()
 object BNavGraph : AppDestination()
 object ScreenA : AppDestination()
 object CNavGraph : AppDestination()
 object ScreenB : AppDestination()
 object ScreenD : AppDestination()
}
```
What do you think?

lydiasama (commented):
What if some screens can be accessed from any screen, such as `pin login`, or some screen is reused by another flow? Shall we have a sample for that case here too?

luongvo:
> What if some screens can be accessed from any screen, such as `pin login`, or some screen is reused by another flow? Shall we have a sample for that case here too?

@lydiasama for that case, we can separate these screens into a new nav graph to be navigated from any previous nav graph easily. However, the sample is a bit out of this PR's scope. I recommend doing it in a separate task if we really need to have that sample.

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppDestination.kt
luongvo: @kaungkhantsoe good idea 👍 
I refactored the AppDestination into 3 components:
- BaseDestination: base destination logic, such as route, arguments, destination, or Up action.
- AppDestination: all nav graphs, such as RootNavGraph, RootNavGraph
- MainDestination: each nav graph's destinations, such as Main and Home.

Updated in https://github.com/nimblehq/android-templates/pull/529/commits/1fc5e8b41786dab053d08e57d522ed46cf26cfab

luongvo: This also reflects the structure from the GT project cc @hoangnguyen92dn 💪

kaungkhantsoe (approved):

ryan-conway (approved):

PR #530, opened 2023-10-06
[#547] Define "results" for the "Up" navigation to deliver result to previous screens

luongvo:
Closes https://github.com/nimblehq/android-templates/issues/547

## What happened 👀

- Define "results" for the "Up" navigation to deliver results to previous screens.
- Add a sample to show the `Updated!` message on Home screen after clicking on the `Update` button on the Second screen.

## Insight 📝

- To navigate back to the previous screen with a result, call this for example:

	```
	navigator(BaseDestination.Up().addResult(KeyResultOk, true))
	```

## Proof Of Work 📹

https://github.com/nimblehq/android-templates/assets/16315358/8953291a-88f3-4083-ba9b-d5659666ccc5


kaungkhantsoe (approved):
LGTM

sample-compose/app/src/main/java/co/nimblehq/sample/compose/extensions/SavedStateHandleExt.kt
```diff
@@ -0,0 +1,11 @@
+package co.nimblehq.sample.compose.extensions
+
+import androidx.lifecycle.SavedStateHandle
+
+fun <T> SavedStateHandle.getAndRemove(key: String): T? {
```
lydiasama, line 5:
```suggestion
fun <T> SavedStateHandle.getThenRemove(key: String): T? {
```
I think we better use `then` word.

sample-compose/app/src/main/java/co/nimblehq/sample/compose/ui/AppDestination.kt
```diff
@@ -16,7 +17,7 @@ sealed class AppDestination(val route: String = "") {
 
     open var parcelableArgument: Pair<String, Any?> = "" to null
 
-    object Up : AppDestination()
+    data class Up(val results: List<Result> = emptyList()) : AppDestination()
```
lydiasama, line 20:
```suggestion
    data class Up(val results: Map<String, Any> = emptyList()) : AppDestination()
```
Why don't we use `Map` instead? 🤔

luongvo: @lydiasama good suggestion, fixed in https://github.com/nimblehq/android-templates/pull/530/commits/54ef3e3b4700b5aaa04f10fb134876ce54c3ac48

sample-compose/app/src/main/java/co/nimblehq/sample/compose/extensions/SavedStateHandleExt.kt
luongvo: fixed in https://github.com/nimblehq/android-templates/pull/530/commits/4ed2969da6824ce98cefed38d01dedb7c7a5d788

ryan-conway (approved):

lydiasama (approved):

chornerman (approved):

PR #536, opened 2023-10-06
[#549] Refactor: data module packages

luongvo:
Closes https://github.com/nimblehq/android-templates/issues/549

## What happened 👀

- Separate between `local` vs `remote` data sources.
- Move all `requests` and `responses` models into `remote/models` package.

## Insight 📝

We forgot to apply "Use the plural for packages with homogeneous contents" for the sample-compose from #520.

## Proof Of Work 📹

The app works properly ✅


template-compose/data/src/main/java/co/nimblehq/template/compose/data/local/storages/BaseSharedPreferences.kt
```diff
@@ -1,4 +1,4 @@
-package co.nimblehq.template.compose.data.storages
+package co.nimblehq.template.compose.data.local.storages
```
ryan-conway, line 1: Shall we rename this package to `local.preferences`? For example, we may have a local database which may end up in a `local.database` package. "Storage" is typically an uncountable word, so "storages" sounds a bit strange to read 😅 What do you think?

The rest LGTM 🚀

luongvo: @ryan-conway makes sense. At least all components are `Preferences` 🤓 

image

Fixed in https://github.com/nimblehq/android-templates/pull/536/commits/bde5cf5ced84e6ca22bbf53dc88519fa2a71f7bc

ryan-conway (approved):

lydiasama (approved):

kaungkhantsoe (approved):

PR #577, opened 2024-05-28
[#228] Add TokenAuthenticator setup

### LINKED ISSUE #228: Add TokenAuthenticator setup

kaungkhantsoe:
#228 

## What happened 👀

Apply token authentication setup from GT to template

## Insight 📝

N/A

## Proof Of Work 📹

https://github.com/nimblehq/android-templates/assets/32578035/1a6925b5-1017-4a07-aae7-a9a0caf9948b

## Summary by CodeRabbit

- **New Features**
  - Added configurable API settings for the base URL, client ID, and client secret.
  - Added secure authentication state and token storage.
  - Added automatic authorization headers and token refresh handling, including retry and logout behavior.
  - Added support for authorized API requests alongside unauthenticated requests.

- **Documentation**
  - Added setup instructions and a sample configuration file.

- **Bug Fixes**
  - Improved recovery when encrypted authentication data becomes invalid.

- **Tests**
  - Added coverage for authentication status, token refresh, retries, and error handling.


template-compose/app/src/main/java/co/nimblehq/template/compose/di/modules/OkHttpClientModule.kt
```diff
@@ -31,6 +41,26 @@ class OkHttpClientModule {
         }
     }.build()
 
+    @Authorized
```
thawzintoe-ptut:
```suggestion
    @Authorized
    @Singleton
```

